PR #19, opened 2022-11-16
Add flag --grpc-meta for attaching metadata

feedmeapples:
## What was changed

Added `--grpc-meta` flag that allows to pass custom headers with Temporal requests

## Why?

allows to pass headers such as `authorization` header

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/client/factory.go
+++ b/client/factory.go
@@ -102,7 +102,7 @@ func configureSDK(ctx *cli.Context) error {
 		)
 	}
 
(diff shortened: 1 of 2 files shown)


feedmeapples:
> Please add an automated test for this. I'd use a mock gRPC server to verify the headers are properly sent.

added an item to track that

PR #49, opened 2022-12-06
Copy nvm install.sh script

feedmeapples:
## What was changed

Copies install script from nvm https://github.com/nvm-sh/nvm/blob/master/install.sh

The actual PR with changes is here https://github.com/temporalio/cli/pull/50

## Why?

Working on `temporal` install.sh script. Created a separate PR for ease of review

## Checklist

1. Closes 

2. How was this tested:

not tested. Just a copy

3. Any docs updates needed?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,502 @@
+#!/usr/bin/env bash
+
+# The MIT License (MIT)
+
+# Copyright (c) 2010 Tim Caswell
+
+# Copyright (c) 2014 Jordan Harband
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,502 @@
+#!/usr/bin/env bash
+
+# The MIT License (MIT)
+
+# Copyright (c) 2010 Tim Caswell
```
bergundy, line 5: Probably want to update this

bergundy (commented):
Do we want to put this script here?
It should be in a cdn.

bergundy (commented):
Not sure why we need the entire nvm script here tbh

install.sh
feedmeapples: updated the license here https://github.com/temporalio/cli/pull/50

feedmeapples:
> Not sure why we need the entire nvm script here tbh

i've created this PR first prior to https://github.com/temporalio/cli/pull/50 so it's easier (maybe) to review #50

PR #50, opened 2022-12-06
Adapt install.sh for temporal binary

feedmeapples:
## What was changed

Modified nvm's install script to work for `temporal` single binary

## Why?

new distribution model with .sh script 

## Checklist

1. Closes 

2. How was this tested:

execute the script:
```
./install.sh
```

verified that executing `temporalite` from bash works on linux (requires changing the URL to /cli repo when release manager works for Temporal CLI)   

3. Any docs updates needed?


install.sh
```diff
@@ -73,47 +71,88 @@ nvm_profile_is_bash_or_zsh() {
...
-    else
-      nvm_echo >&2 "Unexpected value \"$NVM_METHOD\" for \$NVM_METHOD"
+temporal_fetch_meta() {
+  local TEMPORAL_SOURCE_URL
+  TEMPORAL_SOURCE_URL="$TEMPORAL_SOURCE"
+  local TEMPORAL_ARCH
+  TEMPORAL_ARCH="$(temporal_arch)"
+  local TEMPORAL_PLATFORM
+  TEMPORAL_PLATFORM="$(temporal_os)"
+  TEMPORAL_SOURCE_META_URL="https://temporal.download/temporalite/latest?platform=${TEMPORAL_PLATFORM}&arch=${TEMPORAL_ARCH}"
```
feedmeapples, line 83:
```suggestion
  TEMPORAL_SOURCE_META_URL="https://temporal.download/cli/latest?platform=${TEMPORAL_PLATFORM}&arch=${TEMPORAL_ARCH}"
```

robholland:
I think this script is more complex than we need because it involves things like adjusting user's shell profile. I don't think we need that for Temporal, nvm needs it because it's a version manager, we are just a single binary.

install.sh
```diff
@@ -0,0 +1,537 @@
...
+    esac
+    local _exe_name="temporal$_bext"
+    local _exe="$_dir/$_exe_name"
+    ensure chmod u+x "$_exe"
+
+    local _retval=$?
+    return "$_retval"
+}
+
+check_proc() {
```
robholland, line 164: This isn't used.

install.sh
```diff
@@ -0,0 +1,537 @@
...
+    # Check for /proc by looking for the /proc/self/exe link
+    # This is only run on Linux
+    if ! test -L /proc/self/exe; then
+        err "fatal: Unable to find /proc/self/exe.  Is /proc mounted?  Installation cannot proceed without /proc."
+    fi
+}
+
+get_bitness() {
```
robholland, line 172: Unused.

install.sh
```diff
@@ -0,0 +1,537 @@
...
+    if [ "$_current_exe_head" = "$(printf '\177ELF\001')" ]; then
+        echo 32
+    elif [ "$_current_exe_head" = "$(printf '\177ELF\002')" ]; then
+        echo 64
+    else
+        err "unknown platform bitness"
+    fi
+}
+
+is_host_amd64_elf() {
```
robholland, line 191: Unused.

install.sh
```diff
@@ -0,0 +1,537 @@
...
+    need_cmd tail
+    # ELF e_machine detection without dependencies beyond coreutils.
+    # Two-byte field at offset 0x12 indicates the CPU,
+    # but we're interested in it being 0x3E to indicate amd64, or not that.
+    local _current_exe_machine
+    _current_exe_machine=$(head -c 19 /proc/self/exe | tail -c 1)
+    [ "$_current_exe_machine" = "$(printf '\076')" ]
+}
+
+get_endianness() {
```
robholland, line 202: Unused.

install.sh
```diff
@@ -0,0 +1,537 @@
...
+
+unzip() {
+    local _file="$1"
+    local _dir="$2"
+
+    tar -xzvf "$_file" -C "$_dir"
+}
+
+get_default_install_dir() {
+    [ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.temporalio" || printf %s "${XDG_CONFIG_HOME}/temporalio"
```
robholland, line 382: We're installing a binary rather than config, not sure XDG_CONFIG_HOME is really appropriate. I think just use `"${HOME}/.temporalio"`

robholland (approved):
Looks good once unused code is removed. Is this tested anywhere?

feedmeapples:
> 

added a simple test here https://github.com/temporalio/cli/pull/74

feedmeapples:
> I think this script is more complex than we need because it involves things like adjusting user's shell profile. I don't think we need that for Temporal, nvm needs it because it's a version manager, we are just a single binary.

added a simple test here https://github.com/temporalio/cli/pull/74

install.sh
feedmeapples: updated

PR #54, opened 2022-12-13
Allowing configuring UI codec URL via flag

feedmeapples:
## What was changed

Added flag `--ui-codec-endpoint` for configuring codec endpoint used by UI

copy of https://github.com/temporalio/temporalite/pull/174

resolve https://github.com/temporalio/cli/issues/52

## Why?

enables the use of codec via a single run comand

## Checklist

1. Closes 

2. How was this tested:

Manually

3. Any docs updates needed?

### DIFF
--- a/server/ui.go
+++ b/server/ui.go
@@ -37,11 +37,9 @@ import (
 	uiserveroptions "github.com/temporalio/ui-server/v2/server/server_options"
 )
 
-func newUIOption(frontendAddr string, uiIP string, uiPort int, configDir string) (ServerOption, error) {
-	cfg, err := NewUIConfig(
-		frontendAddr,
-		uiIP,
-		uiPort,
+func newUIOption(c *uiconfig.Config, configDir string) (ServerOption, error) {
+	cfg, err := MergeWithConfigFile(
+		c,
 		configDir,
 	)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


robholland (approved):

server/ui.go
```diff
@@ -50,19 +48,37 @@ func newUIOption(frontendAddr string, uiIP string, uiPort int, configDir string)
...
+func NewUIConfig(c *uiConfig, configDir string) (*uiconfig.Config, error) {
 	cfg := &uiconfig.Config{
-		Host: uiIP,
-		Port: uiPort,
+		Host:                c.Host,
+		Port:                c.Port,
+		TemporalGRPCAddress: c.TemporalGRPCAddress,
+		EnableUI:            c.EnableUI,
+		Codec: uiconfig.Codec{
+			Endpoint: c.CodecEndpoint,
```
bergundy, line 58: Is it okay if this is empty?

server/ui.go
```diff
@@ -50,19 +48,37 @@ func newUIOption(frontendAddr string, uiIP string, uiPort int, configDir string)
...
+
+	// See https://github.com/temporalio/temporalite/pull/174/files#r1035958308
+	// for context about those overrides.
+	cfg.TemporalGRPCAddress = c.TemporalGRPCAddress
+	cfg.EnableUI = c.EnableUI
+
 	return cfg, nil
 }
+
+type uiConfig struct {
```
bergundy, line 78: Why not use `uiconfig.Config` instead of creating this new struct and renaming `NewUIConfig` to `MergeWithConfigFile`?
This is a bit confusing to me.

server/ui.go
feedmeapples: yea the default value is empty. This means it's disabled

server/ui.go
feedmeapples: good point

bergundy (approved):

PR #65, opened 2022-12-31
Copyedit start-dev help text

lorensr:


lorensr:
Getting this CI failure on Windows. Rerunning

image

lorensr:
Failure is persistent and on `main` as well, recommend getting green before launch

server/commands.go
```diff
@@ -56,74 +56,75 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
...
 					Name:    common.FlagLogFormat,
-					Usage:   `customize the log formatting (allowed: ["json" "pretty"])`,
+					Usage:   `Set the log formatting. Options: ["json", "pretty"].`,
 					EnvVars: nil,
 					Value:   "json",
 				},
 				&cli.StringFlag{
 					Name:    common.FlagLogLevel,
-					Usage:   `customize the log level (allowed: ["debug" "info" "warn" "error" "fatal"])`,
+					Usage:   `Set the log level. Options: ["debug" "info" "warn" "error" "fatal"].`,
```
feedmeapples, line 108: no need for the dots at the end

server/commands.go
```diff
@@ -56,74 +56,75 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
...
 					Name:        common.FlagUIIP,
-					Usage:       `IPv4 address to bind the web UI to instead of localhost`,
-					DefaultText: "same as --ip (eg. 127.0.0.1)",
+					Usage:       `IPv4 address to bind the Web UI to`,
+					DefaultText: "same as --ip",
 				},
 				&cli.StringFlag{
 					Name:    common.FlagLogFormat,
-					Usage:   `customize the log formatting (allowed: ["json" "pretty"])`,
+					Usage:   `Set the log formatting. Options: ["json", "pretty"].`,
```
feedmeapples, line 102: as another option, no strong opinion. So this in another cli and liked
```suggestion
					Usage:   `Set the log formatting. Supports: 'json', 'pretty'`,
```

server/commands.go
feedmeapples, line 108:
```suggestion
					Usage:   `Set the log level. Supports 'debug', 'info', 'warn', 'error', fatal',
```

lorensr: periods are to remove double parens and give separation between [] and () and be consistent with the switch to "sentences". 

`customize the log level (allowed: ["debug" "info" "warn" "error" "fatal"]) (default: "info")`

vs

`Set the log level. Options: ["debug" "info" "warn" "error" "fatal"]. (default: "info")`

just my stylistic preference, could remove second period. maybe @djmagee has an opinion

from:

image


to:

image

server/commands.go
lorensr: I like this better. I didn't change the array format because I saw that elsewhere in the code, it's printed with a Go stringification command. Can those be changed to this format as well?

server/commands.go
```diff
@@ -56,74 +56,75 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
...
 					Name:    common.FlagPort,
 					Aliases: []string{"p"},
-					Usage:   "Port for the temporal-frontend GRPC service",
+					Usage:   "Port for the frontend gRPC service",
 					Value:   sconfig.DefaultFrontendPort,
 				},
 				&cli.IntFlag{
 					Name:  common.FlagMetricsPort,
-					Usage: "Port for the metrics listener",
+					Usage: "Port for /metrics",
```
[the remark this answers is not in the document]
lorensr: some won't, i was trying to go for brevity. perhaps: "Port for metrics. http://localhost:<port>/metrics (default: disabled)"

server/commands.go
feedmeapples: ok i'm good with the current approach. If we decide to change the format we can do this separately so it's consistent. Ps saw the `Supports 'xxx', 'yyy'` format in kubectl and liked it because it is short and clean

feedmeapples (approved):

server/commands.go
lorensr: 👍 to be clear, by "this" i meant your suggestion 😊

PR #68, opened 2023-01-05
Allow using custom assets for the UI.

robholland:
This makes integration testing of UI changes much easier as we avoid the need to re-compile the ui-server and the Temporal CLI.

## What was changed
Added support for using a custom UI asset path.

## Why?
Makes testing UI changes much easier, avoiding any need to recompile ui-server or the Temporal CLI.

## Checklist

1. Closes 

2. How was this tested:
`./temporal server start-dev --ui-asset-path $PWD/../ui/.vercel/output/static/` versus `./temporal server start-dev`. Confirmed the adjusted UI is served when the option is passed.

3. Any docs updates needed?
No.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -149,3 +149,5 @@ require (
 )
 
 replace github.com/grpc-ecosystem/grpc-gateway => github.com/temporalio/grpc-gateway v1.17.0
+
+replace github.com/temporalio/ui-server/v2 v2.9.1 => github.com/temporalio/ui-server/v2 v2.9.2-0.20230105175408-7b947c15ce48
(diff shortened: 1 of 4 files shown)


robholland:
Note: `go.mod` replace will be removed once https://github.com/temporalio/ui-server/pull/267 is merged and released, before this is merged.

feedmeapples (approved):

go.mod
```diff
@@ -149,3 +149,5 @@ require (
 )
 
 replace github.com/grpc-ecosystem/grpc-gateway => github.com/temporalio/grpc-gateway v1.17.0
+
+replace github.com/temporalio/ui-server/v2 v2.9.1 => github.com/temporalio/ui-server/v2 v2.9.2-0.20230105175408-7b947c15ce48
```
feedmeapples, line 153: just so we don't forget to remove

PR #71, opened 2023-01-11
Auto generate cli docs

DJSanti:
## What was changed
A program that extracts information about the commands and converts it to Markdown files. This would then be exported to the documentation team to use as source docs for their purposes.

## Why?
Saves time on figuring out the intricacies of each command.
Keeps information updated as the main branch changes.

## Checklist
- [x] Link or download generated documents to /documentation
- [x] Atomize options
- [x] Link to Temporal terms mentioned in the text
- [x] Include additional information and code samples
- [x] Create frontmatter
Note: already created program that breaks down the markdown-converted application into separate files and directories, with basic Markdown formatting.

## How was this tested:
I iterated each step of this program. I can confirm that it will work whether a /docs folder exists or not. It will continue to update what's present.

## Any docs updates needed?
YES.
Docs team is going to download these files and use them to generate pages for the website.

-----

Any feedback to the code, especially where additional information needs to be added for docs, would be greatly appreciated. I'm pretty new to Go and don't know how to advance from this point beyond creating more manual work in the docs repo.

UPDATE: program is nearly complete. Just needs additional text in UsageText and it will fully generate docs. And, also, linking up to docs repo.


common/defs-flags.go
```diff
@@ -0,0 +1,60 @@
+package common
+const (
+	// Shared flag definitions
+	FlagEnvDefinition = "Name of the environment to read environmental variables from."
```
feedmeapples, line 4: why separating the definitions? These don't seem to be re-used anywhere

cmd/doc/main.go
```diff
@@ -0,0 +1,216 @@
...
+title: temporal {{.Title}}{{if not .IsIndex}} {{.ID}}{{end}}
+sidebar_label:{{if .IsIndex}} {{.Title}}{{else}} {{.ID}}{{end}}
+description: {{.Description}}
+tags:
+	- cli
+---
+
+`
+
+type FMStruct struct {
```
feedmeapples, line 34: no need to say Struct, just FrontMatter should be fine

common/defs-flags.go
DJSanti: Some are repeated. This was mostly to make it easier to find what needs to be changed (and can be useful should the CLI ever support different written languages).

DJSanti: Last job I worked in had a heavy focus on optimizing for the translators, so force of habit.

feedmeapples: right makes sense

feedmeapples (approved):

flossypurse (approved):
OK, looking good to me.

As long as @feedmeapples is good with it!

DJSanti:
Changed links for linkmagic

PR #74, opened 2023-01-16
Add temporal in PATH

feedmeapples:
## What was changed

When installing Temporal CLI with shell script also add the CLI in PATH

## Why?

allows to execute temporal from any path in terminal

## Checklist

1. Closes 

2. How was this tested:

Installed Temporal CLI with `./install.sh` and
```
which temporal                                                                                                                                                                             ✔ 
/home/player0/.temporalio/temporal
```

3. Any docs updates needed?


bergundy:
We need to automated tests for features like this so they don't break and hurt the getting started experience.
We'd need coverage for this script with different shells and test different edge cases.
I'm not convinced we should touch the user's profile and append the path automatically, instead we can instruct the user to do this themselves if the choose to.

feedmeapples:
> We need to automated tests for features like this so they don't break and hurt the getting started experience

Added a simple test that verifies temporal can be executed after `install.sh` installation.

> I'm not convinced we should touch the user's profile and append the path automatically, instead we can instruct the user to do this themselves if the choose to.

because there is a chance to break the user profile or?

bergundy:
> because there is a chance to break the user profile?

It's more "polite", I've generally seen tools like this instruct how to add to the profile instead of editing on behalf of the user.

feedmeapples:
> > because there is a chance to break the user profile?
> 
> It's more "polite", I've generally seen tools like this instruct how to add to the profile instead of editing on behalf of the user.

changed it to only suggest adding the path string to the user profile

install.sh
```diff
@@ -382,6 +384,34 @@ get_default_install_dir() {
...
+    *:"$_dir":*)
+        ;;
+    *)
+        export PATH="$_dir:\$PATH"
+        ;;
+esac
+EOL
+
+    local _source=". $_dir/env"
+    say "To add temporal to PATH update your user profile with the following (ex. ~/.bash_profile):"
```
robholland, line 411: I wouldn't create an env file, just suggest they add the directory to their path, I think we can expect users to be able to do that.

feedmeapples: This still could be helpful especially for students / people who just learn to code?

robholland: Can do, the env file won't work with all shells though so I'd at least want to make that clear, "If you are using bash you can..."

install.sh
```diff
@@ -0,0 +1,567 @@
...
```
feedmeapples, line 411:
```suggestion
    say "If you are using bash you can add temporal to PATH by appending the following to your user profile (ex. ~/.bash_profile):"
```

install.sh
```diff
@@ -0,0 +1,567 @@
+#!/bin/sh
+# shellcheck shell=dash
```
bergundy, line 2: We should be running shellcheck in CI

.github/workflows/ci.yml
```diff
@@ -29,3 +29,19 @@ jobs:
         env:
           CGO_ENABLED: 0
         run: go test -v ./...
+
+  test-install:
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        shell: [sh, bash, zsh]
```
bergundy, line 38: Rustup supports more shells, can we test a few more of them please?

.github/workflows/ci.yml
```diff
@@ -29,3 +29,19 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Print build information
+        run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
+
+      - uses: actions/checkout@v3
+
+      - name: Test install.sh script
+        run: ${{ matrix.shell }} ./install.sh
```
bergundy, line 47: How do you know this works?

bergundy: Add some assertions like running the server post installation and possibly `tee` the output and grepping on expected strings.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+    echo 'temporal does not work with this ksh93 version; please try bash!' >&2
+    exit 1
+fi
+
+set -u
+
+#XXX: If you change anything here, please make the same changes in setup_mode.rs
+usage() {
+    cat 1>&2 <<EOF
+temporal.sh 0.1.0
```
bergundy, line 54: That's not the name of the script, `install.sh` or `temporal-install.sh` or `install-temporal.sh` are all better names IMHO.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+The installer for temporal
```
bergundy, line 55:
```suggestion
The installer for the Temporal CLI
```

install.sh
djmagee:
```suggestion
    say "If you are using bash, you can add Temporal to PATH by appending the following to your user profile (ex. ~/.bash_profile):"
```

install.sh
```diff
@@ -0,0 +1,567 @@
...
+                case "$sub_arg" in
+                h)
+                    usage
+                    exit 0
+                    ;;
+                y)
+                    # user wants to skip the prompt --
+                    # we don't need /dev/tty
+                    need_tty=no
+                    ;;
```
bergundy, line 116: We don't have any prompts, not need to require a TTY.

install.sh
lorensr:
```suggestion
    say "If you are using bash, you can add temporal to PATH by appending the following to your user profile (e.g. ~/.bash_profile):"
```

the temporal command is lowercase

install.sh
```diff
@@ -0,0 +1,567 @@
...
+
+    if $_ansi_escapes_are_valid; then
+        printf "\33[1minfo:\33[0m downloading Temporal CLI\n" 1>&2
+    else
+        printf '%s\n' 'info: downloading Temporal CLI' 1>&2
+    fi
+
+    # TODO replace with
+    # local _url="https://temporal.download/temporalite/archive/latest?platform=${_platform}&arch=${_arch}"
+    local _url="https://temporal.download/assets/temporalio/cli/releases/download/v0.2.0/cli_0.2.0_${_platform}_${_arch}.tar.gz"
```
bergundy, line 133: The release artifact should be called `temporal-cli_...` so when users manually download it they can identify the file.

bergundy: We need to fix this but it's not blocking this PR.

bergundy: OIC I already made an issue for this: https://github.com/temporalio/cli/issues/56

install.sh
```diff
@@ -0,0 +1,567 @@
...
+        _ext=".zip"
+        ;;
+    esac
+
+    local _dir="$(ensure get_install_dir)"
+    ensure mkdir -p "$_dir"
+
+    local _archive="${_dir}/temporal_cli_latest${_ext}"
+    ensure downloader "$_url" "$_archive" "$_arch"
+    ensure unzip "$_archive" "$_dir"
```
bergundy, line 147: Does this work on windows?

install.sh
```diff
@@ -0,0 +1,567 @@
...
+    if [ "$_current_exe_head" = "$(printf '\177ELF\001')" ]; then
+        echo 32
+    elif [ "$_current_exe_head" = "$(printf '\177ELF\002')" ]; then
+        echo 64
+    else
+        err "unknown platform bitness"
+    fi
+}
+
+is_host_amd64_elf() {
```
bergundy, line 193: Delete, it's unused

bergundy: I'd check rustup to understand how it's used.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+
+check_proc() {
+    # Check for /proc by looking for the /proc/self/exe link
+    # This is only run on Linux
+    if ! test -L /proc/self/exe; then
+        err "fatal: Unable to find /proc/self/exe.  Is /proc mounted?  Installation cannot proceed without /proc."
+    fi
+}
+
+get_bitness() {
```
bergundy, line 174: Also unused, not sure if we care or if we should just fail in the CDN script

install.sh
```diff
@@ -0,0 +1,567 @@
...
+    local _exe="$_dir/$_exe_name"
+    ensure chmod u+x "$_exe"
+
+    ensure install_env "$_dir"
+
+    local _retval=$?
+    return "$_retval"
+}
+
+check_proc() {
```
bergundy, line 166: Unused

install.sh
```diff
@@ -0,0 +1,567 @@
...
+    need_cmd tail
+    # ELF e_machine detection without dependencies beyond coreutils.
+    # Two-byte field at offset 0x12 indicates the CPU,
+    # but we're interested in it being 0x3E to indicate amd64, or not that.
+    local _current_exe_machine
+    _current_exe_machine=$(head -c 19 /proc/self/exe | tail -c 1)
+    [ "$_current_exe_machine" = "$(printf '\076')" ]
+}
+
+get_endianness() {
```
bergundy, line 204: Unused

install.sh
```diff
@@ -0,0 +1,567 @@
...
+# will immediately terminate with an error showing the failing
+# command.
+ensure() {
+    if ! "$@"; then err "command failed: $*"; fi
+}
+
+# This is just for indicating that commands' results are being
+# intentionally ignored. Usually, because it's being executed
+# as part of error handling.
+ignore() {
```
bergundy, line 293: Unused

install.sh
bergundy: You can have a shell specific message here.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+get_install_dir() {
+    local _dir
+    _dir="$(get_default_install_dir)"
+    if [ -n "${TEMPORAL_DIR-}" ]; then
+        _dir="$TEMPORAL_DIR"
+    fi
+    printf %s "$_dir"
+}
+
+install_env() {
```
bergundy, line 396: This would need to be renamed since essentially we don't install anything.
`prompt_for_path` would be a better name.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+
+main() {
+    downloader --check
+    need_cmd uname
+    need_cmd mktemp
+    need_cmd chmod
+    need_cmd mkdir
+    need_cmd rm
+    need_cmd rmdir
+    need_cmd tar
```
bergundy, line 70: Please check that you actually need all of these commands and see what's needed in Windows to unzip.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+    local _url="https://temporal.download/assets/temporalio/cli/releases/download/v0.2.0/cli_0.2.0_${_platform}_${_arch}.tar.gz"
+
+    local _ext="tar.gz"
+    case "$_arch" in
+    *windows*)
+        _ext=".zip"
+        ;;
+    esac
+
+    local _dir="$(ensure get_install_dir)"
```
bergundy, line 142: I would download to a temp dir that is automatically cleaned up by the OS and unzip to the install dir.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+
+unzip() {
+    local _file="$1"
+    local _dir="$2"
+
+    tar -xzvf "$_file" -C "$_dir"
+}
+
+get_default_install_dir() {
+    [ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.temporalio" || printf %s "${XDG_CONFIG_HOME}/temporalio"
```
bergundy, line 384: I would put the final binary in this directory at `/bin`.

install.sh
```diff
@@ -0,0 +1,567 @@
...
+}
+
+get_install_dir() {
+    local _dir
+    _dir="$(get_default_install_dir)"
+    if [ -n "${TEMPORAL_DIR-}" ]; then
```
bergundy, line 390: If you want to have users customize the install dir, accept CLI args instead please.

install.sh
feedmeapples: resolving, started this discussion

.github/workflows/ci.yml
feedmeapples: sure. I'm expecting that some of the tests will fail and we just proceed with merging until fixing the tests for other shells

.github/workflows/ci.yml
feedmeapples: it's going to throw 127 error code if `temporal` is not found. Otherwise if `temporal -h` succeeds it ends with 0

install.sh
feedmeapples: no i haven't tested on Windows and it's not going to currently work since unzip() doesn't explicitly call `zip`. I want to test Windows via the Github Actions test, likely in a new PR

install.sh
feedmeapples: updated the PR with the removed code (from another PR)

install.sh
feedmeapples: deleted in an already merged PR. Rebased the current one

install.sh
feedmeapples: deleted in an already merged PR. Rebased the current one

install.sh
feedmeapples: deleted in an already merged PR. Rebased the current one

install.sh
feedmeapples: sg

install.sh
feedmeapples: good point done

install.sh
feedmeapples: sg

install.sh
```diff
@@ -52,7 +52,7 @@ set -u
 usage() {
     cat 1>&2 <<EOF
 temporal.sh 0.1.0
```
feedmeapples, line 54:
```suggestion
install.sh 0.1.0
```

install.sh
```diff
@@ -319,7 +314,35 @@ unzip() {
...
+    *:"$_dir":*)
+        ;;
+    *)
+        export PATH="$_dir:\$PATH"
+        ;;
+esac
+EOL
+
+    local _source=". $_dir/env"
+    say "If you are using bash, you can add temporal to PATH by appending the following to your user profile (e.g. ~/.bash_profile):"
```
djmagee, line 344:
```suggestion
    say "If you are using bash, you can add the temporal command to PATH by appending the following to your user profile (e.g. ~/.bash_profile):"
```
To distinguish the command name from our company and product names.

install.sh
feedmeapples: removed

djmagee (approved):

install.sh
feedmeapples: great suggestion, added a job in CI

install.sh
feedmeapples: 👍 added a `--dir path` arg

install.sh
bergundy: Okay, let's make sure to track this.

install.sh
bergundy: Hmm.. why not download the latest?

.github/workflows/ci.yml
feedmeapples: slightly updated the action

install.sh
```diff
@@ -319,7 +295,45 @@ unzip() {
...
+            *)
+                ;;
+        esac
+        shift
+    done
+
+    printf %s "$_dir"
+}
+
+prompt_for_path() {
```
bergundy, line 320: I'd replace this entire thing with:

```
say "Temporal CLI installed at $_dir/temporal"
if echo ":$PATH:" | grep -q "$_dir"; then
    say "Start the server with: temporal server start-dev"
    say "Or start a workflow with: temporal workflow start"
    say "For usage, run: temporal --help"
else
    say "For convenience, we recommend adding it to your PATH"
    say "If using bash, run echo "export PATH=\"$PATH:$_dir\" >> ~/.bash_profile"
fi
```

And maybe add the "getting started" commands to the "else" statement too.

bergundy (commented):
Please remove `-v` from the tar flags.

feedmeapples:
Updated install tests

image

install.sh
feedmeapples: done

install.sh
```diff
@@ -137,23 +113,33 @@ main() {
...
+        exit 1
+    fi
 
-    local _archive="${_dir}/temporal_cli_latest${_ext}"
+    local _archive="${_temp}/temporal_cli_latest${_ext}"
     ensure downloader "$_url" "$_archive" "$_arch"
-    ensure unzip "$_archive" "$_dir"
-    ensure rm "$_archive"
+    ensure unzip "$_archive" "$_temp"
+    local _dir="$(ensure get_install_dir "$@")"
```
robholland, line 126: I think this would need manual exit propagation as well?

install.sh
```diff
@@ -315,13 +298,49 @@ unzip() {
...
+}
+
+prompt_for_path() {
+    local _dirbin="$1"
+
+    local _source="export PATH=\"\$PATH:$_dirbin\" >> ~/.bash_profile"
+
+    say "Temporal CLI installed at $_dirbin/temporal"
+
+    if echo ":$PATH:" | grep -q "$_dirbin"; then
```
robholland, line 334:
```suggestion
    if echo ":$PATH:" | grep -q ":$_dirbin:"; then
```

.github/workflows/ci.yml
```diff
@@ -29,3 +29,28 @@ jobs:
...
         run: go test -v ./...
+
+  test-shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Run ShellCheck
+        uses: ludeeus/action-shellcheck@master
```
robholland, line 40:
```suggestion
        uses: ludeeus/action-shellcheck@master
        with:
            ignore_paths: ./.devcontainer
```

README.md
```diff
@@ -11,6 +11,10 @@ The Temporal CLI is a command-line interface for running Temporal Server and int
 
 ### Installation
 
+#### curl
+
+`curl -sSf https://temporal.download/cli/ | sh`
```
cretz, line 16: Personally, I would prefer `https://temporal.download/cli/install.sh`

feedmeapples: same :)

install.sh
feedmeapples: updated to the latest. Initially had an issue with /archive links which are now fixed

install.sh
feedmeapples: addressed

README.md
feedmeapples: as discussed internally updated to https://temporal.download/cli.sh

.github/workflows/ci.yml
bergundy: Make sure this is tracked before merging please.

README.md
```diff
@@ -11,6 +11,10 @@ The Temporal CLI is a command-line interface for running Temporal Server and int
 
 ### Installation
 
+#### curl
+
+`curl -sSf https://temporal.download/cli.sh | sh`
```
bergundy, line 16: I'm having second thoughts here, maybe call this `install.sh`?
I understand if you don't want to change this yet again.

install.sh
```diff
@@ -315,13 +300,49 @@ unzip() {
     local _file="$1"
     local _dir="$2"
 
-    tar -xzvf "$_file" -C "$_dir"
+    tar -xzf "$_file" -C "$_dir"
```
bergundy, line 303: This still doesn't work on windows. it should be fixed or have an issue opened to track.

install.sh
```diff
@@ -315,13 +300,49 @@ unzip() {
...
+        shift
+    done
+
+    printf %s "$_dir"
+}
+
+prompt_for_path() {
+    local _dirbin="$1"
+
+    local _source="export PATH=\"\\\$PATH:$_dirbin\" >> ~/.bashrc"
```
bergundy, line 332: Should this be `.bash_profile`?

install.test.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
bergundy, line 1: I would really like to avoid writing more shell scripts than necessary.
We can merge this, but please open an issue to rewrite this in an actual language.

bergundy (approved):

install.sh
feedmeapples: there is an issue open

README.md
feedmeapples: let's discuss in a thread

install.sh
feedmeapples: with .bash_profile users will have to re-login for the command to take effect. With .bashrc they can just open a new terminal and have `temporal` working

bergundy: Yeah, I brushed up on this and bashrc is better here.

PR #80, opened 2023-01-22
Rename namespace register -> create

feedmeapples:
## What was changed

Renamed `temporal operator namespace register` ->  `temporal operator namespace create`

## Why?

Consistency with other commands and Temporal Cloud UI

## Checklist

1. Closes 

2. How was this tested:

updated tests

3. Any docs updates needed?

### DIFF
--- a/namespace/namespace.go
+++ b/namespace/namespace.go
@@ -36,10 +36,10 @@ func NewNamespaceCommands() []*cli.Command {
 			Name:      "create",
 			Aliases:   []string{"register"},
 			Usage:     "Register a new Namespace",
-			Flags:     registerNamespaceFlags,
+			Flags:     createNamespaceFlags,
 			ArgsUsage: "namespace_name [cluster_name...]",
 			Action: func(c *cli.Context) error {
-				return RegisterNamespace(c)
+				return createNamespace(c)
 			},
 		},
 		{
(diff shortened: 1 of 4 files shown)


namespace/namespace.go
```diff
@@ -33,7 +33,7 @@ func NewNamespaceCommands() []*cli.Command {
 			},
 		},
 		{
-			Name:      "register",
```
robholland, line 36: We should probably add `register` as an alias to avoid breaking existing docs etc.

namespace/namespace.go
```diff
@@ -33,7 +33,7 @@ func NewNamespaceCommands() []*cli.Command {
...
+			Name:      "create",
 			Usage:     "Register a new Namespace",
 			Flags:     registerNamespaceFlags,
```
robholland, line 38: Flags variable should be renamed to `createNamespaceFlags`

robholland (changes requested):

namespace/namespace.go
feedmeapples: good point, updated

namespace/namespace.go
feedmeapples: updated

robholland (approved):

PR #93, opened 2023-02-02
Persist cluster Id when running server in-memory

feedmeapples:
## What was changed

Persists Cluster ID between dev server in-memory runs

relies on https://github.com/temporalio/temporal/pull/3883

## Why?

consistent Cluster ID between runs

## Checklist

1. Closes 

2. How was this tested:

manually, WIP unit tests

3. Any docs updates needed?


server/commands.go
```diff
@@ -274,6 +276,20 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
 				}
 				if c.String(common.FlagDBPath) == "" {
 					opts = append(opts, WithPersistenceDisabled())
+
+					if clusterCfg, err := cliconfig.NewConfig("temporalio", "cluster"); err == nil {
```
bergundy, line 280: Maybe call this:

```suggestion
					if clusterCfg, err := cliconfig.NewConfig("temporalio", "server"); err == nil {
```

bergundy: So we can add more server configs there if we want.

bergundy: Oh and it could be better to have the same structure as the server config.

server/commands.go
```diff
@@ -274,6 +276,20 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
 				}
 				if c.String(common.FlagDBPath) == "" {
 					opts = append(opts, WithPersistenceDisabled())
+
+					if clusterCfg, err := cliconfig.NewConfig("temporalio", "temporal-version-info"); err == nil {
```
bergundy, line 280: Please add some comments in this block to explain that we're explicitly ignoring errors.

server/commands.go
feedmeapples: i'll look into reusing server's config structure

bergundy: On second thought, having a separate file just for this is better, now that you changed the name I'm okay with the PR as is.

bergundy (approved):
Had one minor comment.
Otherwise LGTM.

server/commands.go
bergundy, line 280: nit:

```suggestion
					if clusterCfg, err := cliconfig.NewConfig("temporalio", "version-info"); err == nil {
```

PR #96, opened 2023-02-03
Upgrade server to v1.19.1@master

feedmeapples:
## What was changed

Upgraded Temporal server dependency to the latest commit

## Why?

addressing breaking changes so https://github.com/temporalio/temporal/pull/3883 can be imported

## Checklist

1. Closes 

2. How was this tested:

unit tests + ran server manually

3. Any docs updates needed?


server/config/config.go
```diff
@@ -198,10 +198,11 @@ func Convert(cfg *Config) *config.Config {
...
 	}
 	baseConfig.Services = map[string]config.Service{
-		"frontend": cfg.mustGetService(0),
-		"history":  cfg.mustGetService(1),
-		"matching": cfg.mustGetService(2),
-		"worker":   cfg.mustGetService(3),
+		"frontend":          cfg.mustGetService(0),
+		"history":           cfg.mustGetService(1),
+		"matching":          cfg.mustGetService(2),
+		"internal-frontend": cfg.mustGetService(3),
```
rodrigozhou, line 204: Wasn't this removed in more recent PR? Do we need it here? cc: @dnr

dnr: I'm not 100% sure what's going on here.. this is programatically building the static config for an in-memory/sqlite server? In that case, don't bother with internal-frontend, it's unnecessary (unless you're using an Authorizer, which you're not, right?)

feedmeapples: yea it's not used. However i get a server error if internal-frontend is not set

feedmeapples:
```
...
received non-nil error from function "go.temporal.io/server/temporal".ServerOptionsProvider
	/home/player0/go/pkg/mod/go.temporal.io/server@v1.18.1-0.20230203202137-a4943bbd2f2f/temporal/fx.go:159:
config validation error: "internal-frontend" service is missing in config
```

dnr: in `server/server.go`, change
```
		temporal.ForServices(temporal.Services),
```
to
```
		temporal.ForServices(temporal.DefaultServices),
```

dnr: also, I'd drop the Authorizer and ClaimMapper stuff from that file. you shouldn't be using those in a local server

feedmeapples: ok thank you!

feedmeapples: what if we are running e2e tests and want to check the UI behavior when the authrorizer denies a request

go.mod
```diff
@@ -12,30 +12,29 @@ require (
...
 	github.com/temporalio/tctl-kit v0.0.0-20230104170414-10932650d727
-	github.com/temporalio/ui-server/v2 v2.9.2-0.20230106074906-529addb76f82
+	github.com/temporalio/ui-server/v2 v2.10.1
 	github.com/urfave/cli/v2 v2.24.2
-	go.temporal.io/api v1.15.0
-	go.temporal.io/sdk v1.20.0
-	go.temporal.io/server v1.19.1
+	go.temporal.io/api v1.15.1-0.20230130221739-35f91d43296f
+	go.temporal.io/sdk v1.20.1-0.20230131233224-093eabe1f8d1
+	go.temporal.io/server v1.18.1-0.20230203000623-e43e5b500d81
```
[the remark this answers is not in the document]
feedmeapples: done

server/config/config.go
```diff
@@ -201,7 +201,7 @@ func Convert(cfg *Config) *config.Config {
 		"frontend": cfg.mustGetService(0),
 		"history":  cfg.mustGetService(1),
 		"matching": cfg.mustGetService(2),
-		"worker":   cfg.mustGetService(3),
+		"worker":   cfg.mustGetService(6),
```
rodrigozhou, line 204: Revert this?

server/config/config.go
dnr: :shrug: I guess there's no harm, but it seems hard to do an effective test using only the builtin choices of authorizer+claimmapper. I would want to provide a dummy impl that does what I want, which means I'd have to rebuild the binary and override those bits anyway

server/config/config.go
feedmeapples: sure (took the value it from temporal server)

feedmeapples, line 204:
```suggestion
		"worker":   cfg.mustGetService(3),
```

feedmeapples: reverted

though i wouldn't mind having the same as the default params in Temporal server repo (to address a niche corner case when Temporal repo server starts and this one would complain because of a taken port)

rodrigozhou (approved):

PR #100, opened 2023-02-07
Pass namespace option to search attributes operations

rodrigozhou:
## What was changed

Added `namespace` option to search attribute operations.

## Why?

Passing the namespace now is mandatory for creating search attributes when using SQL DB for visibility.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


searchattribute/search_attribute.go
```diff
@@ -16,6 +16,13 @@ func NewSearchAttributeCommands() []*cli.Command {
 			Usage: common.CreateSearchAttributeDefinition,
 			UsageText: common.SearchAttributeCreateUsageText,
 			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     common.FlagNamespace,
+					Aliases:  common.FlagNamespaceAlias,
+					Required: true,
+					Usage:    "Temporal namespace",
+					Category: common.CategoryMain,
+				},
```
rodrigozhou, line 25: I added the flag here because I could get from the global flags. Unsure which way is better. cc: @feedmeapples

feedmeapples: there are no more global flags, they have been moved under almost each individual command including the `search-attribute` ones

```
temporal operator search-attribute create -h                                                                                                                                                                                                                                                                                                                        

OPTIONS:
   Client Options:
....
   --namespace value, -n value              Temporal workflow namespace (default: "default") ....

   Display Options:
...
```

rodrigozhou: Oh, so I don't need to add this flag here. Gonna remove it.

PR #126, opened 2023-02-23
Fix batch operations when advanced visibility is not used

feedmeapples:
## What was changed

Removed reliance on `CountWorkflow` API when starting batch jobs

## Why?

Fixes batch jobs for cases when ES is disabled (including dev server)

## Checklist

1. Closes 

2. How was this tested:

manually. requires dev server e2e setup for proper tests 

3. Any docs updates needed?

### DIFF
--- a/batch/batch_commands.go
+++ b/batch/batch_commands.go
@@ -156,7 +156,7 @@ func startBatchJob(c *cli.Context, req *workflowservice.StartBatchOperationReque
 	if err == nil {
 		promptMsg = fmt.Sprintf("Will start a batch job operating on %v Workflow Executions. Continue? Y/N", color.Yellow(c, "%v", count.GetCount()))
 	} else {
-		promptMsg = fmt.Sprintf("Will start a batch job on all Workflow Executions that match %s. Continue? Y/N", color.Yellow(c, "%v", query))
+		promptMsg = fmt.Sprintf("Unable to count impacted workflows: %s. Will start a batch job on all Workflow Executions that match %s. Continue? Y/N", err, color.Yellow(c, "%v", query))
 	}
 
 	if !common.PromptYes(promptMsg, c.Bool(common.FlagYes)) {
--- a/batch/batch_commands.go
+++ b/batch/batch_commands.go
@@ -151,14 +151,14 @@ func startBatchJob(c *cli.Context, req *workflowservice.StartBatchOperationReque
 		Namespace: namespace,
 		Query:     query,
 	})
-	if err != nil {
-		return fmt.Errorf("unable to count impacted workflows: %w", err)
+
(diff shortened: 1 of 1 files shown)


batch/batch_commands.go
```diff
@@ -151,14 +151,14 @@ func startBatchJob(c *cli.Context, req *workflowservice.StartBatchOperationReque
...
 		Query:     query,
 	})
-	if err != nil {
-		return fmt.Errorf("unable to count impacted workflows: %w", err)
+
+	var promptMsg string
+	if err == nil {
+		promptMsg = fmt.Sprintf("Will start a batch job operating on %v Workflow Executions. Continue? Y/N", color.Yellow(c, "%v", count.GetCount()))
+	} else {
+		promptMsg = fmt.Sprintf("Will start a batch job on all Workflow Executions that match the %s. Continue? Y/N", color.Yellow(c, "%v", query))
```
bergundy, line 159: Hmm.. I didn't know we were prompting here.
Is there a way to disable the prompt?

bergundy:
Can we add a test for this please?

batch/batch_commands.go
feedmeapples, line 159:
```suggestion
		promptMsg = fmt.Sprintf("Will start a batch job on all Workflow Executions that match %s. Continue? Y/N", color.Yellow(c, "%v", query))
```

feedmeapples: it can be disabled if the `--yes` flag is passed

batch/batch_commands.go
```diff
@@ -151,14 +151,14 @@ func startBatchJob(c *cli.Context, req *workflowservice.StartBatchOperationReque
...
 		Query:     query,
 	})
-	if err != nil {
-		return fmt.Errorf("unable to count impacted workflows: %w", err)
+
+	var promptMsg string
+	if err == nil {
+		promptMsg = fmt.Sprintf("Will start a batch job operating on %v Workflow Executions. Continue? Y/N", color.Yellow(c, "%v", count.GetCount()))
+	} else {
+		promptMsg = fmt.Sprintf("Will start a batch job on all Workflow Executions that match %s. Continue? Y/N", color.Yellow(c, "%v", query))
```
alexshtin, line 159:
```suggestion
		promptMsg = fmt.Sprintf("Unable to count impacted workflows: %s. Will start a batch job on all Workflow Executions that match %s. Continue? Y/N", err, color.Yellow(c, "%v", query))
```

alexshtin (approved):

alexshtin:
Just FYI, @rodrigozhou added support for `Count` to SQL based advanced visibility. Standard visibility will be deprecated soon and `Count` should be always supported.

Issue #43, opened 2022-12-01
[Bug] Hitting Ctrl-C causes dev server to exit with RC 1 and print an error


bergundy:
Output after hitting Ctrl-C:
```
Error: All services are stopped.
('export TEMPORAL_CLI_SHOW_STACKS=1' to see stack traces)
```

PR #127, opened 2023-02-24
Fix exit code when Ctrl+C'ing dev server

feedmeapples:
## What was changed

Made `Ctrl+C` exit code 0 instead of 1 when running dev server 

## Why?

proper RC

address #43 

## Checklist

1. Closes 

2. How was this tested:

```
temporal server start-dev
```

Hit `Ctrl+C`

```
echo $?
0
```

3. Any docs updates needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Binaries for programs and plugins
 /temporal
+test-temporal
 /temporal-doc-gen
 *.exe
 *.exe~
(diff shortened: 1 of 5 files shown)


bergundy (commented):
Please test for both dev server and non dev server

feedmeapples:
> Please test for both dev server and non dev server

i've checked `temporalio/temporal` server and it properly returns RC 0. They use urfave/cli's default error handler `cli.HandleExitCoder(err)` which we now also use after logging custom messages

bergundy:
I understand but it would be awesome if we had automated tests for this

feedmeapples:
> I understand but it would be awesome if we had automated tests for this

Added a test

PR #129, opened 2023-03-01
Don't prompt when creating search attribute

feedmeapples:
## What was changed

Removed the prompt for `temporal operator search-attribute create`

## Why?

The prompt was rather unnecessary since the operation doesn't break/mutate existing data

## Checklist

1. Closes 

2. How was this tested:

manually

3. Any docs updates needed?


feedmeapples:
fyi `temporal operator search-attribute remove` still prompts

bergundy (approved):

PR #130, opened 2023-03-01
Increase persisence QPS

### LINKED ISSUE #255: [Feature Request] Server local dev mode shouldn't have low persitence QPS limits
If you hammer away at the server in `start-dev` mode, you can semi-easily run into things like

"Unable to read search attributes types: Persistence Max QPS Reached."

We should automatically set quite high dynamic config value for these IMO. In dev mode, there's no need to limit damage like this and it's just annoying if you're trying to run something intense.

feedmeapples:
## What was changed

Increased persistence QPS

## Why?

Address occasional UI errors such as 

image

## Checklist

1. Closes 

2. How was this tested:

WIP

3. Any docs updates needed?

Closes #255


feedmeapples:
fixing the failing CIs https://github.com/temporalio/cli/pull/257 and then merging this

PR #135, opened 2023-03-07
CLI: custom help template + minor fixes

DJSanti:
## What was changed
A custom help template has been implemented to ensure that the CLI help text prints correctly.

## Why?
This is a good idea for customizing how our helptext looks. Also, it prevents information that stays in documentation from being mistakenly printed.

## How was this tested:
Created binary, ran it.

3. Any docs updates needed?
Since this relates to the doc gen tool, documentation will be updated accordingly.

### DIFF
--- a/common/defs-cmds.go
+++ b/common/defs-cmds.go
@@ -223,9 +223,8 @@ These actions are performed through subcommands for each Operator area.
 To run an Operator command, run ` + "`" + `temporal operator [command] [subcommand] [command options] [arguments]` + "`" + `.
 `
 
-const ActivityUsageText = `Activity commands operate on [Activity Executions](/concepts/what-is-an-activity-execution).
-const ActivityUsageText = `Activity commands operate on [Activity Executions](/concepts/what-is-an-activity-execution).
-`
+const ActivityUsageText = `Activity commands operate on [Activity Executions](/concepts/what-is-an-activity-execution).`
+
 const CompleteActivityUsageText = `The ` + "`" + `temporal activity complete` + "`" + ` command completes an [Activity Execution](/concepts/what-is-an-activity-execution).
 
 Use the options listed below to change the behavior of this command.
@@ -247,6 +246,7 @@ Make sure to write the command as follows:
 ` + "`" + `temporal operator cluster health [command options] [arguments]` + "`" + `
 
 `
+
 const ClusterUsageText = `Cluster commands enabled operations on [Temporal Clusters](/concepts/what-is-a-temporal-cluster).`
 
 const ClusterDescribeUsageText = `The ` + "`" + `temporal operator cluster describe` + "`" + ` command shows information about the [Cluster](/concepts/what-is-a-temporal-cluster).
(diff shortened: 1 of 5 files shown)


common/defs-cmds.go
```diff
@@ -98,7 +98,7 @@ const DescribeBatchUsageText = `The `+"`"+`temporal batch describe`+"`"+` comman
 
 Use the command options listed below to change the information returned by this command.
 Make sure to write the command in this format:
-`+"`"+`temporal batch describe [command options] [arguments]`+"`"+`
+`+"`"+`temporal batch describe [command options]`+"`"+`
```
bergundy, line 101: This change applies to almost all of the commands

common/defs-cmds.go
```diff
@@ -223,7 +223,7 @@ These actions are performed through subcommands for each Operator area.
 To run an Operator command, run `+"`"+`temporal operator [command] [subcommand] [command options] [arguments]`+"`"+`.
 `
 
-const ActivityUsageText = `Activity commands enable operations on [Activity Executions](/concepts/what-is-an-activity-execution).
+const ActivityUsageText = `Activity commands operate on [Activity Executions](/concepts/what-is-an-activity-execution).
```
bergundy, line 226:
```suggestion
const ActivityUsageText = `Commands to operate on [Activity Executions](/concepts/what-is-an-activity-execution).
```

Make sure also that this isn't rendered in the CLI output, only for markdown.

common/defs-cmds.go
```diff
@@ -539,4 +539,13 @@ const StartDevUsageText = `The `+"`"+`temporal server start-dev`+"`"+` command s
 The results of any command run on the Server can be viewed at http://localhost:7233.
 `
 
-const CompletionText = "testing something"
\ No newline at end of file
+const CustomTemplateHelpCLI = `NAME:
```
bergundy, line 542: What does `CLI` in the variable name stand for?

bergundy (commented):
Hard to review without seeing this rendered.

common/defs-cmds.go
DJSanti: I needed to make the variable name stand out.

common/defs-cmds.go
DJSanti: Which ones stay the same, then? Would make it easier than striking all of them and then going back.

feedmeapples: the ones that use args:

```
temporal operator namespace describe
temporal operator namespace create
temporal operator namespace update
temporal operator namespace delete

temporal env get
temporal env set
```

The rest of the commands do not use args, only options

feedmeapples:
btw what code editor do you use? for some reason the code is not formatted. With vscode and go extension the code is auto-formatted

DJSanti:
> btw what code editor do you use? for some reason the code is not formatted. With vscode and go extension the code is auto-formatted
> 
> also adding linting so this is caught with CI

for some reason, pulling changes and fixing conflicts caused some errors in the constants that I was just able to sort out. I am on VSCode, so aside from being occupied last week and not being as diligent as I should, not sure what the heck happened there.

feedmeapples (approved):

PR #139, opened 2023-03-09
Test installation in zsh and fish shells

feedmeapples:
## What was changed

Testing install .sh script in zsh and fish

## Why?

testing zsh & fish, getting the green checkmark for CI 

## Checklist

1. Closes 

2. How was this tested:

CI

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,9 +83,8 @@ jobs:
           echo ${{ matrix.install-approach }}
           ${{ matrix.install-approach }}
 
-      # - name: Update SHELL env variable
-      #   run: "env SHELL=$(which ${{ matrix.shell }})"
+      - name: Update SHELL env variable
+        run: "env SHELL=$(which ${{ matrix.shell }})"
 
       - name: "Run Install Script"
-        shell: ${{ matrix.shell }}
         run: "./install.sh"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,46 +7,70 @@ on:
     branches: [main]
 
 jobs:
-  test:
-    strategy:
-      fail-fast: true
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    runs-on: ${{ matrix.os }}
-    steps:
-      - name: Print build information
-        run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}, os: ${{ matrix.os }}"
+  # test:
+  #   strategy:
+  #     fail-fast: true
+  #     matrix:
+  #       os: [windows-latest]
+  #   runs-on: ${{ matrix.os }}
+  #   steps:
+  #     - name: Print build information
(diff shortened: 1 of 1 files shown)


robholland (approved):

.github/workflows/ci.yml
```diff
@@ -39,20 +39,31 @@ jobs:
...
       - name: Print build information
         run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
 
       - uses: actions/checkout@v3
 
-      - name: "Run with ${{ matrix.shell }}"
-        run: sh -c '${{ matrix.shell }} -c "./install.sh"'
+      - name: Install ${{matrix.shell}} using apt-get
+        if: matrix.os == 'ubuntu-latest' && matrix.shell != 'sh'
+        run: "sudo apt-get install -y ${{ matrix.shell }}"
```
cretz, line 59: Can you do something at the end to confirm installed?

.github/workflows/ci.yml
```diff
@@ -39,20 +39,31 @@ jobs:
...
-            ignore_paths: ./.devcontainer
+          ignore_paths: ./.devcontainer
 
   test-install:
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        shell: [sh, bash, zsh]
+        shell: [sh, bash, zsh, fish]
```
cretz, line 49:
```suggestion
        include:
          - os: ubuntu-latest
            install-approach: apt
          - os: ubuntu-latest
            install-approach: sh
          - os: macos-latest
            install-approach: brew
```

Etcetera. You should consider using [includes](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude) here so you don't have a bunch of do-nothings like "windows fish".

cretz, line 49: You should not do these (just) by "shell" you should (also) do them also by "approach". In many cases the shell does not matter.

.github/workflows/ci.yml
```diff
@@ -39,20 +39,31 @@ jobs:
...
+      - name: Install ${{matrix.shell}} using apt-get
+        if: matrix.os == 'ubuntu-latest' && matrix.shell != 'sh'
+        run: "sudo apt-get install -y ${{ matrix.shell }}"
+
+      - name: Install ${{matrix.shell}} using Homebrew
+        if: matrix.os == 'macos-latest'  && matrix.shell != 'sh'
+        run: "brew update && brew install ${{ matrix.shell }}"
+
+      - name: Update SHELL env variable
+        run: "env SHELL=$(which ${{ matrix.shell }})"
```
cretz, line 66: Should we be using https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell instead?

feedmeapples: i've tried this though have been running into syntax issues with `shell: ${{ .. }}` approach. Hardcoding the shells worked

.github/workflows/ci.yml
feedmeapples: updated

PR #140, opened 2023-03-09
Remove testing install script for Windows

feedmeapples:
## What was changed

removed the Windows platform when testing install .sh script

## Why?

For Windows users we will recommend either downloading from the Github Releases or possibly Windows oriented package managers (e.g winget)  

## Checklist

1. Closes 

2. How was this tested:

CI

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


cretz:
Can merge with #139. Should just have a proper matrix using includes that has all permutations of "os" and "install approach" you want to confirm.

PR #153, opened 2023-03-15
passed custom template in BuildApp()

DJSanti:
## What was changed
A custom help template was massed in the arguments for BuildApp().

## Why?
The template prevents text meant for documentation (text with markdown) from being displayed in the CLI tool itself.

## Checklist
- Pass variable
- Compile and confirm that it works

## How was this tested:
Compiled app and ran commands and -h.

## Any docs updates needed?
Documents will be updated upon next generation.


bergundy:
Nice, we still have some work here but I'm no longer seeing markdown rendered.

I'm seeing `[arguments...]` show up in commands that don't support them, e.g:
```
   temporal task-queue describe [command options] [arguments...]
```

And very long lines that are hard to read and should be broken up, e.g:
```
   The Server records the last time of each poll request. Poll requests can last up to a minute, so a LastAccessTime under a minute is normal. If it's over a minute, then likely either the Worker is at capacity (all Workflow and Activity slots are full) or it has shut down. Once it has been 5 minutes since the last poll request, the Worker is removed from the list.
```

bergundy:
Hmm... nope still seeing markdown rendered:

```
$ temporal workflow --help
NAME:
   temporal workflow - Operations that can be performed on Workflows.

USAGE:
   Workflow commands allow operations to be performed on [Workflow Executions](/concepts/what-is-a-workflow-execution).
```

Please review all of the help outputs and verify we don't render markdown.
We should consider having a single help text in markdown form and rendering it as text when printed to console.

PR #166, opened 2023-03-22
Implement update/get worker build id commands

Sushisource:
Draft until https://github.com/temporalio/temporal/pull/4086 is merged and I can get rid of my go.mod replace

## What was changed
Added commands for updating and fetching worker-build-id compatibility sets on task queues.

## Why?
New API!

## Checklist

1. Closes https://github.com/temporalio/cli/issues/165

2. How was this tested:
Manually verified against new server

3. Any docs updates needed?
Yes. I'll be doing a docs PR as well.


headers/headers.go
```diff
@@ -15,8 +15,10 @@ const (
 	SupportedFeaturesHeaderDelim      = ","
 )
 
+const DEV_VERSION = "0.0.0-DEV"
+
 // Set by GoReleaser using ldflags
-var Version = "DEV"
+var Version = DEV_VERSION
```
Sushisource, line 21: I changed this because with this definition you can't simply `go run ./cmd/temporal` against a local dev server and have it work. It'll fail on parsing the version info.

If this is an issue for some reason let me know.

bergundy: @feedmeapples

bergundy (commented):
@Sushisource can you please add tests?
We can use the built-in dev server for that.

Other than than the command structure LGTM.
I mostly focused on that, not on the implementation.

headers/headers.go
feedmeapples: yea this change is good with me. We've recently discussed this with @mindaugasrukas . Could you remove the ldflag from makefile as well

Sushisource: Done

app/build_id_compat_cli_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	args, clientOpts := newServerAndClientOpts(port)
+
+	go func() {
+		if err := s.app.RunContext(ctx, args); err != nil {
+			fmt.Println("Server closed with error:", err)
+		}
+	}()
+
+	s.client = assertServerHealth(s.T(), ctx, clientOpts)
```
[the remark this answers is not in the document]
bergundy: It would require a compiled binary though.

common/defs-cmds.go
```diff
@@ -79,9 +81,19 @@ const (
 	ScheduleDescribeDefinition = "Get Schedule configuration and current state."
 	ScheduleDeleteDefinition   = "Deletes a Schedule."
 	ScheduleListDefinition     = "Lists Schedules."
+
+	// Update build id subcommand definitions
+	AddNewDefaultBuildIDDefinition         = "Add a new default (incompatible) build ID to the Task Queue version sets."
+	AddNewDefaultBuildIDDefinitionUsage    = "Creates a new build id set which will become the new overall default for the queue with the provided build id as its only member. This new set is incompatible with all previous sets/versions."
```
bergundy, line 87: nit:

```suggestion
	AddNewDefaultBuildIDDefinitionUsage    = "Creates a new build ID set which will become the new overall default for the queue with the provided build ID as its only member. This new set is incompatible with all previous sets/versions."
```

I would also see how this renders and consider adding line breaks.

bergundy (approved):
Some usage examples would be great too @Sushisource but it already LGTM

app/build_id_compat_cli_test.go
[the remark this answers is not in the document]
feedmeapples: done here https://github.com/temporalio/cli/tree/main/tests

headers/headers.go
feedmeapples: merged here https://github.com/temporalio/cli/pull/202

common/defs-cmds.go
Sushisource: Any semi-competent CLI flags library should be able to do this automatically (which this one maybe barely qualifies as lol). It looks like the one we're using has some kind of support for that: https://github.com/urfave/cli/pull/1119/files#diff-4d3f36d62cad94424544bb84fc93b33b73ffe3491a4ddca81a7f825539afc19cR1161-R1169

So we might consider doing that in another PR rather than manually inserting line breaks in a few places.

app/build_id_compat_cli_test.go
```diff
@@ -105,3 +105,11 @@ func (s *buildIdCompatSuite) TestPromoteBuildIdInSet() {
 		"--build-id", "foo"))
 	s.Nil(err)
 }
+
+func (s *buildIdCompatSuite) TestReachability() {
```
bergundy, line 109: Can we test the effect of this and not just that it doesn't return an error?

PR #170, opened 2023-03-25
Create integration tests suite

feedmeapples:
## What was changed

Create a suite for running integration tests

## Why?

adding integration tests

## Checklist

1. Closes 

2. How was this tested:

tested with History Replay integration test

3. Any docs updates needed?


robholland (commented):
Would be good to include at least one test which uses the server.

feedmeapples:
sure let's just follow up here with tests https://github.com/temporalio/cli/pull/171

PR #185, opened 2023-04-03
Add trace WorkflowExecutionUpdate and its worker

Sikian:
## What was changed
Add `GetWorkflowExecutionUpdates` function that uses `GetWorkflowHistory` to aggregate the workflow execution state. New updates can be retrieved using the HasNext()/Next() pattern, I hope I implemented this correctly.

This uses a workerpool to spawn new aggregation jobs for child workflows. I decided to use `pond` due to being able to easily wait for all jobs to complete (instead of closing and waiting as other workerpools do) and to easily propagate contexts and errors. Let me know if there's any problem with this.

This also includes adding a mutex lock to the `WorkflowExecutionState` maps storing the child workflows/activity/timers to avoid concurrent read/writes.

## Why?
This is the main engine of the workflow tracing. 

Next step will be the frontend, which uses `GetWorkflowExecutionUpdates` to get the latest `WorkflowExecutionState` and display it.

## Checklist

There are tests for `GetWorkflowExecutionUpdates` which cover the whole functionality. Let me know if you'd like more specific tests for the different components.

### DIFF
--- a/trace/execution_state.go
+++ b/trace/execution_state.go
@@ -25,9 +25,11 @@
 package trace
 
 import (
-	"context"
 	"fmt"
-	sdkclient "go.temporal.io/sdk/client"
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/failure/v1"
+	"go.temporal.io/api/history/v1"
 	"strconv"
 	"sync"
 	"time"
@@ -72,18 +74,17 @@ type WorkflowExecutionState struct {
 
 	// ChildStates contains all the ExecutionStates contained by this WorkflowExecutionState in order of execution.
 	ChildStates []ExecutionState
+	// childMapLock locks the state for read/write access. This is mainly used to avoid concurrent read/writes on the child maps (activityMap, childWorkflowMap and timerMap).
+	childMapLock sync.RWMutex
 	// activityMap contains all the activities executed in the Workflow, indexed by the EVENT_TYPE_ACTIVITY_TASK_SCHEDULED event id
 	// Used to retrieve the activities from events
-	activityMap     map[int64]*ActivityExecutionState
-	activityMapLock sync.RWMutex
-	// childWfMap contains all the child workflows executed in the Workflow, indexed by the EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED event id
+	activityMap map[int64]*ActivityExecutionState
+	// childWorkflowMap contains all the child workflows executed in the Workflow, indexed by the EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED event id
 	// Used to retrieve the child workflows from events
-	childWfMap     map[int64]*WorkflowExecutionState
-	childWfMapLock sync.RWMutex
+	childWorkflowMap map[int64]*WorkflowExecutionState
 	// timerMap contains all the timers executed in the Workflow, indexed by the EVENT_TYPE_TIMER_STARTED event id
 	// Used to retrieve the timers from events
-	timerMap     map[int64]*TimerExecutionState
-	timerMapLock sync.RWMutex
+	timerMap map[int64]*TimerExecutionState
 
 	// Non-successful closed states
 	// Failure contains the last failure that the Execution has reported (if any).
@@ -111,14 +112,6 @@ func NewWorkflowExecutionState(wfId, runId string) *WorkflowExecutionState {
 	return &WorkflowExecutionState{
 		Execution: &common.WorkflowExecution{WorkflowId: wfId, RunId: runId},
 	}
-}
-
-func GetWorkflowExecutionInfo(ctx context.Context, client sdkclient.Client, wfId, runId string) (*workflow.WorkflowExecutionInfo, error) {
-	res, err := client.DescribeWorkflowExecution(ctx, wfId, runId)
-	if err != nil {
-		return nil, err
-	}
-	return res.WorkflowExecutionInfo, nil
 }
 
 func (state *WorkflowExecutionState) GetName() string {
@@ -149,13 +142,10 @@ func (state *WorkflowExecutionState) GetDuration() *time.Duration {
 func (state *WorkflowExecutionState) newActivityFromEvent(event *history.HistoryEvent) *ActivityExecutionState {
 	if state.activityMap == nil {
 		state.activityMap = make(map[int64]*ActivityExecutionState)
-		state.activityMapLock = sync.RWMutex{}
 	}
 	activityState := NewActivityExecutionState()
 	activityState.Update(event)
 
-	state.activityMapLock.Lock()
-	defer state.activityMapLock.Unlock()
 	state.activityMap[event.EventId] = activityState
 	state.ChildStates = append(state.ChildStates, activityState)
 
@@ -164,8 +154,8 @@ func (state *WorkflowExecutionState) newActivityFromEvent(event *history.History
 
 // updateActivity updates a child ActivityExecutionState with a HistoryEvent by its scheduleId
 func (state *WorkflowExecutionState) updateActivity(scheduledId int64, event *history.HistoryEvent) {
-	state.activityMapLock.RLock()
-	defer state.activityMapLock.RUnlock()
+	state.childMapLock.RLock()
+	defer state.childMapLock.RUnlock()
 	if activityState, ok := state.activityMap[scheduledId]; ok {
 		activityState.Update(event)
 	}
@@ -173,27 +163,26 @@ func (state *WorkflowExecutionState) updateActivity(scheduledId int64, event *hi
 
 // newChildWorkflowFromEvent adds a new WorkflowExecutionState to the WorkflowExecutionState's ChildStates.
 func (state *WorkflowExecutionState) newChildWorkflowFromEvent(event *history.HistoryEvent) *WorkflowExecutionState {
-	if state.childWfMap == nil {
-		state.childWfMap = make(map[int64]*WorkflowExecutionState)
-		state.childWfMapLock = sync.RWMutex{}
+	if state.childWorkflowMap == nil {
+		state.childWorkflowMap = make(map[int64]*WorkflowExecutionState)
 	}
 	attrs := event.GetStartChildWorkflowExecutionInitiatedEventAttributes()
-	childWfState := NewWorkflowExecutionState(attrs.GetWorkflowId(), "")
-	childWfState.Type = attrs.GetWorkflowType()
+	childWorkflowState := NewWorkflowExecutionState(attrs.GetWorkflowId(), "")
+	childWorkflowState.Type = attrs.GetWorkflowType()
 
-	state.childWfMapLock.Lock()
-	defer state.childWfMapLock.Unlock()
-	state.childWfMap[event.EventId] = childWfState
-	state.ChildStates = append(state.ChildStates, childWfState)
+	state.childMapLock.Lock()
+	defer state.childMapLock.Unlock()
+	state.childWorkflowMap[event.EventId] = childWorkflowState
+	state.ChildStates = append(state.ChildStates, childWorkflowState)
 
-	return childWfState
+	return childWorkflowState
 }
 
 // GetChildWorkflowByEventId returns a child workflow for a given initiated event id
 func (state *WorkflowExecutionState) GetChildWorkflowByEventId(initiatedEventId int64) (*WorkflowExecutionState, bool) {
-	state.childWfMapLock.RLock()
-	defer state.childWfMapLock.RUnlock()
-	childWfState, ok := state.childWfMap[initiatedEventId]
+	state.childMapLock.RLock()
+	defer state.childMapLock.RUnlock()
+	childWfState, ok := state.childWorkflowMap[initiatedEventId]
 	return childWfState, ok
 }
 
@@ -201,13 +190,13 @@ func (state *WorkflowExecutionState) GetChildWorkflowByEventId(initiatedEventId 
 func (state *WorkflowExecutionState) newTimerFromEvent(event *history.HistoryEvent) *TimerExecutionState {
 	if state.timerMap == nil {
 		state.timerMap = make(map[int64]*TimerExecutionState)
-		state.timerMapLock = sync.RWMutex{}
+		state.childMapLock = sync.RWMutex{}
 	}
 	timerState := &TimerExecutionState{}
 	timerState.Update(event)
 
-	state.timerMapLock.Lock()
-	defer state.timerMapLock.Unlock()
+	state.childMapLock.Lock()
+	defer state.childMapLock.Unlock()
 	state.timerMap[event.EventId] = timerState
 	state.ChildStates = append(state.ChildStates, timerState)
 
@@ -216,26 +205,19 @@ func (state *WorkflowExecutionState) newTimerFromEvent(event *history.HistoryEve
 
 // updateTimer updates a child TimerExecutionState with a HistoryEvent by its startedId
 func (state *WorkflowExecutionState) updateTimer(startedId int64, event *history.HistoryEvent) {
-	state.timerMapLock.RLock()
-	defer state.timerMapLock.RUnlock()
+	state.childMapLock.RLock()
+	defer state.childMapLock.RUnlock()
 	if timerState, ok := state.timerMap[startedId]; ok {
 		timerState.Update(event)
 	}
 }
 
 // IsClosed returns true when the Workflow Execution is closed. A Closed status means that the Workflow Execution cannot make further progress.
-func (state *WorkflowExecutionState) IsClosed() bool {
-	switch state.Status {
-	case enums.WORKFLOW_EXECUTION_STATUS_CANCELED,
-		enums.WORKFLOW_EXECUTION_STATUS_COMPLETED,
-		enums.WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW,
-		enums.WORKFLOW_EXECUTION_STATUS_FAILED,
-		enums.WORKFLOW_EXECUTION_STATUS_TERMINATED,
-		enums.WORKFLOW_EXECUTION_STATUS_TIMED_OUT:
-		return true
-	default:
-		return false
+func (state *WorkflowExecutionState) IsClosed() (bool, error) {
+	if state.Status == enums.WORKFLOW_EXECUTION_STATUS_UNSPECIFIED {
+		return false, fmt.Errorf("workflow status in an unspecified state")
 	}
+	return state.Status != enums.WORKFLOW_EXECUTION_STATUS_RUNNING, nil
 }
 
 // Update updates the WorkflowExecutionState and its child states with a HistoryEvent.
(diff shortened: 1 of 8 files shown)


trace/workflow_execution_update.go
```diff
@@ -0,0 +1,115 @@
...
+
+// WorkflowExecutionUpdateIterator is the interface the provides iterative updates, analogous to the HistoryEventIterator interface.
+type WorkflowExecutionUpdateIterator interface {
+	HasNext() bool
+	Next() (*WorkflowExecutionUpdate, error)
+}
+
+// WorkflowExecutionUpdateIteratorImpl implements the iterator interface. Keeps information about the last processed update and
+// receives new updates through the updateChan channel.
+type WorkflowExecutionUpdateIteratorImpl struct {
```
bergundy, line 22: Is the separation of interface and implementation for mocking purposes?

trace/workflow_execution_update_test.go
```diff
@@ -0,0 +1,263 @@
...
+	suite.Run(t, new(WorkflowExecutionUpdateSuite))
+}
+
+func (s *WorkflowExecutionUpdateSuite) SetupTest() {
+	s.ctx = context.Background()
+}
+
+// SetDescribeWorkflowMocks mocks DescribeWorkflowExecution to return a list of events for a wfId/runId.
+// This is used in the WorkflowStateJob.Run method to retrieve the history length.
+func (s *WorkflowExecutionUpdateSuite) SetDescribeWorkflowMocks(client *mocks.Client, wfId, runId string, events []*history.HistoryEvent) {
```
bergundy, line 35: I'd prefer integration tests over unit tests with mocks to get better confidence that this code works with a real server.

trace/execution_state.go
```diff
@@ -45,15 +72,18 @@ type WorkflowExecutionState struct {
...
-	// childWfMap contains all the child workflows executed in the Workflow, indexed by the EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED event id.
-	// Used to retrieve the child workflows from events.
-	childWfMap map[int64]*WorkflowExecutionState
-	// timerMap contains all the timers executed in the Workflow, indexed by the EVENT_TYPE_TIMER_STARTED event id.
-	// Used to retrieve the timers from events.
-	timerMap map[int64]*TimerExecutionState
+	// activityMap contains all the activities executed in the Workflow, indexed by the EVENT_TYPE_ACTIVITY_TASK_SCHEDULED event id
+	// Used to retrieve the activities from events
+	activityMap     map[int64]*ActivityExecutionState
+	activityMapLock sync.RWMutex
```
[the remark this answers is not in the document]
bergundy: To be safe, I think we should lock the entire struct on any field access.
I haven't looked super closely at the access patterns but since this is all meant to be accessed concurrently, I'd rather be on the safe side.

bergundy (commented):
Thanks!
Overall LGTM.

trace/workflow_execution_update_test.go
Sikian: Good idea. Let me get to it 👍

trace/workflow_execution_update.go
Sikian: Not really. I was just following the iterator pattern but didn't have a better name for the implementation. Happy to change this if it's not the right approach.

trace/execution_state.go
```diff
@@ -83,6 +113,14 @@ func NewWorkflowExecutionState(wfId, runId string) *WorkflowExecutionState {
 	}
 }
 
+func GetWorkflowExecutionInfo(ctx context.Context, client sdkclient.Client, wfId, runId string) (*workflow.WorkflowExecutionInfo, error) {
```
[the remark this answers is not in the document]
Sikian: Probably not, I'll remove it.

trace/execution_state.go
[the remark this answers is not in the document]
Sikian: That makes sense, thanks for the suggestion.

trace/execution_state.go
```diff
@@ -45,15 +72,18 @@ type WorkflowExecutionState struct {
...
-	// Used to retrieve the timers from events.
-	timerMap map[int64]*TimerExecutionState
+	// activityMap contains all the activities executed in the Workflow, indexed by the EVENT_TYPE_ACTIVITY_TASK_SCHEDULED event id
+	// Used to retrieve the activities from events
+	activityMap     map[int64]*ActivityExecutionState
+	activityMapLock sync.RWMutex
+	// childWfMap contains all the child workflows executed in the Workflow, indexed by the EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED event id
+	// Used to retrieve the child workflows from events
+	childWfMap     map[int64]*WorkflowExecutionState
+	childWfMapLock sync.RWMutex
```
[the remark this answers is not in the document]
Sikian: Renamed this to `childWorkflowMap`.

trace/execution_state.go
```diff
@@ -107,14 +145,17 @@ func (state *WorkflowExecutionState) GetDuration() *time.Duration {
 	return getDuration(state.StartTime, state.CloseTime)
 }
 
-// newActivity adds a new ActivityExecutionState to the WorkflowExecutionState's ChildStates.
-func (state *WorkflowExecutionState) newActivity(event *history.HistoryEvent) *ActivityExecutionState {
+// newActivityFromEvent adds a new ActivityExecutionState to the WorkflowExecutionState's ChildStates.
+func (state *WorkflowExecutionState) newActivityFromEvent(event *history.HistoryEvent) *ActivityExecutionState {
 	if state.activityMap == nil {
 		state.activityMap = make(map[int64]*ActivityExecutionState)
+		state.activityMapLock = sync.RWMutex{}
```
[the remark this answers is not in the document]
Sikian: Wasn't aware of this, thanks!

trace/execution_state.go
```diff
@@ -107,14 +145,17 @@ func (state *WorkflowExecutionState) GetDuration() *time.Duration {
...
 	}
 	activityState := NewActivityExecutionState()
 	activityState.Update(event)
 
+	state.activityMapLock.Lock()
+	defer state.activityMapLock.Unlock()
```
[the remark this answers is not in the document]
Sikian: The activity map is accessed in the following line.

trace/workflow_execution_update.go
```diff
@@ -0,0 +1,115 @@
+package trace
+
+import (
+	"context"
+	"fmt"
+	"github.com/alitto/pond"
```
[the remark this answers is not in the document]
Sikian: I'm not sure if there's a vanilla workerpool. I'm happy to implement it myself (I did this initially) but I believe it might be better in the long run to use a well-tested workerpool.

trace/execution_state.go
```diff
@@ -169,18 +216,26 @@ func (state *WorkflowExecutionState) newTimer(event *history.HistoryEvent) *Time
...
-// IsCompleted returns true when the Workflow Execution is completed in a non-failed state.
-// This is useful to know if we should fetch child workflows or fold the information.
-// For now this is when the workflow is completed, terminated or canceled.
-func (state *WorkflowExecutionState) IsCompleted() bool {
-	return state.Status == enums.WORKFLOW_EXECUTION_STATUS_COMPLETED ||
-		state.Status == enums.WORKFLOW_EXECUTION_STATUS_CANCELED ||
-		state.Status == enums.WORKFLOW_EXECUTION_STATUS_TERMINATED
+// IsClosed returns true when the Workflow Execution is closed. A Closed status means that the Workflow Execution cannot make further progress.
+func (state *WorkflowExecutionState) IsClosed() bool {
+	switch state.Status {
```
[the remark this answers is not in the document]
Sikian: Good point.

trace/workflow_execution_update.go
```diff
@@ -0,0 +1,115 @@
...
+		return true
+	}
+}
+
+// Next return the last processed execution update. HasNext has to be called first (following the HasNext/Next pattern).
+func (iter *WorkflowExecutionUpdateIteratorImpl) Next() (*WorkflowExecutionUpdate, error) {
+	// Make sure HasNext() has been called first.
+	if !iter.updated {
+		return nil, fmt.Errorf("please call HasNext() first")
+	}
```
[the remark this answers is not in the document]
Sikian: I think HasNext() doesn't return any error for `HistoryEventIterator`. However, I can see that that implementation will advance the index every time `Next()` is called rather than in `HasNext()`. I can change that if that makes more sense.

trace/workflow_execution_update_test.go
Sikian: I've been giving this some thought and I'm not sure I understand what you're trying to convey. Is your suggestion to setup a server with fixture responses and test against that?

If that is the case: while I agree that a server would make these tests closer to real, I think leveraging the sdk mocks can make the tests easier to maintain (no server to maintain) and we can get exactly the response we want for each test. Additionally, failures in `DescribeWorkflowExecution` and `GetWorkflowHistory` in the sdk would bleed into the tests here.

In any case, I'm happy to review this and add more tests to better cover this functionality.

trace/execution_state.go
```diff
@@ -45,14 +74,16 @@ type WorkflowExecutionState struct {
 
 	// ChildStates contains all the ExecutionStates contained by this WorkflowExecutionState in order of execution.
 	ChildStates []ExecutionState
-	// activityMap contains all the activities executed in the Workflow, indexed by the EVENT_TYPE_ACTIVITY_TASK_SCHEDULED event id.
-	// Used to retrieve the activities from events.
+	// childMapLock locks the state for read/write access. This is mainly used to avoid concurrent read/writes on the child maps (activityMap, childWorkflowMap and timerMap).
+	childMapLock sync.RWMutex
```
bergundy, line 78: Are these the only properties that will be accessed concurrently?
If you're certain of this, let's keep the way it is but otherwise, I'd put a lock on the entire struct.

trace/execution_state.go
```diff
@@ -107,8 +138,8 @@ func (state *WorkflowExecutionState) GetDuration() *time.Duration {
 	return getDuration(state.StartTime, state.CloseTime)
 }
 
-// newActivity adds a new ActivityExecutionState to the WorkflowExecutionState's ChildStates.
-func (state *WorkflowExecutionState) newActivity(event *history.HistoryEvent) *ActivityExecutionState {
+// newActivityFromEvent adds a new ActivityExecutionState to the WorkflowExecutionState's ChildStates.
+func (state *WorkflowExecutionState) newActivityFromEvent(event *history.HistoryEvent) *ActivityExecutionState {
```
bergundy, line 142: Don't we need to lock the map here?

trace/execution_state.go
```diff
@@ -123,44 +154,49 @@ func (state *WorkflowExecutionState) newActivity(event *history.HistoryEvent) *A
...
 	attrs := event.GetStartChildWorkflowExecutionInitiatedEventAttributes()
-	childWfState := NewWorkflowExecutionState(attrs.GetWorkflowId(), "")
-	childWfState.Type = attrs.GetWorkflowType()
+	childWorkflowState := NewWorkflowExecutionState(attrs.GetWorkflowId(), "")
+	childWorkflowState.Type = attrs.GetWorkflowType()
 
-	state.childWfMap[event.EventId] = childWfState
-	state.ChildStates = append(state.ChildStates, childWfState)
+	state.childMapLock.Lock()
+	defer state.childMapLock.Unlock()
```
bergundy, line 174: You should move the lock to the top of this function.

trace/execution_state.go
```diff
@@ -123,44 +154,49 @@ func (state *WorkflowExecutionState) newActivity(event *history.HistoryEvent) *A
...
+	childWfState, ok := state.childWorkflowMap[initiatedEventId]
+	return childWfState, ok
+}
+
+// newTimerFromEvent adds a new TimerExecutionState to the WorkflowExecutionState's ChildStates.
+func (state *WorkflowExecutionState) newTimerFromEvent(event *history.HistoryEvent) *TimerExecutionState {
 	if state.timerMap == nil {
 		state.timerMap = make(map[int64]*TimerExecutionState)
+		state.childMapLock = sync.RWMutex{}
 	}
```
bergundy, line 194: Don't need all of this anymore.

trace/execution_state.go
```diff
@@ -123,44 +154,49 @@ func (state *WorkflowExecutionState) newActivity(event *history.HistoryEvent) *A
...
+	defer state.childMapLock.RUnlock()
+	childWfState, ok := state.childWorkflowMap[initiatedEventId]
+	return childWfState, ok
+}
+
+// newTimerFromEvent adds a new TimerExecutionState to the WorkflowExecutionState's ChildStates.
+func (state *WorkflowExecutionState) newTimerFromEvent(event *history.HistoryEvent) *TimerExecutionState {
 	if state.timerMap == nil {
 		state.timerMap = make(map[int64]*TimerExecutionState)
+		state.childMapLock = sync.RWMutex{}
```
bergundy, line 193: Leftover?

trace/execution_state.go
```diff
@@ -275,10 +312,10 @@ func (state *WorkflowExecutionState) Update(event *history.HistoryEvent) {
 	// CHILD WORKFLOW EVENTS
 	case enums.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED:
 		// First child workflow
-		state.newChildWorkflow(event)
+		state.newChildWorkflowFromEvent(event)
 	case enums.EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_STARTED:
 		attrs := event.GetChildWorkflowExecutionStartedEventAttributes()
-		if child, ok := state.childWfMap[attrs.InitiatedEventId]; ok {
+		if child, ok := state.childWorkflowMap[attrs.InitiatedEventId]; ok {
```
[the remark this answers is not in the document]
bergundy: This comment applies to anywhere you access the maps in this function.

trace/workflow_execution_update.go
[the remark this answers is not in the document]
bergundy: I was thinking just use goroutines and a semaphore.

trace/workflow_execution_update.go
bergundy: Let's merge the interface and implementation.

trace/workflow_execution_update.go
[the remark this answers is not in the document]
bergundy: Yeah, you could panic to be on the safe side but it's fine for internal use.

trace/workflow_execution_update_test.go
bergundy: My suggestion is that we test real server responses, that would mean starting a workflow and getting it to a specific state.

I think these unit tests are fine but I'd want at least a single integration test, can wait for when we implement the display logic.

trace/workflow_state_worker.go
```diff
@@ -0,0 +1,139 @@
...
+	if state == nil {
+		return nil, errors.New("workflow state cannot be nil for a workflow state job")
+	}
+	if updateChan == nil {
+		return nil, errors.New("updateChan cannot be nil for a workflow state job")
+	}
+
+	// Get workflow execution's description, so we can know if we're up-to-date. Doing this synchronously will allow us to correctly
+	// assess how many events need to be processed (otherwise only the ones from the root workflow will be counted).
+	// We don't mind if this fails, since HistoryLength is used only to display event processing progress.
```
bergundy, line 38: I'd rather just return the error here instead of silently ignoring it. There's no good reason for this call to fail and if it does it'll be hard to debug.

trace/workflow_state_worker.go
```diff
@@ -0,0 +1,139 @@
...
+
+		return nil
+	}
+}
+
+// GetChildJob gets a new child job and appends it to the list of childJobs. These jobs don't start until the parent is catched up.
+func (job *WorkflowStateJob) GetChildJob(event *history.HistoryEvent) (*WorkflowStateJob, error) {
+	// Retrieve child workflow from parent and create a new job to fetch events for it
+	childAttrs := event.GetChildWorkflowExecutionStartedEventAttributes()
+	wf, ok := job.state.GetChildWorkflowByEventId(childAttrs.GetInitiatedEventId())
```
bergundy, line 114: Looks like `wf` is accessed concurrently, so I'm now convinced we should have a lock for the entire struct.

Sikian: That's what the lock in `GetChildWorkflowByEventId` is for.

trace/workflow_state_worker.go
Sikian: I disagree, but I'll update it as required.

trace/workflow_execution_update_test.go
Sikian: That's a significant investment into some tests which will probably not provide better coverage. If you believe it's going to add value, I invite you to implement it in the rest of the tool.

trace/execution_state.go
Sikian: There's no problem with properties being accessed concurrently, only maps.

trace/execution_state.go
Sikian: Done 👍

trace/execution_state.go
Sikian: Removed, thanks.

trace/execution_state.go
Sikian: Done, thanks 👍

trace/workflow_state_worker.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// WorkflowStateJob implements a WorkerJob to retrieve updates for a WorkflowExecutionState and its child workflows.
+type WorkflowStateJob struct {
+	ctx    context.Context
+	client sdkclient.Client
+
+	state      *WorkflowExecutionState
+	depth      int
+	fetchAll   bool
```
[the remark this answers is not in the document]
Sikian: Added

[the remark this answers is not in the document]
Sikian: Done 👍

trace/execution_state.go
bergundy: It depends if we care about multiple properties being updated atomically.
I generally would like to make less assumptions and be on the safe side.

trace/workflow_execution_update_test.go
bergundy: Yes, we need more integration coverage in general but I won't block the PR on this.

trace/workflow_state_worker.go
bergundy: Yes but then you pass `wf` to `NewWorkflowStateJob` unlocked.

Again this might not matter today since we update a couple of attributes (`HistoryLength` and `IsArchived`) that may not need to be atomically updated but, as mentioned above, I'd rather be on the safe side and make sure all access to this struct is serialized.

trace/workflow_state_worker.go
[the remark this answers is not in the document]
Sikian: `	foldStatus []enums.WorkflowExecutionStatus`

trace/workflow_execution_update.go
bergundy: Not blocking but would be nice to address before merging.

trace/workflow_state_worker.go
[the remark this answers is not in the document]
bergundy: So what's `fetchAll` for?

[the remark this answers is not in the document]
bergundy: I see, I thought we agreed to remove the `--all` flag in favor of the individual statuses.

trace/workflow_state_worker.go
bergundy: Each state struct would have its own lock, you wouldn't hold the same lock for all workflows.
Again, I'd rather have code that makes less assumptions about concurrent usage but I think we can move on to the next PR.

bergundy (approved):
Left a few more comments.
Can address in the next PR.
Let's merge this and move forward.

trace/workflow_state_worker.go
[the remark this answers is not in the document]
Sikian: We're also using the `--no-fold` to disable the behavior: https://github.com/temporalio/cli/blob/main/workflow/workflow.go#L381-L385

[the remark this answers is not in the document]
bergundy: Right, I forgot that what we settled on.

bergundy:
@feedmeapples looks like CI is failing on test-install, I'm merging this as the failures aren't related to the change.

PR #200, opened 2023-04-18
Add command for workflow update 

samanbarghi:
## What was changed
Added 'workflow update' command to update workflows.

## Why?

There is a new feature to update workflows synchronously: https://github.com/temporalio/temporal/pull/3822

This change adds a command to the client to call that API.

## Checklist

1. How was this tested:
Tested locally.

2. Any docs updates needed?

The "workflow" section of CLI doc should be updated: https://docs.temporal.io/cli/workflow

### DIFF
--- a/common/defs-cmds.go
+++ b/common/defs-cmds.go
@@ -25,6 +25,7 @@ const (
 	DeleteWorkflowDefinition    = "Deletes a Workflow Execution."
 	ResetWorkflowDefinition     = "Resets a Workflow Execution by Event ID or reset type."
 	TraceWorkflowDefinition     = "Trace progress of a Workflow Execution and its children."
+	UpdateWorkflowDefinition    = "Updates a running workflow synchronously."
 
 	// Activity subcommand definitions
 	CompleteActivityDefinition = "Completes an Activity Execution."
@@ -196,7 +197,6 @@ Workflow terminations require a valid [Workflow ID](/concepts/what-is-a-workflow
 ` + "`" + `temporal workflow terminate --workflow-id=meaningful-business-id` + "`" + `
 
 Use the options listed below to change termination behavior.`
-
 
 const ResetWorkflowUsageText = `The ` + "`" + `temporal workflow reset` + "`" + ` command resets a [Workflow Execution](/concepts/what-is-a-workflow-execution).
 A reset allows the Workflow to resume from a certain point without losing its parameters or [Event History](/concepts/what-is-an-event-history).
@@ -544,6 +544,10 @@ const WorkflowDeleteUsageText = `The ` + "`" + `temporal workflow delete` + "`" 
 Use the options listed below to change the command's behavior.`
 
 const WorkflowTraceUsageText = `The ` + "`" + `temporal workflow trace` + "`" + ` command tracks the progress of a [Workflow Execution](/concepts/what-is-a-workflow-execution) and any  [Child Workflows](/concepts/what-is-a-child-workflow-execution) it generates.
+
+Use the options listed below to change the command's behavior.`
+
+const WorkflowUpdateUsageText = `The ` + "`" + `temporal workflow update` + "`" + ` command synchronously updates a running [Workflow Execution](/concepts/what-is-a-workflow-execution).
 
 Use the options listed below to change the command's behavior.`
 
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -60,6 +60,7 @@ const (
 	FlagInputFileSignal            = "Input for the Signal from file. Formatted in JSON."
 	FlagUpdateHandlerName          = "Update handler Name"
 	FlagUpdateHandlerInput         = "Args for the Update handler. Formatted in JSON."
+	FlagUpdateIDDefinition         = "UpdateID to check the result of an update (either UpdateID or Update handler name should be passed)"
 	FlagCancelWorkflow             = "Cancel Workflow Execution with given Workflow Id."
 	FlagWorkflowIDTerminate        = "Terminate Workflow Execution with given Workflow Id."
 	FlagQueryTerminate             = "Terminate Workflow Executions with given List Filter."
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -5,7 +5,7 @@ const (
 	FlagEnvDefinition            = "Environment to read environmental variables from."
 	FlagAddrDefinition           = "The host and port (formatted as host:port) for the Temporal Frontend Service."
 	FlagNSAliasDefinition        = "Identifies a Namespace in the Temporal Workflow."
-	FlagMetadataDefinition       = "Contains gRPC metadata to send with requests (formatted as key=value). Values must be in a valid JSON format."
+	FlagMetadataDefinition       = "Contains gRPC metadata to send with requests (formatted as key=value)."
 	FlagTLSCertPathDefinition    = "Path to x509 certificate."
 	FlagTLSKeyPathDefinition     = "Path to private certificate key."
 	FlagTLSCaPathDefinition      = "Path to server CA certificate."
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -1,131 +1,134 @@
 package common
+
 const (
 	// Shared flag definitions
-	FlagEnvDefinition = "Environment to read environmental variables from."
-	FlagAddrDefinition = "The host and port (formatted as host:port) for the Temporal Frontend Service."
-	FlagNSAliasDefinition = "Identifies a Namespace in the Temporal Workflow."
-	FlagMetadataDefinition = "Contains gRPC metadata to send with requests (formatted as key=value). Values must be in a valid JSON format."
-	FlagTLSCertPathDefinition = "Path to x509 certificate."
-	FlagTLSKeyPathDefinition = "Path to private certificate key."
-	FlagTLSCaPathDefinition = "Path to server CA certificate."
-	FlagTLSDisableHVDefinition = "Disables TLS host-name verification."
-	FlagTLSServerNameDefinition = "Overrides target TLS server name."
+	FlagEnvDefinition            = "Environment to read environmental variables from."
+	FlagAddrDefinition           = "The host and port (formatted as host:port) for the Temporal Frontend Service."
+	FlagNSAliasDefinition        = "Identifies a Namespace in the Temporal Workflow."
+	FlagMetadataDefinition       = "Contains gRPC metadata to send with requests (formatted as key=value). Values must be in a valid JSON format."
+	FlagTLSCertPathDefinition    = "Path to x509 certificate."
+	FlagTLSKeyPathDefinition     = "Path to private certificate key."
+	FlagTLSCaPathDefinition      = "Path to server CA certificate."
+	FlagTLSDisableHVDefinition   = "Disables TLS host-name verification."
+	FlagTLSServerNameDefinition  = "Overrides target TLS server name."
 	FlagContextTimeoutDefinition = "Optional timeout for the context of an RPC call (in seconds)."
-	FlagCodecEndpointDefinition = "Endpoint for a remote Codec Server."
-	FlagCodecAuthDefinition = "Sets the authorization header on requests to the Codec Server."
-	FlagArchiveDefinition = "List archived Workflow Executions. Currently an experimental feature."
+	FlagCodecEndpointDefinition  = "Endpoint for a remote Codec Server."
+	FlagCodecAuthDefinition      = "Sets the authorization header on requests to the Codec Server."
+	FlagArchiveDefinition        = "List archived Workflow Executions. Currently an experimental feature."
 
 	// Execution flags
-	FlagWorkflowId = "Workflow Id"
-	FlagRunIdDefinition = "Run Id"
-	FlagJobIDDefinition = "Batch Job Id"
+	FlagWorkflowId           = "Workflow Id"
+	FlagRunIdDefinition      = "Run Id"
+	FlagJobIDDefinition      = "Batch Job Id"
 	FlagScheduleIDDefinition = "Schedule Id"
 
 	// ShowWorkflow flags
-	FlagOutputFilenameDefinition = "Serializes Event History to a file."
-	FlagMaxFieldLengthDefinition = "Maximum length for each attribute field."
+	FlagOutputFilenameDefinition  = "Serializes Event History to a file."
+	FlagMaxFieldLengthDefinition  = "Maximum length for each attribute field."
 	FlagResetPointsOnlyDefinition = "Only show Workflow Events that are eligible for reset."
-	FlagFollowAliasDefinition = "Follow the progress of a Workflow Execution."
+	FlagFollowAliasDefinition     = "Follow the progress of a Workflow Execution."
 
 	// StartWorkflow flags
-	FlagWFTypeDefinition = "Workflow Type name."
-	FlagTaskQueueDefinition = "Task Queue"
-	FlagWorkflowRunTimeoutDefinition = "Timeout (in seconds) of a Workflow Run."
+	FlagWFTypeDefinition                   = "Workflow Type name."
+	FlagTaskQueueDefinition                = "Task Queue"
+	FlagWorkflowRunTimeoutDefinition       = "Timeout (in seconds) of a Workflow Run."
 	FlagWorkflowExecutionTimeoutDefinition = "Timeout (in seconds) for a WorkflowExecution, including retries and `ContinueAsNew` tasks."
-	FlagWorkflowTaskTimeoutDefinition = "Start-to-close timeout for a Workflow Task (in seconds)."
-	FlagCronScheduleDefinition = "Optional Cron Schedule for the Workflow. Cron spec is formatted as: \n" +
-	"\t┌───────────── minute (0 - 59) \n" +
-	"\t│ ┌───────────── hour (0 - 23) \n" +
-	"\t│ │ ┌───────────── day of the month (1 - 31) \n" +
-	"\t│ │ │ ┌───────────── month (1 - 12) \n" +
-	"\t│ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday) \n" +
-	"\t│ │ │ │ │ \n" +
-	"\t* * * * *"
+	FlagWorkflowTaskTimeoutDefinition      = "Start-to-close timeout for a Workflow Task (in seconds)."
+	FlagCronScheduleDefinition             = "Optional Cron Schedule for the Workflow. Cron spec is formatted as: \n" +
+		"\t┌───────────── minute (0 - 59) \n" +
+		"\t│ ┌───────────── hour (0 - 23) \n" +
+		"\t│ │ ┌───────────── day of the month (1 - 31) \n" +
+		"\t│ │ │ ┌───────────── month (1 - 12) \n" +
+		"\t│ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday) \n" +
+		"\t│ │ │ │ │ \n" +
+		"\t* * * * *"
 	FlagWorkflowIdReusePolicyDefinition = "Allows the same Workflow Id to be used in a new Workflow Execution. Options are: AllowDuplicate, AllowDuplicateFailedOnly, RejectDuplicate, TerminateIfRunning."
-	FlagInputDefinition = "Optional JSON input to provide to the Workflow. Pass \"null\" for null values."
-	FlagInputFileDefinition = "Passes optional input for the Workflow from a JSON file. If there are multiple JSON files, concatenate them and separate by space or newline. Input from the command line will overwrite file input."
-	FlagSearchAttributeDefinition = "Passes Search Attribute in key=value format. Use valid JSON formats for value."
-	FlagMemoDefinition = "Passes a memo in key=value format. Use valid JSON formats for value."
-	FlagMemoFileDefinition = "Passes a memo as file input, with each line following key=value format. Use valid JSON formats for value."
+	FlagInputDefinition                 = "Optional JSON input to provide to the Workflow. Pass \"null\" for null values."
+	FlagInputFileDefinition             = "Passes optional input for the Workflow from a JSON file. If there are multiple JSON files, concatenate them and separate by space or newline. Input from the command line will overwrite file input."
+	FlagSearchAttributeDefinition       = "Passes Search Attribute in key=value format. Use valid JSON formats for value."
+	FlagMemoDefinition                  = "Passes a memo in key=value format. Use valid JSON formats for value."
+	FlagMemoFileDefinition              = "Passes a memo as file input, with each line following key=value format. Use valid JSON formats for value."
 
 	// Other Workflow flags
-	FlagResetPointsUsage = "Only show auto-reset points."
-	FlagPrintRawUsage = "Print properties without changing their format."
-	QueryFlagTypeUsage = "The type of Query to run."
-	FlagWorkflowSignalUsage = "Signal Workflow Execution by Id."
-	FlagQueryDefinition = "Signal Workflow Executions by List Filter."
-	FlagSignalName = "Signal Name"
-	FlagInputSignal = "Input for the Signal. Formatted in JSON."
-	FlagInputFileSignal = "Input for the Signal from file. Formatted in JSON."
-	FlagCancelWorkflow = "Cancel Workflow Execution with given Workflow Id."
-	FlagWorkflowIDTerminate = "Terminate Workflow Execution with given Workflow Id."
-	FlagQueryTerminate = "Terminate Workflow Executions with given List Filter."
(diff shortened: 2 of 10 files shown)


common/defs-flags.go
```diff
@@ -1,131 +1,134 @@
...
+	FlagResetPointsUsage           = "Only show auto-reset points."
+	FlagPrintRawUsage              = "Print properties without changing their format."
+	QueryFlagTypeUsage             = "The type of Query to run."
+	FlagWorkflowSignalUsage        = "Signal Workflow Execution by Id."
+	FlagQueryDefinition            = "Signal Workflow Executions by List Filter."
+	FlagSignalName                 = "Signal Name"
+	FlagInputSignal                = "Input for the Signal. Formatted in JSON."
+	FlagInputFileSignal            = "Input for the Signal from file. Formatted in JSON."
+	FlagUpdateHandlerName          = "Update handler Name"
+	FlagUpdateHandlerInput         = "Args for the Update handler. Formatted in JSON."
```
samanbarghi, line 62: `gofmt` reformatted the whole file; these are the new additions.

workflow/workflow.go
```diff
@@ -385,5 +385,38 @@ func NewWorkflowCommands() []*cli.Command {
...
+				&cli.StringFlag{
+					Name:     common.FlagUpdateFirstExecutionRunID,
+					Usage:    "Do not run update if workflow exectuion (--run-id or if not set, latest workflow execution)  is not part of the same execution chain as this id.",
+					Category: common.CategoryMain,
+				},
+			),
+			Action: func(c *cli.Context) error {
+				return UpdateWorkflow(c)
+			},
+		},
```
samanbarghi, line 420: `gofmt` reformatted the file, here is the new addition.

workflow/workflow_commands.go
```diff
@@ -1445,6 +1451,56 @@ func TraceWorkflow(c *cli.Context) error {
...
+		return fmt.Errorf("must specify valid wait policy: %v", strings.Join(mapKeysToArray(updateWaitPolicyMap), ", "))
+	}
+	sdk, err := client.GetSDKClient(c)
+	if err != nil {
+		return err
+	}
+	ctx, cancel := common.NewContext(c)
+	defer cancel()
+
+	_, err = sdk.WorkflowService().UpdateWorkflowExecution(ctx, &workflowservice.UpdateWorkflowExecutionRequest{
```
cretz, line 1478: I think we may want to use the high level update calls available on the client. Also, we need to know whether the user wants to wait for update accepted by server, accepted by workflow, or completed. Also, if users do want to wait for completed, we need to show the result similar to how we do for queries.

We may need to step back and outline all of the workflow update commands (there will also be ones on getting the result of an already started update I would think).

samanbarghi: > I think we may want to use the high level update calls available on the client. 
Sounds good.

>  Also, we need to know whether the user wants to wait for update accepted by server, accepted by workflow, or completed. Also, if users do want to wait for completed, we need to show the result similar to how we do for queries.

Good point, I was just about to ask whether we want to process the result in any way, I check the queries and make some updates. 

> We may need to step back and outline all of the workflow update commands (there will also be ones on getting the result of an already started update I would think).

Sounds good to me.

tests/workflows/update/update.go
```diff
@@ -0,0 +1,44 @@
...
+	log := workflow.GetLogger(ctx)
+	counter := 0
+
+	if err := workflow.SetUpdateHandlerWithOptions(
+		ctx,
+		FetchAndAdd,
+		func(ctx workflow.Context, i int) (int, error) {
+			tmp := counter
+			counter += i
+			log.Info("counter updated", "addend", i, "new-value", counter)
```
bergundy, line 24:
```suggestion
			log.Info("counter updated", "added", i, "new-value", counter)
```

tests/workflow_test.go
```diff
@@ -51,3 +53,31 @@ func (s *e2eSuite) TestWorkflowShow_ReplayableHistory() {
...
+	s.NoError(err)
+	signalWorkflow := func() {
+		// send a Signal to stop the workflow
+		err = c.SignalWorkflow(context.Background(), wfr.GetID(), wfr.GetRunID(), update.Done, nil)
+		s.NoError(err)
+	}
+
+	defer signalWorkflow()
+
+	// update
```
bergundy, line 79: Worth testing both update accepted and rejected and doing some assertion on the output

workflow/workflow.go
```diff
@@ -385,5 +385,38 @@ func NewWorkflowCommands() []*cli.Command {
...
 			Action: TraceWorkflow,
 		},
+		{
+			Name:      "update",
+			Usage:     common.UpdateWorkflowDefinition,
+			UsageText: common.WorkflowUpdateUsageText,
+			Flags: append(common.FlagsForExecution,
+				&cli.StringFlag{
+					Name:     common.FlagUpdateWaitPolicy,
+					Usage:    "Wait policy determines how long the client should wait for a return value from the server:" + strings.Join(mapKeysToArray(updateWaitPolicyMap), ", "),
```
bergundy, line 395: It's not "how long", it's what to wait for.
And FTR currently only COMPLETED is implemented.

workflow/workflow.go
```diff
@@ -385,5 +385,38 @@ func NewWorkflowCommands() []*cli.Command {
...
+				},
+				&cli.StringFlag{
+					Name:     common.FlagInput,
+					Aliases:  common.FlagInputAlias,
+					Usage:    common.FlagUpdateHandlerInput,
+					Category: common.CategoryMain,
+				},
+				&cli.StringFlag{
+					Name:     common.FlagUpdateFirstExecutionRunID,
+					Usage:    "Do not run update if workflow exectuion (--run-id or if not set, latest workflow execution)  is not part of the same execution chain as this id.",
```
bergundy, line 413:
```suggestion
					Usage:    "Run update on the last execution in the chain that started with this run ID.",
```

bergundy: You'll also need an optional `--run-id` flag.

workflow/workflow_commands.go
samanbarghi: Made the changes to have all these. `waitPolicy` and related functions has just been added to sdk client and not yet available. I added the required logic and flags to extend this when these are released. 

Currently either `--name` or `--update-id` should be passed to command, former is to do an update, latter is for getting the result of an update.

workflow/workflow.go
```diff
@@ -385,5 +385,46 @@ func NewWorkflowCommands() []*cli.Command {
...
+			Usage:     common.UpdateWorkflowDefinition,
+			UsageText: common.WorkflowUpdateUsageText,
+			Flags: append(common.FlagsForExecution,
+				&cli.StringFlag{
+					Name:     common.FlagUpdateWaitPolicy,
+					Usage:    "Wait policy determines which status the client should wait to receive from the server:" + strings.Join(mapKeysToArray(updateWaitPolicyMap), ", "),
+					Category: common.CategoryMain,
+					Required: true,
+					// TODO: Remove these after more wait policies has been added
+					Hidden:     true,
```
mmcshane, line 399: Is my understanding correct that `Hidden: true` here will prevent this flag from be printed in the help output?

workflow/workflow_commands.go
```diff
@@ -54,6 +54,9 @@ var (
 		"Signal": enumspb.RESET_REAPPLY_TYPE_SIGNAL,
 		"None":   enumspb.RESET_REAPPLY_TYPE_NONE,
 	}
+	updateWaitPolicyMap = map[string]interface{}{
+		"Completed": enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED,
```
mmcshane, line 58: Is it standard practice for us to capitalize these tokens? I would have expected this to be all lowercase but if this is the common practice then it's fine. Alternatively we could put both spellings here.

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("unable to update workflow: %w", err)
+		}
+		result := map[string]interface{}{
+			"Name":     request.UpdateName,
+			"UpdateID": workflowUpdateHandle.UpdateID(),
+			"Result":   valuePtr,
+		}
+
+		fmt.Println(color.Green(c, "update workflow succeeded:"))
```
mmcshane, line 1515: Maybe only print this line if stdout is a tty. If someone is piping this output to another tool (e.g. `jq`) then a user would prefer to just have the json output and nothing pretty.

tests/workflows/update/update.go
mmcshane: addend was right here

workflow/workflow.go
samanbarghi: yes, that is correct. Since right now we only have 1 wait policy, I think we have 3 options: 
- ignore wait policy altogether and implement it later 
- implement it and show the flag with a single wait policy. 
- implement it and hide the flag for now and add it later. 

WDYT?

cretz: This sounds good to me

workflow/workflow_commands.go
cretz: In general we should have a quiet option I think and use normal logging, but I don't know how bad the CLI is at respecting stdout decisions from callers. What does running a workflow or query do? We should copy that.

workflow/workflow_commands.go
samanbarghi: That's a very good question, I followed the implementation right above this that uses camel case and feeds it directly to a switch block where it is case sensitive. I am not sure if we have standard practice though. Let me follow up on this and if there is one, leave a comment here to make it clear, and if we don't have one have a discussion around it to come up with one.

workflow/workflow_commands.go
samanbarghi: That's another very good point, I am not sure if we have standard practice for the output or not. Right now there are some commands that print out json and some that print out messages. There is a `--color` option which disables color, but I wonder if we need a `--json` that change the output format to json for all commands. Let me follow up on that.

workflow/workflow.go
mmcshane: I'd say the first option. It's not so hard to add later and it prevents accidental async invocation from well-intentioned code divers.

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
 	return nil
 }
 
+func UpdateWorkflow(c *cli.Context) error {
+	wid := c.String(common.FlagWorkflowID)
+	rid := c.String(common.FlagRunID)
+	name := c.String(common.FlagName)
+	uid := c.String(common.FlagUpdateID)
+	first_execution_run_id := c.String(common.FlagUpdateFirstExecutionRunID)
```
MichaelSnowden, line 1456: ditto on camel case

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
...
+	args, err := common.UnmarshalInputsFromCLI(c)
+	if err != nil {
+		return err
+	}
+	waitPolicy, ok := updateWaitPolicyMap[c.String(common.FlagUpdateWaitPolicy)]
+	if !ok {
+		return fmt.Errorf("must specify valid wait policy: %v", strings.Join(mapKeysToArray(updateWaitPolicyMap), ", "))
```
MichaelSnowden, line 1463: Let's follow the err113 guidelines and wrap a static error here, e.g.

```go
return fmt.Errorf(
	"%w: provided argument for %s, %q, is not one of the available options: %s",
	ErrInvalidWaitPolicy,
	common.FlagUpdateWaitPolicy,
	c.String(common.FlagUpdateWaitPolicy),
	strings.Join(mapKeysToArray(updateWaitPolicyMap), ", "),
)
```

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
...
+	}
+	if (len(name) == 0 && len(uid) == 0) || (len(name) > 0 && len(uid) > 0) {
```
MichaelSnowden, line 1465: You can simplify this to a XOR: `len(name) > 0 != len(uid) > 0`

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
...
+			Args:                args,
+			FirstExecutionRunID: first_execution_run_id,
+			// TODO: uncomment when waitPolicy is available
+			// waitPolicy:          waitPolicy,
+		}
+		return updateWorkflowHelper(c, &request, waitPolicy)
+	} else {
+		// TODO: implement this when GetWorkflowUpdateHandle is available in SDKClient
+		return fmt.Errorf("Getting update result is not yet implemented")
+	}
```
MichaelSnowden, line 1483: Use an early return instead. Also, don't capitalize error messages.


```suggestion
	if len(name) == 0 {
		// TODO: implement this when GetWorkflowUpdateHandle is available in SDKClient
		return fmt.Errorf("getting update result is not yet implemented")
	}
	request := sdkclient.UpdateWorkflowWithOptionsRequest{
	        WorkflowID:          wid,
	        RunID:               rid,
	        UpdateName:          name,
	        Args:                args,
	        FirstExecutionRunID: first_execution_run_id,
	        // TODO: uncomment when waitPolicy is available
	        // waitPolicy:          waitPolicy,
        }
        return updateWorkflowHelper(c, &request, waitPolicy)
```

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
 	return nil
 }
 
+func UpdateWorkflow(c *cli.Context) error {
+	wid := c.String(common.FlagWorkflowID)
+	rid := c.String(common.FlagRunID)
+	name := c.String(common.FlagName)
+	uid := c.String(common.FlagUpdateID)
```
MichaelSnowden, line 1455: It looks like we tell users that exactly one of FlagName or FlagUpdateID must be provided, but then we also ignore did, and we verify that FlagName must be non-empty. Why don't we just not add FlagUpdateID until it's ready?

workflow/workflow_commands.go
```diff
@@ -1445,6 +1448,76 @@ func TraceWorkflow(c *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("unable to update workflow: %w", err)
+	}
+
+	switch waitPolicy {
+	case enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED:
+		var valuePtr interface{}
+		err = workflowUpdateHandle.Get(ctx, &valuePtr)
+		if err != nil {
+			return fmt.Errorf("unable to update workflow: %w", err)
```
MichaelSnowden, line 1507: Can you use a different error message than on line 1499? It makes debugging easier

MichaelSnowden (changes requested):

tests/e2e_test.go
```diff
@@ -51,10 +59,12 @@ func (s *e2eSuite) SetupTest() {
 	})
 	s.writer = &MemWriter{}
 	s.app.Writer = s.writer
+
+	// noop exiter to prevent the app from exiting mid test
```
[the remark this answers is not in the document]
samanbarghi: urfave/cli that we use to run the app tends to exit on error, it should not happen according to [this](https://github.com/urfave/cli/pull/595), but since this behavior has been changing among different versions, I guess it's safe to assign a noop exiter, unless we want to test the app Exiting for sure. For some reason, without this the app either exits mid tests or it gets stuck forever 🤷‍♂️ I can dig into this further, but for now this is a work around that works.

tests/workflow_test.go
```diff
@@ -51,3 +55,49 @@ func (s *e2eSuite) TestWorkflowShow_ReplayableHistory() {
...
+	signalWorkflow := func() {
+		// send a Signal to stop the workflow
+		err = c.SignalWorkflow(context.Background(), wfr.GetID(), wfr.GetRunID(), update.Done, nil)
+		s.NoError(err)
+	}
+
+	defer signalWorkflow()
+
+	// successful update with wait policy Completed, should show the result
+	err = s.app.Run([]string{"", "workflow", "update", "--workflow-id", wfr.GetID(), "--run-id", wfr.GetRunID(), "--name", update.FetchAndAdd, "-i", strconv.Itoa(randomInt)})
```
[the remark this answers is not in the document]
samanbarghi: hmmm, good point, I checked and the framework does not support it. @feedmeapples do you know if this is possible?

workflow/workflow.go
samanbarghi: I see, very good point. Ok then, I can remove this and `--update-id` flag which will be ready after wait policies become available via sdk.

workflow/workflow_commands.go
samanbarghi: Yeah I think the issue is that there is no consistency in this specific file. Some subcommands just print out messages, some, like query, print out a serialized version of an object, others, like describe, prints out json. 
Another thing is for testing we need to print to `app.Writer` to be able to capture the output while testing without adding stdout capture, so need to `FPrintf` all these to `app.Writer`

common/defs-cmds.go
```diff
@@ -547,6 +547,10 @@ const WorkflowTraceUsageText = `The ` + "`" + `temporal workflow trace` + "`" +
 
 Use the options listed below to change the command's behavior.`
 
+const WorkflowUpdateUsageText = `The ` + "`" + `temporal workflow update` + "`" + ` command synchronously updates a running [Workflow Execution](/concepts/what-is-a-workflow-execution).
```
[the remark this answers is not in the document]
samanbarghi: Agree that it looks awkward and all other variables in this file are defined the same way, I leave it as is to keep it consistent with the rest.

workflow/workflow_commands.go
mmcshane: Are you talking about CLI and ID? Those would fall under the [initialism exception](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

tests/workflow_test.go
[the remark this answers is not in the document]
MichaelSnowden: I also checked and looks like it's not

tests/workflow_test.go
```diff
@@ -51,3 +55,49 @@ func (s *e2eSuite) TestWorkflowShow_ReplayableHistory() {
...
+	want := fmt.Sprintf(": %v", randomInt)
+	s.Contains(s.writer.GetContent(), want)
+
+	// successful update with wait policy Completed, passing first-execution-run-id
+	err = s.app.Run([]string{"", "workflow", "update", "--workflow-id", wfr.GetID(), "--run-id", wfr.GetRunID(), "--name", update.FetchAndAdd, "-i", "1", "--first-execution-run-id", wfr.GetRunID()})
+	s.NoError(err)
+
+	// update rejected, when name or update-id is not available
+	err = s.app.Run([]string{"", "workflow", "update", "--workflow-id", "non-existent-ID", "--run-id", wfr.GetRunID(), "-i", "1"})
+	s.Error(err)
```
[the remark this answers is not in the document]
MichaelSnowden: You can actually remove s.Error if you already have s.ErrorContains

workflow/workflow_commands.go
```diff
@@ -1445,6 +1445,56 @@ func TraceWorkflow(c *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("update workflow failed: %w", err)
+	}
+	result := map[string]interface{}{
+		"Name":     request.UpdateName,
+		"UpdateID": workflowUpdateHandle.UpdateID(),
+		"Result":   valuePtr,
+	}
+
+	fmt.Println(color.Green(c, "update workflow succeeded:"))
```
MichaelSnowden, line 1493: I think we should get rid of this entirely--I don't see any other commands using fmt.Print*, and it will make jq parsing harder

tests/workflows/update/update.go
```diff
@@ -0,0 +1,44 @@
...
+	log := workflow.GetLogger(ctx)
+	counter := val
+
+	if err := workflow.SetUpdateHandlerWithOptions(
+		ctx,
+		FetchAndAdd,
+		func(ctx workflow.Context, i int) (int, error) {
+			tmp := counter
+			counter += i
+			log.Info("counter updated", "added", i, "new-value", counter)
```
MichaelSnowden, line 24: It's a small thing, but this is a new workflow.Context, so you should probably get the logger here

tests/workflows/update/update.go
```diff
@@ -0,0 +1,44 @@
...
+			return tmp, nil
+		},
+		workflow.UpdateHandlerOptions{Validator: nonNegative},
+	); err != nil {
+		return 0, err
+	}
+
+	_ = workflow.GetSignalChannel(ctx, Done).Receive(ctx, nil)
+	return counter, ctx.Err()
```
MichaelSnowden, line 33: This isn't the right way to handle context errors. It should be this instead:

```suggestion
	if err := workflow.GetSignalChannel(ctx, Done).Receive(ctx, nil); err != nil {
	    return 0, err
	}
	return counter, nil
```

If the context is canceled, then GetSignalChannel will return that to you as an error (unless the implementation is wrong)

tests/e2e_test.go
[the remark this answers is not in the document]
MichaelSnowden: I think it's not a bug in urfave/cli but something we added in our wrapper. We exit the app unconditionally when any error occurs:

https://github.com/temporalio/cli/blob/a8bc242c033d2030026e7bd42bf34649347a9418/app/app.go#L68-L84

However, I'm actually not sure why we need to change this test at all for this PR?

workflow/workflow_commands.go
samanbarghi: Sounds good, I also feel the same way.

tests/workflows/update/update.go
samanbarghi: nice, TIL

tests/e2e_test.go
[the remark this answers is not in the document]
samanbarghi: Nice find, yes that explains it! I think the suite is still a work in progress after my chat with @feedmeapples , but I agree that localizing the side-effects from app and server is a good idea. Let me fix that then.

workflow/workflow_commands.go
MichaelSnowden: I think we're getting rid of this check, but just to correct myself, the condition I wrote above should be negated to XNOR because this condition evaluates to true iff we consider the arguments _invalid_, not valid.

workflow/workflow_commands.go
```diff
@@ -1445,6 +1445,55 @@ func TraceWorkflow(c *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("update workflow failed: %w", err)
+	}
+	result := map[string]interface{}{
+		"Name":     request.UpdateName,
+		"UpdateID": workflowUpdateHandle.UpdateID(),
+		"Result":   valuePtr,
+	}
+
+	common.PrettyPrintJSONObject(c, result)
```
MichaelSnowden, line 1493: One last thing, can you confirm that this will never cause an unexpected panic if the object is not JSON-serializable?

MichaelSnowden (approved):
LGTM pending comments

workflow/workflow_commands.go
samanbarghi: This is implemented using `JSONPBEncoder` which is a wrapper for [gogo/protobuf](https://github.com/gogo/protobuf), and it looks like it does the right thing, and should not panic.

workflow/workflow_commands.go
samanbarghi: Fro now I removed this message from the output per @MichaelSnowden suggestion and we only print out update information if this is successful. I tried to use the same approach as query by stringifying the result, but that does not look very good: 
Json
```
{
  "Name": "fetch_and_add",
  "Result": 23,
  "UpdateID": "a6dfc688-8a2b-41b4-ab3f-259ac31110eb"
}
```
vs stringify: 
```
map{Name:fetch_and_add, Result:23, UpdateID:a6dfc688-8a2b-41b4-ab3f-259ac31110eb}
```

So I keep this as JSON and will start a conversation about the format of the outputs to make them all consistent.

tests/e2e_test.go
```diff
@@ -33,28 +33,33 @@ func TestClientIntegrationSuite(t *testing.T) {
...
-func (s *e2eSuite) SetupSuite() {
+func (s *e2eSuite) SetupTest() {
 	s.app = app.BuildApp()
-	server, err := testsuite.StartDevServer(context.Background(), testsuite.DevServerOptions{})
+	server, err := testsuite.StartDevServer(context.Background(), testsuite.DevServerOptions{
+		ExtraArgs: []string{
+			// server logs are too noisy, limit server logs
+			"--log-level", "error",
+			//TODO: remove this flag when update workflow is enabled in the server by default
+			"--dynamic-config-value", "frontend.enableUpdateWorkflowExecution=true",
```
alexshtin, line 43: I don't think this should be enabled here. Let it rely on default value.

Issue #55, opened 2022-12-13
[Feature Request] delete command should support batch mode when given a query


bergundy:

feedmeapples:
@yiminc @yux0 do we plan to support batch delete?

PR #201, opened 2023-04-18
Add batch workflow delete

feedmeapples:
## What was changed

Update `temporal workflow delete` adding option to perform a batch operation using `--query` filter

resolve https://github.com/temporalio/cli/issues/55

## Why?

Supporting batch delete for workflows

## Checklist

1. Closes 

2. How was this tested:

new E2E test `TestWorkflowDelete_Batch` WIP

3. Any docs updates needed?


cretz:
Hrmm, I wonder if there's a usability concern with single workflow calls and multi-workflow calls being separated only by `--query`. Also, can you confirm that with `--yes` this prompts the user to confirm?

feedmeapples:
> 

It prompts when `--yes` is not passed. About the usability:  we've ended up with filter options such as `--workflow-id` + `--run-id`, `--query` are in a single command instead of multiple after internal discussions, the prompt and good `--query` flag should be enough? So far haven't seen/heard from people running into issues with `--query` flag 

The prompt is shown by default when `--yes` is not passed. `--yes` auto-confirms the prompts

samanbarghi (approved):

PR #204, opened 2023-04-20
Fix Dockerfile entrypoint

kurayama:
## What was changed
Fixes the Dockerfile entrypoint as the path for the temporal binary was not correct (/temporal instead of temporal).
Also moves the arguments from the ENTRYPOINT to CMD, so there's no need to override the entrypoint to run other commands.

## Why?
To be able to run the image without overriding the entrypoint.

## Checklist
2. How was this tested:

```bash
docker build -t temporal/cli:latest .
docker run docker run --rm -it temporal/cli:latest --help
docker run docker run --rm -it -p 8233:8233 temporal/cli:latest
open http://localhost:8233
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,4 +27,5 @@ EXPOSE 7233
 EXPOSE 8233
 
 # Keep the container running.
-ENTRYPOINT ["/temporal", "server", "start-dev", "-n", "default", "--ip" , "0.0.0.0"]
+ENTRYPOINT ["temporal"]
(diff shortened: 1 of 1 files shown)


feedmeapples:
there is a chance this dockerfile will be removed since instead we distribute temporal with `admin-tools` docker image. Could you elaborate do you prefer this dockerfile over the admin-tools image?

kurayama:
@feedmeapples i wasn't aware that temporal cli is now bundled with admin-tools

Issue #203, opened 2023-04-20
[Feature Request] Add a TLS configuration based on the URL format


r-mol:
### Is your feature request related to a problem? Please describe.

> Currently if TLS is enabled OR Nginx is used for the TLS termination towards the Frontend tctl does not know if it should activate the TLS unless one of the relevant flags is toggled

This feature request is taken from [issue](https://github.com/temporalio/tctl/issues/146) in the temporalio/tctl about activating TLS only by relevant flag such as `--tls_server_name`. And currently there is the same inconvenience in the temporalio/cli - https://github.com/temporalio/cli/blob/92a78d8ba0f48b37a5a3d7bd0e58c11d581f5ade/client/factory.go#L210

Currently enabling TLS requires a user to specify hostname twice — first in `--address` and then in `--tls_server_name` — which is not very convenient.

### Describe the solution you'd like

The solution proposal is exactly the same as in the issue attached above:
"Instead of activating TLS via flags, format of the temporal address can be like this `grpc://<IP or DNS>:<port>` for plain connection and `grpcs://<IP or DNS>:<port>` for the TLS. If preffix is not specified then connection can still revert to the non-TLS OR TLS based on one of the flags added. Adding this feature would help with making the Temporal protocol communication intent between the client and the frontend clean."

##### Current method:

```
temporal --address public-frontend.superhost.com:443 --tls_server_name public-frontend.superhost.com  ...
```

##### Proposed method:

```
temporal --address grpcs://public-frontend.superhost.com:443  ...
```

##### Current method:

```
temporal --address public-frontend.superhost.com:443 ...
```

##### Proposed method:

```
temporal --address grpc://public-frontend.superhost.com:443  ...
```

```
temporal --address public-frontend.superhost.com:443 ...
```

cretz:
Some thoughts (I am not a maintainer on this repo):

> Currently enabling TLS requires a user to specify hostname twice — first in `--address` and then in `--tls_server_name` — which is not very convenient.

Using a custom scheme is not the best, but I would support `--tls` with no options if you just want to enable TLS with no TLS options (e.g. you are running your own server without mTLS and your server cert is trusted). Then all `--tls-X` just imply `--tls`.

Alternatives:

* Use `grpcs://` - custom scheme is too specific to us. Also implies URL support (e.g. `grpcs://myuser:mypass@myhost`).
* Use `https://` - this can work, but implies normal HTTP API which it is not, it is a gRPC endpoint. Also, implies URL support like the above.
[...]

PR #210, opened 2023-04-22
add support for --tls flag (#203)

r-mol:
## What was changed
1. client/factory.go -> Added parsing `--tls` flag with checking on error. And added the logic that when the flag is activated, the host is taken from the address/localhost
2. common/defs-flags.go -> Added related definition for ` --tls` flag
3. common/flags.go -> Added const of `--tls` flag name. And added `--tls` flag to SharedFlag slice

## Why?
I am connecting to my server over a TLS connection and I had to duplicate my host from the `--address` flag to the `--tls_server_name` flag. This cluttered up the command, as well as misleading.

## Checklist

1. Closes [#203](https://github.com/temporalio/cli/issues/203)

4. How was this tested:
- By internal tests, but with adding this flag
- By connecting to local server with this flag

5. Any docs updates needed? - NO

## Related to same issue
https://github.com/temporalio/tctl/pull/360


client/factory.go
```diff
@@ -216,6 +216,11 @@ func (b *clientFactory) createTLSConfig(c *cli.Context) (*tls.Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("unable to read TLS disable host verification flag: %w", err)
 	}
+	enableTLSS := c.String(common.FlagTLS)
+	enableTLS, err := strconv.ParseBool(enableTLSS)
```
feedmeapples, line 220: what's the reason to do custom parsing? (instead of `c.Bool(..)`)

common/defs-flags.go
```diff
@@ -1,10 +1,12 @@
 package common
+
 const (
 	// Shared flag definitions
 	FlagEnvDefinition = "Environment to read environmental variables from."
 	FlagAddrDefinition = "The host and port (formatted as host:port) for the Temporal Frontend Service."
 	FlagNSAliasDefinition = "Identifies a Namespace in the Temporal Workflow."
 	FlagMetadataDefinition = "Contains gRPC metadata to send with requests (formatted as key=value). Values must be in a valid JSON format."
+	FlagTLSDefinition = "Enable TLS"
```
feedmeapples, line 9: we should elaborate here for users, maybe
```suggestion
	FlagTLSDefinition = "Enable TLS encryption without additional options such as mTLS or client certificates"
```

feedmeapples (changes requested):

client/factory.go
r-mol: I'm doing custom parsing to keep consistency with your code above: https://github.com/temporalio/cli/blob/6a16f1c6ef215dfe88b4bb8598bd7c21bc4976c9/client/factory.go#L214

feedmeapples: ok so these values are accepted with strconv.ParseBool, true: `"1", "t", "T", "true", "TRUE", "True"`, false: `"0", "f", "F", "false", "FALSE", "False"`

I'd go with the c.Bool for consistency with the rest of the CLI

feedmeapples (approved):

client/factory.go
feedmeapples: updated

common/defs-flags.go
```diff
@@ -6,6 +6,7 @@ const (
 	FlagAddrDefinition           = "The host and port (formatted as host:port) for the Temporal Frontend Service."
 	FlagNSAliasDefinition        = "Identifies a Namespace in the Temporal Workflow."
 	FlagMetadataDefinition       = "Contains gRPC metadata to send with requests (formatted as key=value)."
+  FlagTLSDefinition            = "Enable TLS encryption without additional options such as mTLS or client certificates"
```
feedmeapples, line 9:
```suggestion
	FlagTLSDefinition            = "Enable TLS encryption without additional options such as mTLS or client certificates"
```

feedmeapples:
Tests fail due to recent regression unrelated to the current PR, specifically verifying RC code in `TestServerInterruptRC` and E2E tests running indefinitely. Approving and fixing the tests in different PR

r-mol:
And could you also check the PR with the same issue https://github.com/temporalio/tctl/pull/360.

PR #221, opened 2023-04-27
Fix e2e test suite

samanbarghi:
## What was changed

Fix e2e suite

## Why?

Tests were failing

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -29,11 +29,7 @@ type (
 	}
 )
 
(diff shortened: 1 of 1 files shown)


feedmeapples (approved):

feedmeapples:
Running tests in parallel with clean servers is still a goal. Approving until getting it right

Issue #209, opened 2023-04-21
[Bug] namespace update - update ns retention not working


tsurdilo:
temporal -v
temporal version 0.8.0 (server 1.20.1) (ui 2.13.3)
running server version 1.20.1 on local docker (tried with minikube too)

command 
temporal operator namespace update default --retention 3

does not update the namespace retention, no error result of command is "Namespace default successfully updated."

works ok with tctl for example: 
tctl --ns default n u -rd 3

sahilvv:
This bug is impacting our transition from tctl to temporal cli. Related command is also not working:
```
temporal operator namespace update temporal-system --email xxx --retention 90
```

bergundy:
User reported on slack that this command was not setting email and retention either:
```
temporal operator namespace create test-cli --retention 10 --email abc@gmail.com
```

feedmeapples:
working on this

feedmeapples:
the namespace argument should come last. I think we should clarify this in docs and --help

```
temporal operator namespace update --email xxx --retention 90 temporal-system
```

feedmeapples:
checking if we can make the argument positioned differently. Otherwise keeping the item open to detect misplaced namespace arg

feedmeapples:
If namespace arg comes before the flags, the flags are ignored by urfave/cli and non of the fields are actually updated.

We can change the output of the command so it lists the fields that has actually changed:

before
```
temporal operator namespace update --email xxx --retention 90 default
Namespace default successfully updated.
```

after
```
temporal operator namespace update default --email xxx --retention 90
Namespace update changes implemented:

Field        Before    After    
Retention      1        90
Email                   xxx
```

feedmeapples:
we can also detect that there are flags passed after the arg and throw an error saying to pass the arg the last

PR #222, opened 2023-04-28
Clarify namespace update result with --verbose flag

feedmeapples:
## What was changed

- Started printing Namepeace Update request as JSON
- Rewritten namespace update tests as E2E tests instead of unit
- Increased context deadline for workflow update tests

## Why?

- Help users when running into this issue, resolve https://github.com/temporalio/cli/issues/209
- Workflow update tests would occasionally fail due to context deadline errors

## Checklist

1. Closes 

2. How was this tested:

Added E2E tests

3. Any docs updates needed?


bergundy (commented):
Let's just print out what changed instead of diffing.
I don't see a good enough reason to do the extra describe call.

feedmeapples:
> Let's just print out what changed instead of diffing. I don't see a good enough reason to do the extra describe call.

I started with this initially, seemed harder to maintain as ended up with `if` case for each property and filling in a map of change

bergundy:
I mean just print out what we sent to the server, not the actual diff.

app/namespace_test.go
```diff
@@ -70,53 +70,6 @@ var describeNamespaceResponseServer = &workflowservice.DescribeNamespaceResponse
...
-	resp := describeNamespaceResponseServer
-	s.frontendClient.EXPECT().DescribeNamespace(gomock.Any(), gomock.Any()).Return(resp, nil).Times(2)
-	s.frontendClient.EXPECT().UpdateNamespace(gomock.Any(), gomock.Any()).Return(nil, nil).Times(2)
-	err := s.app.Run([]string{"", "operator", "namespace", "update", cliTestNamespace})
-	s.Nil(err)
-	err = s.app.Run([]string{"", "operator", "namespace", "update", "--description", "another desc", "--email", "another@uber.com", "--retention", "1", cliTestNamespace})
-	s.Nil(err)
-}
-
-func (s *cliAppSuite) TestNamespaceUpdate_Data() {
```
feedmeapples, line 83: Re-created these as E2E tests. Removed unit tests as these do not really test much, even here the usage was wrong and it read `"true"` as the name of the namespace

samanbarghi (approved):

PR #227, opened 2023-05-04
Replace custom bool conversion with urfave Bool

feedmeapples:
## What was changed

Replaced `strconv.ParseBool` for TLS flags with urfave's `c.Bool`

## Why?

Consistency with other Bool flags

related https://github.com/temporalio/cli/pull/210/files#r1180775765

## Checklist

1. Closes 

2. How was this tested:

Manually verified the same conversion. Backwards compatible since `c.Bool` does strconv.ParseBool internally. Also  previously something like `temporal workflow list --tls-disable-host-verification 0` would result in a string value to be equal to `true` because this flag is defined as `cli.BoolFlag`.

3. Any docs updates needed?

### DIFF
--- a/client/factory.go
+++ b/client/factory.go
@@ -5,12 +5,10 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
-	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"os"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bergundy:
Just to clarify, passing `--tls` with no value should enable TLS. Is that what's happening here?

feedmeapples:
> Just to clarify, passing `--tls` with no value should enable TLS. Is that what's happening here?

yes

bergundy (approved):

PR #229, opened 2023-05-08
Disable CSRF for dev server

feedmeapples:
## What was changed

Disabled CSRF for dev server

## Why?

since this is a dev server CSRF protection unnecessarily complicates hosting dev server, requiring setting up reverse proxy / HTTPS if hosted with non-localhost 

## Checklist

1. Closes 
1. How was this tested:

1. Any docs updates needed?


josh-berry:
No, we should be secure by default unless there's a specific reason not to be. People should not be hosting the dev server anyway.

PR #231, opened 2023-05-08
Fix parsing dynamic config numbers

rodrigozhou:
## What was changed

Parse dynamic config numbers correctly: use `json.Number`, and then tries to parse to int, and then to float.

## Why?

Without using `json.Decoder`, all number are parsed to `float64` which is not read correctly by dynamic config expecting `int` type value.

## Checklist

1. Closes 

2. How was this tested:

Started Temporal Server via CLI tool:
```
$ ./temporal server start-dev --dynamic-config-value frontend.searchAttributesSizeOfValueLimit=20480 --dynamic-config-value frontend.searchAttributesTotalSizeLimit=40480
```
Run search attribute sample workflow with very large values.

3. Any docs updates needed?

### DIFF
--- a/server/commands.go
+++ b/server/commands.go
@@ -27,6 +27,7 @@
 package server
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


cretz (approved):

feedmeapples:
approving given the failing unit test is updated

feedmeapples (approved):

PR #233, opened 2023-05-09
Add frontend to trace command

Sikian:
## What was changed
Added the frontend to the trace command. This includes the following:

1. `TermWriter`: a way to manage printing/clearing the terminal. This is not Windows-compliant.
2. `ExecutionTemplate`: takes care of rendering the display for ExecutionStatus.
3. Tied everything together so the command words correctly.

### Example

This is an example outlining the different workflow and activity status:
Screenshot 2023-05-10 at 17 40 26

You can see the following (in order)

- Root workflow fails due to a non-retryable error from an activity
- First activity run completes successfully
- First child workflow completes successfully (folded)
- Second child workflow times out 
- 5s timer (fired successfully)
- Failing activity (that fails the root workflow)

## Why?

These changes complete the upstream of the trace functionality that has been used in DataDog for the last year.

## Checklist

1. Closes `trace` upstream.

4. Used extensively at DataDog + unit tests and running it against well-known examples (see Example)

5. Any docs updates needed?
Probably those related to the temporal cli regarding how to watch workflow executions.

### DIFF
--- a/trace/execution_icons.go
+++ b/trace/execution_icons.go
@@ -1,0 +1,102 @@
+package trace
+
+import (
+	"github.com/fatih/color"
+	"go.temporal.io/api/enums/v1"
+)
+
+var (
+	StatusRunning              string = color.BlueString("▷")
+	StatusCompleted            string = color.GreenString("✓")
+	StatusTerminated           string = color.RedString("x")
+	StatusCanceled             string = color.YellowString("x")
+	StatusFailed               string = color.RedString("!")
+	StatusContinueAsNew        string = color.GreenString("»")
+	StatusTimedOut             string = color.RedString("⏱")
+	StatusUnspecifiedScheduled string = "•"
+	StatusCancelRequested      string = color.YellowString("▷")
+	StatusTimerWaiting         string = color.BlueString("⧖")
+	StatusTimerFired           string = color.GreenString("⧖")
+	StatusTimerCanceled        string = color.YellowString("⧖")
+)
+
+var workflowIcons = map[enums.WorkflowExecutionStatus]string{
+	enums.WORKFLOW_EXECUTION_STATUS_UNSPECIFIED:      StatusUnspecifiedScheduled,
+	enums.WORKFLOW_EXECUTION_STATUS_RUNNING:          StatusRunning,
+	enums.WORKFLOW_EXECUTION_STATUS_COMPLETED:        StatusCompleted,
+	enums.WORKFLOW_EXECUTION_STATUS_TERMINATED:       StatusTerminated,
+	enums.WORKFLOW_EXECUTION_STATUS_CANCELED:         StatusCanceled,
+	enums.WORKFLOW_EXECUTION_STATUS_FAILED:           StatusFailed,
+	enums.WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW: StatusContinueAsNew,
+	enums.WORKFLOW_EXECUTION_STATUS_TIMED_OUT:        StatusTimedOut,
+}
+var activityIcons = map[ActivityExecutionStatus]string{
+	ACTIVITY_EXECUTION_STATUS_UNSPECIFIED:      StatusUnspecifiedScheduled,
+	ACTIVITY_EXECUTION_STATUS_SCHEDULED:        StatusUnspecifiedScheduled,
+	ACTIVITY_EXECUTION_STATUS_RUNNING:          StatusRunning,
+	ACTIVITY_EXECUTION_STATUS_COMPLETED:        StatusCompleted,
+	ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED: StatusCancelRequested,
+	ACTIVITY_EXECUTION_STATUS_CANCELED:         StatusCanceled,
+	ACTIVITY_EXECUTION_STATUS_FAILED:           StatusFailed,
+	ACTIVITY_EXECUTION_STATUS_TIMED_OUT:        StatusTimedOut,
+}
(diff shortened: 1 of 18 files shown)


trace/execution_state.go
```diff
@@ -246,6 +246,9 @@ func (state *WorkflowExecutionState) Update(event *history.HistoryEvent) {
 		state.StartTime = event.EventTime
 		state.Attempt = attrs.GetAttempt()
 		state.Type = attrs.GetWorkflowType()
+		if state.Execution.GetRunId() == "" {
+			state.Execution.RunId = attrs.GetOriginalExecutionRunId()
```
Sikian, line 250: Needed to obtain RunId when only workflow id has been provided.

trace/execution_icons.go
```diff
@@ -0,0 +1,102 @@
+package trace
+
+import (
+	"github.com/fatih/color"
+	"go.temporal.io/api/enums/v1"
+)
+
+var (
```
bergundy, line 8: All of this is great, I would also add spinners on anything that's currently in progress.
The colors are awesome but not super intuitive to me and don't account for color blindness.

trace/execution_icons.go
```diff
@@ -0,0 +1,102 @@
...
+	return "?"
+}
+
+// StatusIcon has names for each status (useful for help messages).
+type StatusIcon struct {
+	Name string
+	Icon string
+}
+
+var StatusIconsLegend = []StatusIcon{
```
bergundy, line 74: I couldn't figure out how I could see the legend.

trace/execution_state.go
bergundy, line 250: Hmm.. that seems wrong at first glance, when will you not have the run ID?

bergundy: To accurately get the run ID from history events, you'll need to scan all events in reverse order and take the first one that matches [this](https://github.com/temporalio/api/blob/e7b6fa681f663723b9e419f5fa350e68a3fdfc5c/temporal/api/history/v1/message.proto#L230) or fallback to original run ID.

bergundy (commented):
Awesome!
This is great and we merge as is AFAIC.
I added a few comments that we can discuss and you can choose to address now or in a followup PR.

If I run a workflow that's waiting for a signal, I the output shown below:
```
Progress:
Processing HistoryEvents
```

Instead we should show the workflow type + IDs as we do for a workflow with other events.

trace/execution_icons.go
Sikian: I haven't implemented the legend since I'm still trying to understand how the descriptions are built in the cli.

trace/execution_state.go
Sikian: It's the best we can get at this stage of the execution history. This will do for now.

trace/templates/workflow.tmpl
```diff
@@ -0,0 +1,27 @@
+{{ indent .Depth }} {{ faint "╪" }} {{ statusIcon .State }} {{ .State.GetName }} {{ template "extras" . }}
+{{ indent .Depth 1 }} {{ with .State.Execution }}{{ faint "wfId:" }} {{ .WorkflowId |  blue }} {{- faint ", runId:" }} {{ .RunId | blue }}{{ end }}
```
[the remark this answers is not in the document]
Sikian: Space is tight in the terminal and workflow ids are already long enough. I would advise against using the whole word and "id" isn't descriptive by itself.

Feel free to change it if you think otherwise, though.

trace/execution_icons.go
Sikian: I'm not sure to understand what "in progress" means in this context.

I completely understand the concern with colour blindness, but there's so much I've been able to address for now. I'm happy to learn about how to improve the tool in this regard.

bergundy: In progress would be any activity/child workflow/timer that has not yet been concluded (e.g. completed/failed/cancelled...)
I'll open an issue to reconsider the design.

trace/execution_state.go
bergundy: I'll open an issue to track and get the correct run Id.

trace/templates/workflow.tmpl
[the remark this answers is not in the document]
bergundy: We use `Id` in some of the SDKs, i think it's fine here too.

trace/execution_state.go
Sikian: Much appreciated! Would you let me know when this happens to learn about it? Thanks

trace/execution_state_test.go
```diff
@@ -774,3 +774,85 @@ func TestExecutionState_TimerExecutionStateImplementation(t *testing.T) {
...
 	}
 }
+
+func TestWorkflowExecutionState_GetNumberOfEvents(t *testing.T) {
+	tests := map[string]struct {
+		state       *WorkflowExecutionState
+		wantCurrent int64
+		wantTotal   int64
+	}{
+		"no childs": {
```
bergundy, line 784: nit: use "childen"

bergundy (approved):
I'm approving, the few comments that I had left can be address in later PRs by you or someone else on the team.

bergundy (approved):
I'm approving, the few comments that I had left can be address in later PRs by you or someone else on the team.

bergundy:
@feedmeapples can you please take a look and see if we can merge albeit failing tests?

feedmeapples:
> @feedmeapples can you please take a look and see if we can merge albeit failing tests?

i've fixed few of the lint issues, though there is still a number. The tests seem fine

PR #240, opened 2023-05-15
Support custom version in installation via .sh script

feedmeapples:
## What was changed

Added support for CLI version in insall.sh

## Why?

Used by https://github.com/temporalio/docker-builds/pull/112

## Checklist

1. Closes 

2. How was this tested:

Manually so far. There is an install.test.sh file for testing, i'd prefer to have testing with a proper language (not done).

```
sh -c "$(cat ./install.sh)" -- --dir ./temp
(cd ./temp/bin && ./temporal -h)
...
VERSION:
   0.9.0 (server 1.20.2) (ui 2.15.0)
```

```
sh -c "$(cat ./install.sh)" -- --dir ./temp --version v0.3.0
(cd ./temp/bin && ./temporal -h)
...
VERSION:
   0.3.0 (server 1.19.1) (ui 2.9.1)
```

```
sh -c "$(cat ./install.sh)" -- --dir ./temp --version 0.3.0
(cd ./temp/bin && ./temporal -h)
...
VERSION:
   0.3.0 (server 1.19.1) (ui 2.9.1)
```

3. Any docs updates needed?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -337,7 +337,7 @@ get_install_dir() {
     local _dir
     _dir="$(get_default_install_dir)"
 
-    while [[ $# -gt 0 ]]; do
+    while [ $# -gt 0 ]; do
         case "$1" in
         --dir)
             _dir="$2"
@@ -359,7 +359,7 @@ get_version() {
     local _version
     _version="latest"
 
-    while [[ $# -gt 0 ]]; do
+    while [ $# -gt 0 ]; do
         case "$1" in
         --version)
             _version="$2"
@@ -371,9 +371,10 @@ get_version() {
         esac
     done
 
-    if [[ $_version == [0-9]* ]]; then
+    if [ "$(echo "$_version" | grep -E '^[0-9]+')" ]; then
         _version="v$_version"
     fi
+
     printf %s "$_version"
 }
 
--- a/install.sh
+++ b/install.sh
@@ -337,16 +337,16 @@ get_install_dir() {
     local _dir
     _dir="$(get_default_install_dir)"
 
-    while [ $# -gt 0 ]; do
+    while [[ $# -gt 0 ]]; do
         case "$1" in
         --dir)
             _dir="$2"
+            shift 2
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -347,6 +352,32 @@ get_install_dir() {
...
+    local _version
+    _version="latest"
+
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+        --version)
+            _version="$2"
+            shift 2
+            ;;
+        *)
```
dnr, line 368: This and the loop in `get_install_dir` should be written the same way. I like this way better but either way, be consistent

install.sh
```diff
@@ -347,6 +352,32 @@ get_install_dir() {
...
+            shift
+            ;;
+        esac
+    done
+
+    if [[ $_version == v* ]] || [[ $_version == "latest" ]]; then
+        printf %s "$_version"
+    else
+        printf 'v%s' "$_version"
+    fi
```
dnr, line 378: this reads better to me if you do it like
```suggestion
    if [[ $_version == [0-9]* ]]; then
        _version="v$_version"
    fi
    printf %s "$_version"
```

dnr (approved):
couple nitpicks but with those it's good

install.sh
feedmeapples: updated

install.sh
feedmeapples: used your logic + few changes to address shellcheck

install.sh
```diff
@@ -371,9 +371,10 @@ get_version() {
         esac
     done
 
-    if [[ $_version == [0-9]* ]]; then
+    if echo "$_version" | grep -qE '^[0-9]+'; then
```
dnr, line 374: If `if` doesn't work, pretty sure `case` will, but this is fine too.
```suggestion
    if echo "$_version" | grep -q '^[0-9]'; then
```

dnr (approved):

PR #246, opened 2023-05-23
Get reverse history and set scheduled event task ID properly

samanbarghi:
## What was changed

Get history events in reverse and fix the logic to set to first task scheduled
event instead of the last one. 

## Why?

Currently, we iterate through the events starting from the beginning and try to
find a workflowTaskCompleted event and keep going until we reach the end of the
list. If we cannot find a completed event, we set the event ID to the id of the
workflowScheduledTask event. However, this means if we have
'completed->completed->failed' or 'failed->retry(failed)->retry(failed)` events, 
we always set the event ID to the
scheduled event corresponding to the last failed workflow event. Instead, by
iterating through the events in reverse, we can fix this behavior and set the
last workflowCompletedEvent or the first workflowScheduled event (if all
attempts failed). 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


yiminc (approved):

feedmeapples (approved):

feedmeapples:
i'd like e2e tests everywhere (if you have capacity to add one for this)

bergundy:
Just verifying that this doesn’t change anything in the ux

samanbarghi:
> would like to have an e2e for this if you have capacity

adding e2e tests will be a bit involved as we will need to generate many scenarios, the same logic exists in the server and there are unit-tests for it, see this [PR](https://github.com/temporalio/temporal/pull/4157/files#diff-d5912f7976201bc8d1a873ca27f36b8dde1b4eade7afe3833ae7ac9be6db7df1R353). 

I am not sure if e2e testing this specific function have any benefits now, will leave this to be included in e2e tests for `reset` subcommand.

samanbarghi:
> Just verifying that this doesn’t change anything in the ux

It should not, there is an `enum` used in ux to select among different methods. 

Btw, we might want to consider using `BatchReset` API added to server recently in this [PR](https://github.com/temporalio/temporal/pull/4157).

feedmeapples:
> Btw, we might want to consider using `BatchReset` API added to server recently in this [PR](https://github.com/temporalio/temporal/pull/4157).

That's great to hear!

Issue #173, opened 2023-03-26
[Bug] Fix CI


lorensr:
### What are you really trying to do?

Publicize this repo and help people feel confident in using it.

### Describe the bug

Red badge: 

![image](https://user-images.githubusercontent.com/251288/227759413-eecf39f4-95f7-435a-be46-75bca0edef6c.png)

Here's the run on latest `main` commit:

https://github.com/temporalio/cli/actions/runs/4517297860/jobs/7956386140

I'm seeing 3 different things:

`test (windows-latest)`

<img width="562" alt="image" src="https://user-images.githubusercontent.com/251288/227759443-1e1573d3-8f67-46d2-93a5-32533d090d53.png">

<img width="482" alt="image" src="https://user-images.githubusercontent.com/251288/227759449-80afe682-2958-41f9-9077-2a79bb2d974a.png">

<img width="624" alt="image" src="https://user-images.githubusercontent.com/251288/227759452-d3fcceac-5426-4482-841d-5165d3ff92b3.png">

@feedmeapples are you up for prioritizing this? Or do you have a sense of who else would be best to work on it?

PR #257, opened 2023-06-06
Fix Windows build & CI

feedmeapples:
## What was changed

Fixed Windows build, Windows CI and lint issues, `trace` command on Windows, build e2e server locally instead of using the latest release.. etc CI related

## Why?

green checks

resolve https://github.com/temporalio/cli/issues/173
resolve https://github.com/temporalio/cli/issues/268

## Checklist

1. Closes 

2. How was this tested:

CI

3. Any docs updates needed?


trace/term_size_windows.go
```diff
@@ -0,0 +1,46 @@
...
+
+var (
+	kernel32                       = syscall.NewLazyDLL("kernel32.dll")
+	procGetConsoleScreenBufferInfo = kernel32.NewProc("GetConsoleScreenBufferInfo")
+)
+
+func getTerminalSize() (int, int) {
+	var csbi CONSOLE_SCREEN_BUFFER_INFO
+
+	procGetConsoleScreenBufferInfo.Call(uintptr(syscall.Stdout), uintptr(unsafe.Pointer(&csbi)))
```
bergundy, line 43: Does this return anything that can be used to determine if it succeeds?

bergundy (approved):

trace/term_size_windows.go
cretz, line 1: Ug, I was a bit disappointed to see the trace implementation use manual ANSI codes and syscalls everywhere. I wish we had encouraged https://github.com/gdamore/tcell or similar.

Is this change just to fix compilation or can you confirm the trace command works well on Windows? If you can confirm, no prob. If you're just fixing compilation but it's still broken/wrong at runtime, I think you should just make this entire command only available on nix builds and put a disclaimer somewhere on why.

feedmeapples: i'm testing this on Windows. Yeah ideally we wouldn't do platform specific coding

feedmeapples: i'll see before merging this PR if we can remove platform specific code

trace/term_size_windows.go
feedmeapples: it does, added a check

trace/term_size_windows.go
cretz: I don't mind the small platform-specific parts (e.g. sigint sending), I just want to make sure this trace feature works on all platforms we're making it available for

cretz (changes requested):

feedmeapples:
still dealing with this failure https://github.com/temporalio/cli/actions/runs/5262862061/jobs/9512447848#step:5:154

`go test` succeeds when running locally on Windows, however not in Github Actions. Haven't found yet what causes this

feedmeapples:
sent one more blocker PR into Go SDK https://github.com/temporalio/sdk-go/pull/1140

tests/e2e_test.go
```diff
@@ -37,6 +40,10 @@ func (s *e2eSuite) SetupSuite() {
 }
 
 func (s *e2eSuite) TearDownSuite() {
+	// Ensure all servers are stopped in case not explicitly stopped from tests
```
feedmeapples, line 43:
```suggestion
	// ensure all servers are stopped in case not explicitly stopped from tests
```

samanbarghi (approved):

go.mod
```diff
@@ -7,21 +7,21 @@ require (
...
 	github.com/temporalio/tctl-kit v0.0.0-20230328153839-577f95d16fa0
 	github.com/temporalio/ui-server/v2 v2.15.0
 	github.com/urfave/cli/v2 v2.23.6
-	go.temporal.io/api v1.18.2-0.20230324225508-f2c7ab685b44
-	go.temporal.io/sdk v1.21.2
+	go.temporal.io/api v1.23.0
+	go.temporal.io/sdk v1.23.1
 	go.temporal.io/sdk/contrib/tools/workflowcheck v0.0.0-20230328164709-88a40de39c33
-	go.temporal.io/server v1.20.3
+	go.temporal.io/server v1.20.1-0.20230616203625-3392a7ab579a
```
cretz, line 22: Let's wait for server tag

cretz (changes requested):

cretz (approved):
Approving with the guarantee that we are not going to tag CLI with an untagged server dependency

Issue #262, opened 2023-06-11
[Bug] Prevent deleting default env


lorensr:
### Describe the bug

Because once it's deleted, I can't set it:

```
t env delete default
Deleted env: default

t env set default.address localhost:7233
Error: env not found: default
('export TEMPORAL_CLI_SHOW_STACKS=1' to see stack traces)
```

bergundy:
Hmm.. it should be okay to empty the contents of the default env.
In any case `temporal env set default.address localhost:7233` should not fail if the default env doesn't exist.

lorensr:
I get the same

```
Error: env not found: default
```

when running most commands

bergundy:
I got that too.
I think we can make an exception and treat missing default as empty default.

lorensr:
In the meantime, how do I fix? Tried reinstalling, still get the error on
all commands 😄

[image: image.png]

feedmeapples:
if that's a quick fix i'll send a PR. Meanwhile feel free to delete the config `~/.config/temporalio/temporal.yaml`

lorensr:
Deleting worked, thanks!

PR #264, opened 2023-06-16
Fix commands throwing when default env is removed

feedmeapples:
## What was changed

Resolve https://github.com/temporalio/cli/issues/262

## Why?

commands such as `temporal workflow list` would throw "env default not found" if default env is removed. In this case `workflow list` should fall back to flag defaults 

## Checklist

1. Closes 

2. How was this tested:

Added test

3. Any docs updates needed?


app/env_test.go
```diff
@@ -54,6 +55,22 @@ func (s *cliAppSuite) TestDeleteEnv() {
 	s.NotContains(config, "address: 1.2.3.4:5678")
 }
 
+func (s *cliAppSuite) TestDeleteEnv_Default() {
+	defer setupConfig(s.app)()
```
bergundy, line 59: What’s this defer doing?

bergundy (approved):

app/env_test.go
feedmeapples: cleans up env config for the next test

samanbarghi (approved):

PR #280, opened 2023-07-06
Fix bash completion bugs

dnr:
## What was changed
- Bump urfave/cli dependency to pick up https://github.com/urfave/cli/pull/1618, https://github.com/urfave/cli/pull/1676, and more
- Remove `-o nospace` in suggested bash completion function
- Upgrade bash and zsh completion scripts to the ones in v2 branch

## Why?
Makes completing deep subcommands work correctly and easily

## Checklist

1. Closes 

2. How was this tested:
manually

4. Any docs updates needed?


mindaugasrukas (approved):

dnr:
Actually I realized we're using v1 of the bash completion script: https://github.com/urfave/cli/blob/master/autocomplete/bash_autocomplete
v2/v3 looks slightly different and _does_ depend on the `bash-completion` library: https://github.com/urfave/cli/blob/v2-maint/autocomplete/bash_autocomplete

Apparently this fixes bugs with args with colons and quotes: https://github.com/urfave/cli/pull/1674

Is there an intention to move to v2/v3 of the script? Fixing bugs is nice but it works on commands, which is the major win, and the dependency is annoying, since I want to enable this in the admin-tools image (there is an alpine package though so it could be done).

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/stretchr/testify v1.8.4
 	github.com/temporalio/tctl-kit v0.0.0-20230328153839-577f95d16fa0
 	github.com/temporalio/ui-server/v2 v2.16.2
-	github.com/urfave/cli/v2 v2.23.6
+	github.com/urfave/cli/v2 v2.25.7
```
feedmeapples, line 18: i wonder if this is no longer the case for urfave/cli > v2.23.6 https://github.com/temporalio/cli/blob/d37f65cec42f708370ec06d3cf92088ce71eb827/Makefile#L20C147-L20C147

Since we now have e2e setup i'll likely rewrite this test with an actual workflow input https://github.com/temporalio/cli/blob/d37f65cec42f708370ec06d3cf92088ce71eb827/app/app_test.go#L127

feedmeapples (approved):

go.mod
dnr: Hmm.. I just tested it and it seems to work. Should I delete the pinned dependency?

dnr (commented):
Any thoughts on the dependency and script version? I'm tempted to leave it on the old script and live with the bugs in exchange for no dependencies

feedmeapples:
> Any thoughts on the dependency and script version? I'm tempted to leave it on the old script and live with the bugs in exchange for no dependencies

We may move to `cobra` and its auto-completion does depend on `bash-completion`. `git` relies on this package too. I vote it's fine to add dependency for improved UX in bash

dnr:
> We may move to `cobra` and its auto-completion does depend on `bash-completion`. `git` relies on this package too. I vote it's fine to add dependency for improved UX in bash

Ok, I'll update the script and revert the README changes

PR #284, opened 2023-07-12
Clarify namespace flags

feedmeapples:
## What was changed

Clarified and reused namespace flag definitions

## Why?

--help output improvements

## Checklist

1. Closes 

2. How was this tested:

CI

3. Any docs updates needed?

### DIFF
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -136,6 +136,21 @@ const (
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
 
+	// Namespace update flags
+	FlagActiveClusterDefinition           = "Active cluster name."
+	FlagClusterDefinition                 = "Cluster name."
+	FlagDescriptionDefinition             = "Namespace description."
+	FlagHistoryArchivalStateDefinition    = "History archival state, valid values are \"disabled\" and \"enabled\""
+	FlagHistoryArchivalURIDefinition      = "Optionally specify history archival URI (cannot be changed after first time archival is enabled)"
+	FlagIsGlobalNamespaceDefinition       = "Whether the namespace is a global namespace."
+	FlagNamespaceDataDefinition           = "Namespace data in key=value format. Use valid JSON formats for value."
+	FlagNamespaceVerboseDefinition        = "Print applied namespace changes"
+	FlagOwnerDefinition                   = "Owner email."
+	FlagPromoteNamespaceDefinition        = "Promote local namespace to global namespace"
+	FlagRetentionDefinition               = "Length of time (in days) a closed Workflow is preserved before deletion."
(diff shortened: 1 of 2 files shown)


common/defs-flags.go
```diff
@@ -136,6 +136,21 @@ const (
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
 
+	// Namespace update flags
+	FlagActiveClusterDefinition           = "Active cluster name."
+	FlagClusterDefinition                 = "Cluster name."
+	FlagDescriptionDefinition             = "Namespace description."
+	FlagHistoryArchivalStateDefinition    = "History archival state, valid values are \"disabled\" and \"enabled\""
```
lorensr, line 143: should this and `FlagVisibilityArchivalStateDefinition` be boolean flags?

common/defs-flags.go
```diff
@@ -136,6 +136,21 @@ const (
...
+	FlagHistoryArchivalURIDefinition      = "Optionally specify history archival URI (cannot be changed after first time archival is enabled)"
+	FlagIsGlobalNamespaceDefinition       = "Whether the namespace is a global namespace."
+	FlagNamespaceDataDefinition           = "Namespace data in key=value format. Use valid JSON formats for value."
```
lorensr, line 146:
```suggestion
	FlagNamespaceDataDefinition           = "Namespace data in key=value format. Use JSON for values."
```

common/defs-flags.go
```diff
@@ -136,6 +136,21 @@ const (
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
 
+	// Namespace update flags
+	FlagActiveClusterDefinition           = "Active cluster name."
```
lorensr, line 140: I prefer no periods in descriptions unless it's multiple sentences. Is there prior discussion on this? Looks like the built-in options have no period:

image

lorensr (approved):

common/defs-flags.go
feedmeapples: in this case we may want to rename these as well similarly to cluster `--enable-connection` and `--tls-disable-host-verification` 

-- `enable-history-archival`
-- `enable-visibility-archival`

i'll start a thread separately from this PR

feedmeapples: can keep `visibility-archival-state` as well for backwards compatibility and make them invisible in --help

common/defs-flags.go
feedmeapples: i believe we added the dots because this is used for docs generation. We should be able to still remove the dots (or add when generating the docs) programmatically

common/defs-flags.go
feedmeapples: upd. we should keep this as explicitly "enabled", "disabled" values as there is a 3rd option unspecified that afaik just takes the default according to settings

PR #304, opened 2023-08-03
Update help text

lorensr:
- Don't use periods unless more than one phrase (only other period command I found was `react-native`, all the rest didn't have periods)
- On top level, list the available subcommands or most common ones, rather than the abstract 'operations'
- Give most common examples in USAGE section

### Current
image

### Proposed
image

### Current

image

### Proposed

image


server/commands.go
```diff
@@ -126,13 +126,13 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
 				},
 				&cli.StringSliceFlag{
 					Name:    common.FlagPragma,
-					Usage:   fmt.Sprintf("Specify sqlite pragma statements in pragma=value format. Pragma options: %q.", sconfig.GetAllowedPragmas()),
+					Usage:   fmt.Sprintf("Specify SQLite pragma statements in pragma=value format. Pragma options: %q.", sconfig.GetAllowedPragmas()),
```
lorensr, line 129:
```suggestion
					Usage:   fmt.Sprintf("Specify SQLite [pragma statements](https://www.sqlite.org/pragma.html] in pragma=value format. Pragma options: %q.", sconfig.GetAllowedPragmas()),
```

would be nice if we could do this, and have it show up in docs but not CLI output. possible? easy / worth it?

bergundy: @feedmeapples

bergundy (approved):

server/commands.go
feedmeapples: i believe that's already the case https://github.com/temporalio/cli/blob/36afc181974f4a77fca287b432a457657a249bcc/helpprinter/helpprinter.go#L42C10-L42C21

feedmeapples (approved):

server/commands.go
lorensr: thanks!

common/defs-cmds.go
```diff
@@ -582,17 +582,20 @@ const WorkflowUpdateUsageText = `The ` + "`" + `temporal workflow update` + "`"
...
 `
 
-const StartDevUsageText = `The ` + "`" + `temporal server start-dev` + "`" + ` command starts the Temporal Server on ` + "`" + `localhost:7233` + "`" + `.
-The results of any command run on the Server can be viewed at http://localhost:7233.
-`
+const StartDevUsageText = `Start [Temporal Server](/concepts/what-is-the-temporal-server) on ` + "`" + `localhost:7233` + "`" + ` with:
+
+` + "`" + `temporal server start-dev` + "`" + `
+
+View the UI at http://localhost:8233.
```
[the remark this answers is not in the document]
lorensr: Think might do sentences in most USAGE sections, but here want it easily copyable for those in terminals without clickable links, so removing period ☺️

Issue #216, opened 2023-04-26
[Feature Request] workflow reset-batch should prompt when using query flag


Multiply:
### Is your feature request related to a problem? Please describe.
`temporal workflow reset-batch --query "..."` does not prompt the user with a count of workflows that will be affected.
A very open query could affect the whole namespace.
It's very easy to make mistakes here.

### Describe the solution you'd like
For most (if not all?) other commands that supports the `--query` flag, you're presented with a prompt before running the command.
A prompt similar to those would be ideal, showing the count of workflows affected.

Example from `temporal workflow terminate --query "..."`:

```
Will start a batch job operating on 123 Workflow Executions. Continue? Y/N
```

chaptersix:
Closing as resolved. The reset-batch command now counts matching workflows and prompts before proceeding: `Start batch against approximately N workflow(s)? y/N`. Added in https://github.com/temporalio/cli/pull/473.

PR #306, opened 2023-08-04
Refactor workflow batch reset with official API

feedmeapples:
## What was changed

WIP Reworked batch reset to work on top of the official API

## Why?

Standardizing operations with the official API

Resolve https://github.com/temporalio/cli/issues/216

## Checklist

1. Closes 
1. How was this tested:

added test `TestWorkflowReset_Batch`

1. Any docs updates needed?

### DIFF
--- a/batch/batch_commands.go
+++ b/batch/batch_commands.go
@@ -2,6 +2,7 @@ package batch
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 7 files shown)


josh-berry:
We're going to take a different approach, which is to get rid of the batch-specific command and add a `--query` option to `temporal workflow reset`.

PR #339, opened 2023-09-20
README and dev doc updates

josh-berry:
- Consolidate all dev docs into CONTRIBUTING.md
- Document the release process (such as I understand it)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,8 +37,7 @@ go build ./cmd/temporal
 ## Generate docs
 
 ```bash
-go build -o temporal-doc-gen ./cmd/docgen
-./temporal-doc-gen
+go run ./cmd/docgen
 ```
 
 Docs are generated as Markdown files in `./docs`.
@@ -105,9 +104,13 @@ If you're a Temporal engineer / code owner, here's how to do a release:
    the checksum of the source tarball. ([Example PR])
    - Homebrew should take care of rebottling the formula and updating the bottle
      checksums; you don't need to do this yourself.
+7. Re-generate the docs, and follow [the docs team's process] to update [the
+   CLI documentation].
 
 For reference, here are all of the places where releases get published:
 https://docs.temporal.io/cli/#installation
 
 [the temporal formula]: https://github.com/Homebrew/homebrew-core/blob/fc3afa61f03205d6d88f5443e805a7f782171aa2/Formula/t/temporal.rb
 [Example PR]: https://github.com/Homebrew/homebrew-core/commit/23f09be7fe7e2b00eee53a12db9a5a15c1c206ff
+[the docs team's process]: https://github.com/temporalio/documentation/blob/main/README.md#how-to-make-changes-to-this-repository
+[the CLI documentation]: https://github.com/temporalio/documentation/tree/main/docs-src/cli
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,15 @@ Build the `temporal` binary:
 go build ./cmd/temporal
 ```
 
+## Generate docs
+
+```bash
+go build -o temporal-doc-gen ./cmd/docgen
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -28,6 +28,15 @@ Build the `temporal` binary:
 go build ./cmd/temporal
 ```
 
+## Generate docs
+
+```bash
+go build -o temporal-doc-gen ./cmd/docgen
```
bergundy, line 34: I think there's a target in the Makefile (which I'm totally fine with removing BTW).

Another option is running with `go run`.

```
go run ./cmd/docgen
```

CONTRIBUTING.md
```diff
@@ -28,6 +28,15 @@ Build the `temporal` binary:
...
+./temporal-doc-gen
+```
+
+Docs are generated as Markdown files in `./docs`.
```
bergundy, line 38: We should say what to do with these generated docs, how do they get used in our documentation site?

CONTRIBUTING.md
```diff
@@ -73,3 +82,22 @@ To verify that this file is up to date execute:
 ```bash
 go run ./cmd/licensecheck
 ```
+
+## Release process
+
+If you're a Temporal engineer / code owner, here's how to do a release:
+
+1. Do some manual testing to make sure things look good--start a dev-server, run
+   some workflows, etc.
```
bergundy, line 91: Shouldn't the automated tests cover this?
I would rather we replace this with "verify that CI passes for the commit hash".

bergundy: (Which I see that you have below).

CONTRIBUTING.md
```diff
@@ -73,3 +82,22 @@ To verify that this file is up to date execute:
...
+2. Make sure CI is passing on `main`.
+3. Create a tag of the form `vX.Y.Z` off of `main` and push it to GitHub.
+4. GoReleaser will pick up the tag and build the release artifacts.
+5. Open a PR against Homebrew to modify [the temporal formula] to fetch the
+   source tarball using the latest release from GitHub.  Don't forget to update
+   the checksum of the source tarball. ([Example PR])
+   - Homebrew should take care of rebottling the formula and updating the bottle
+     checksums; you don't need to do this yourself.
+6. TODO: Describe how to update `temporal.download` links
```
bergundy, line 100: There's nothing to update there.

CONTRIBUTING.md
```diff
@@ -73,3 +82,22 @@ To verify that this file is up to date execute:
...
+
+[the temporal formula]: https://github.com/Homebrew/homebrew-core/blob/master/Formula/t/temporal.rb
```
bergundy, line 102: May want to use a permalink here in case this changes for whatever reason.

CONTRIBUTING.md
josh-berry: Good point. I don't know the answer to this, I'll do some research.

CONTRIBUTING.md
josh-berry: Yeah, the problem is nobody seems to fully-trust that CI is "good enough" right now—Chad's told me multiple times we need to do manual testing on things like server updates.

bergundy: It's fine as long as there's a step to verify that CI is green.

CONTRIBUTING.md
bergundy: I'd switch to `go run` though.

bergundy (approved):
Just one nit, otherwise LGTM.
Thanks.

josh-berry:
> Just one nit, otherwise LGTM. Thanks.

Just to clarify—is the nit your comment above about how to publish docs, or was there another one?

CONTRIBUTING.md
josh-berry: Added another item for this to the release process.

PR #355, opened 2023-10-12
Remove double setting of caller-type header

pdoerner:
## What was changed

Removed setting of `caller-type` header from `headers_provider` because it is already being set as part of `cliVersionHeaders` so the value was getting appended twice.

## Why?

No need to set the same value twice

## Checklist

1. Closes 

2. How was this tested:

Existing tests and quick local validation

3. Any docs updates needed?

### DIFF
--- a/headersprovider/headers_provider.go
+++ b/headersprovider/headers_provider.go
@@ -2,8 +2,6 @@ package headersprovider
 
 import (
 	"context"
-
-	h "github.com/temporalio/cli/headers"
 )
 
 type HeadersProvider interface {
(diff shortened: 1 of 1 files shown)


headersprovider/headers_provider.go
```diff
@@ -18,18 +16,12 @@ type grpcHeaderProvider struct {
...
-	return provider
-}
-
 func (a grpcHeaderProvider) GetHeaders(ctx context.Context) (map[string]string, error) {
 	return a.headers, nil
 }
 
 func SetGRPCHeadersProvider(headers map[string]string) {
-	headersProvider = newGrpcHeaderProvider(headers)
+	headersProvider = &grpcHeaderProvider{headers}
```
yiminc, line 24: where do we inject that caller-type header?

pdoerner: It's included with the other version and client-name headers: https://github.com/temporalio/cli/blob/main/headers/headers.go#L37

yiminc (approved):

Issue #351, opened 2023-10-06
[Bug] -input-file option not working with multiple JSON objects in the file


jdntra30:
### What are you really trying to do?

Trying to run a new workflow with the input-file flag. It works with one input, but encounters an issue for multiple inputs.

### Describe the bug

In the documentation for input-file flag, it says: "Passes optional input for the Workflow from a JSON file. If there are multiple JSON files, concatenate them and separate by space or newline. Input from the command line will overwrite file input."

So our python script creates a file with two JSON objects:
{ "key1" : "value1"} {"key2" : "value2"}

I tried with both space character and newline character. It doesn't work. It only works with one input.

With two inputs, I get this error: **"Error: input is not valid JSON: invalid character '{' after top-level value"**

### Environment/Versions

- OS and processor: Mac OS Ventura 13.4
- Temporal Version: Tried with both temporal version 0.10.6, and tctl version next
- Are you using Docker or Kubernetes or building Temporal from source? Docker with tctl. Built with the direct source for Temporal

josh-berry:
Sorry about that!

The docs are unfortunately just wrong about how `--input-file` works, and it seems we missed handling multiple arguments entirely. :/ I have a fix pending that allows for passing multiple `--input-file` options to pass distinct arguments (one per file), and clarifies in the help output how this works.

PR #361, opened 2023-10-16
Allow passing multiple `--input-file`s to workflow commands [#351]

josh-berry:
The documentation for how to pass multiple arguments was just wrong--`json.Unmarshal()` does not support parsing multi-line JSON inputs as separate values.

Fix this by accepting multiple `--input-file` flags, parsing each one separately, and turning each file into a distinct argument.

Also, check and throw an error if the user tries to use both `--input` and `--input-file`, since the old behavior--silently ignoring `--input` if `--input-file` is specified--was surprising.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -409,6 +409,10 @@ func ProcessJSONInput(c *cli.Context) (*commonpb.Payloads, error) {
 
 // read multiple inputs presented in json format
 func readJSONInputs(c *cli.Context) ([][]byte, error) {
+	if c.IsSet(FlagInput) && c.IsSet(FlagInputFile) {
+		return nil, fmt.Errorf("you may not combine --input and --input-file; please use one or the other")
+	}
+
 	if c.IsSet(FlagInput) {
 		inputsG := c.Generic(FlagInput)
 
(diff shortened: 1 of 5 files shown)


tests/workflow_test.go
```diff
@@ -23,6 +25,80 @@ const (
...
+	// Check that the workflow produced the result we expect--if it did, that
+	// means the CLI passed the arguments correctly.
+	var result interface{}
+	wf := client.GetWorkflow(context.Background(), "test", "")
+	err = wf.Get(context.Background(), &result)
+	s.NoError(err)
+
+	s.Assert().Equal("[1,\"two\",{\"three\":3},[\"a\",\"b\",\"c\"]]", result)
+}
+
```
dandavison, line 101: This isn't blocking, but just noting that running a real workflow to completion involves quite a lot of machinery that's not related to argument handling. It might be nice to come up with a more directed way to do this.

dandavison (approved):
LGTM!

common/util.go
```diff
@@ -432,14 +436,20 @@ func readJSONInputs(c *cli.Context) ([][]byte, error) {
...
 	} else if c.IsSet(FlagInputFile) {
-		inputFile := c.String(FlagInputFile)
-		// This method is purely used to parse input from the CLI. The input comes from a trusted user
-		// #nosec
-		data, err := os.ReadFile(inputFile)
-		if err != nil {
-			return nil, fmt.Errorf("unable to read input file: %w", err)
+		inputFiles := c.StringSlice(FlagInputFile)
+
+		args := [][]byte{}
```
cretz, line 441: Go tip/suggestion, size your slices if you know how big they'll need to be

```suggestion
		args := make([][]byte, 0, len(inputFiles))
```

Or if you want to use the index to populate it

```suggestion
		args := make([][]byte, len(inputFiles))
```

tests/workflow_test.go
```diff
@@ -23,6 +25,80 @@ const (
...
+		r.RegisterWorkflow(encodejson.Workflow)
+	})
+	defer w.Stop()
+
+	// Run the workflow to completion using the CLI.  (TODO: We unfortunately
+	// don't have a way to check the CLI output directly to make sure it prints
+	// the right result...)
+	err := cli.Run([]string{"", "workflow", "execute",
+		"--input", "1", "--input", "\"two\"", "--input", "{\"three\": 3}",
+		"--input", "[\"a\", \"b\", \"c\"]",
```
cretz, line 46: Go tip, use a raw string literal if you need to include a bunch of quotes (https://go.dev/ref/spec#String_literals):

```suggestion
		"--input", `["a", "b", "c"]`,
```

cretz (commented):
Added a couple tiny Go tips for future

tests/workflow_test.go
josh-berry: Thanks! Just sent you #364 to address these.

PR #366, opened 2023-10-19
Batch operation rate limit

stephanos:
## What was changed

Added a new flag `--rps` to the 4 batch operations.

Note that although `batch-reset` is a batch operation, it performs its work locally instead of on the server - so the flag does not apply here.

## Why?

To fix https://github.com/temporalio/temporal/issues/4926.

## Checklist

1. Closes OSS-1681

2. How was this tested:

3. Any docs updates needed? see https://github.com/temporalio/documentation/pull/2413


common/defs-flags.go
```diff
@@ -63,12 +63,13 @@ const (
...
 	FlagWorkflowIDTerminate        = "Terminate Workflow Execution with given Workflow Id."
+	FlagRPSDefinition              = "Limit for batch operation's requests per second."
 	FlagQueryCancel                = "Cancel Workflow Executions with given List Filter."
 	FlagQueryDelete                = "Delete Workflow Executions with given List Filter."
-	FlagQuerySignal                = "Signal Workflow Executions with given List List Filter."
+	FlagQuerySignal                = "Signal Workflow Executions with given List Filter."
+	FlagQueryReset                 = "Reset Workflow Executions with given List Filter"
 	FlagQueryTerminate             = "Terminate Workflow Executions with given List Filter."
 	FlagEventIDDefinition          = "The Event Id for any Event after WorkflowTaskStarted you want to reset to (exclusive). It can be WorkflowTaskCompleted, WorkflowTaskFailed or others."
-	FlagQueryResetBatch            = "Visibility Query of Search Attributes describing the Workflow Executions to reset. See https://docs.temporal.io/docs/tctl/workflow/list#--query."
```
stephanos, line 71: To align it with the other flags, I changed this into `FlagQueryReset` and standardized the text.

batch/batch_commands.go
```diff
@@ -83,8 +83,10 @@ func ListBatchJobs(c *cli.Context) error {
 // BatchTerminate terminate a list of workflows
 func BatchTerminate(c *cli.Context) error {
 	operator := common.GetCurrentUserFromEnv()
+	rps := float32(c.Float64(common.FlagRPS))
```
stephanos, line 86: AFAICT, the CLI library doesn't have a built-in `Float32` flag, so we need to downcast here. Which is fine; the required precision we expect here will most likely never be more than 2 decimals.

go.mod
```diff
@@ -16,24 +16,24 @@ require (
 	github.com/temporalio/tctl-kit v0.0.0-20230328153839-577f95d16fa0
 	github.com/temporalio/ui-server/v2 v2.18.2
 	github.com/urfave/cli/v2 v2.25.7
-	go.temporal.io/api v1.24.0
+	go.temporal.io/api v1.25.0
```
stephanos, line 19: Upgraded to new API version to pull in the `max_operations_per_second` field.

yiminc (approved):

PR #368, opened 2023-10-19
Fix sources of test instability

### LINKED ISSUE #338: [Bug] Flaky test: TestCreateDataDirectory_ExistingDirectory
There is a flaky test on Windows in `app`:

https://github.com/temporalio/cli/actions/runs/6252251042/job/16975099634?pr=321

```
 error cleaning up temp dir: remove C:\Users\RUNNER~1\AppData\Local\Temp\temporal_test\TestCreateDataDirectory_MissingDirectory\foo\bar\baz.db: The process cannot access the file because it is being used by another process.
2023/09/20 17:56:29 INFO  No logger configured for temporal client. Created default one.
error cleaning up temp dir: remove C:\Users\RUNNER~1\AppData\Local\Temp\temporal_test\TestCreateDataDirectory_ExistingDirectory\foo.db: The process cannot access the file because it is being used by another process.
--- FAIL: TestCreateDataDirectory_ExistingDirectory (2.77s)
```

Seems to fail intermittently on Windows only—I'd guess because something is still holding `foo.db` open during test cleanup. We should figure out what that is and fix it so this test passes consistently.

(I'm guessing this same condition exists on Mac/Linux, too, but we never see errors because you're allowed to delete files while they're still open on UNIXy OSes.)

josh-berry:
- Wait up to 1 second for the server to become healthy before assuming it's failed. After the server starts accepting connections, we were assuming it should be healthy immediately, but that's not always true.

- Make sure the server finishes shutting down before exiting the CLI, because otherwise it will hold open files, and this intermittently breaks tests on Windows (which, unlike POSIXy OSes, will refuse to delete open files).  Closes #338.  I hope.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -320,7 +320,7 @@ func assertServerHealth(ctx context.Context, t *testing.T, opts sdkclient.Option
 		t.Error(clientErr)
 	}
 
-	// Give the server 10 seconds to become healthy.
+	// Give the server 1 second to become healthy.
 	for i := 0; i < 10; i++ {
 		_, clientErr = c.CheckHealth(ctx, nil)
 		if clientErr == nil {
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -320,8 +320,16 @@ func assertServerHealth(ctx context.Context, t *testing.T, opts sdkclient.Option
 		t.Error(clientErr)
 	}
 
-	if _, err := c.CheckHealth(ctx, nil); err != nil {
-		t.Error(err)
+	// Give the server 10 seconds to become healthy.
+	for i := 0; i < 10; i++ {
+		_, clientErr = c.CheckHealth(ctx, nil)
+		if clientErr == nil {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	if clientErr != nil {
+		t.Error(clientErr)
 	}
 
 	// Check for pollers on a system task queue to ensure that the worker service is running.
--- a/server/commands.go
+++ b/server/commands.go
@@ -337,6 +337,7 @@ func NewServerCommands(defaultCfg *sconfig.Config) []*cli.Command {
 				if err := s.Start(); err != nil {
 					return cli.Exit(fmt.Sprintf("Unable to start server. Error: %v", err), 1)
 				}
+				s.Stop()
 				return cli.Exit("All services are stopped.", 0)
 			},
 		},


josh-berry:
Manual testing:
- [ ] Verify `temporal server start-dev` continues to work as expected (and does not immediately shut down the server).
- [ ] Re-run all the GitHub actions a few times to be sure the test flakiness has actually gone away. (Not all of it was reproducible on either of my local machines.)

app/app_test.go
```diff
@@ -320,8 +320,16 @@ func assertServerHealth(ctx context.Context, t *testing.T, opts sdkclient.Option
 		t.Error(clientErr)
 	}
 
-	if _, err := c.CheckHealth(ctx, nil); err != nil {
-		t.Error(err)
+	// Give the server 1 second to become healthy.
```
cretz, line 323: A client should not return from `Dial` unless the server is available. This is an important feature of our SDKs. The `Dial` call invokes `GetSystemInfo` eagerly to get information and check health. There is no reason that call should work but `CheckHealth` should not.

bergundy: You might want to refactor this whole thing to use the Go SDK's built-in [DevServer](https://pkg.go.dev/go.temporal.io/sdk@v1.25.1/testsuite#DevServer) with a specified path and use the dev server's client.

josh-berry: > A client should not return from `Dial` unless the server is available and healthy. This is an important feature of our SDKs. The `Dial` call invokes `GetSystemInfo` eagerly to get information and check health. There is no reason that call should work but `CheckHealth` should not.

I see. We are often seeing failures like this one:

https://github.com/temporalio/cli/actions/runs/6317569213/job/17154638693#step:5:199

...and simply waiting a second makes them go away. I do see the GetSystemInfo call you mention:

- [`resp, err := wc.workflowService.GetSystemInfo(ctx, &workflowservice.GetSystemInfoRequest{})`](https://github.com/temporalio/sdk-go/blob/06c90fb3698708ef11d1992c8d69883d26a20c31/internal/internal_workflow_client.go#L1162)

But it seems like [`CheckHealth()`](https://github.com/temporalio/sdk-go/blob/06c90fb3698708ef11d1992c8d69883d26a20c31/internal/internal_workflow_client.go#L1118) does a gRPC health check, which is different:

- [`resp, err := healthpb.NewHealthClient(wc.conn).Check(ctx, &healthpb.HealthCheckRequest{ ...`](https://github.com/temporalio/sdk-go/blob/06c90fb3698708ef11d1992c8d69883d26a20c31/internal/internal_workflow_client.go#L1124)

I do not know why GetSystemInfo would succeed when the gRPC health check fails. I can dig into server internals to figure out why, but it seems the behavior is indeed different.

josh-berry: > You might want to refactor this whole thing to use the Go SDK's built-in [DevServer](https://pkg.go.dev/go.temporal.io/sdk@v1.25.1/testsuite#DevServer) with a specified path and use the dev server's client.

That would defeat the whole point of this test, which is to launch the dev server using Temporal CLI.

bergundy: How so? The dev server launches Temporal CLI for you.

josh-berry: > How so? The dev server launches Temporal CLI for you.

It does, but by default it doesn't actually test the version of the CLI in CI, because it downloads a released version of the CLI or pulls from the user's PATH. I want to test the version in the repo (possibly with whatever random developer changes are there). That's only possible by invoking the relevant CLI commands directly.

To test with the right version, I would have to make sure the CLI binary is actually built as part of running the tests, and this adds complexity to the tests and an additional layer of abstraction that I think is harmful—it obscures what the test actually does, because the arguments to `temporal server start-dev` are not immediately visible in the test, and we are basically inserting an external dependency in between the test and the thing we're testing. This is also likely to make the test itself more brittle (e.g. if something changes in go-sdk).

Regardless, none of that refactoring would actually help with these test failures, and such a change would be out-of-scope for this PR anyway. So I think we can probably continue this elsewhere.

cretz: > I do not know why GetSystemInfo would succeed when the gRPC health check fails

Me neither, it's important that this is not the case or we have bigger problems with the SDK/server contract/expectations I think.

josh-berry: @cretz How do you feel about:
1. Merging this as-is
2. Opening a separate issue to track down why the gRPC health check is failing even when GetSystemInfo returns successfully
3. Reverting the polling loop for the gRPC health check in this PR once (2) is resolved

cretz: Sure, anything to fix the problem I'm ok with and I don't see why we need to block this issue if it's already broken.

cretz (approved):

app/app_test.go
josh-berry: Opened https://github.com/temporalio/temporal/issues/5015

bergundy: > > How so? The dev server launches Temporal CLI for you.
> 
> It does, but by default it doesn't actually test the version of the CLI in CI, because it downloads a released version of the CLI or pulls from the user's PATH. I want to test the version in the repo (possibly with whatever random developer changes are there). That's only possible by invoking the relevant CLI commands directly.

You could do that with the test framework by giving it a path to an existing binary.

> To test with the right version, I would have to make sure the CLI binary is actually built as part of running the tests, and this adds complexity to the tests and an additional layer of abstraction that I think is harmful—it obscures what the test actually does, because the arguments to `temporal server start-dev` are not immediately visible in the test, and we are basically inserting an external dependency in between the test and the thing we're testing. This is also likely to make the test itself more brittle (e.g. if something changes in go-sdk).

I think we can discuss the pros and cons here.
One pro is that we don't have to rewrite logic that waits for the server to be "ready".

> Regardless, none of that refactoring would actually help with these test failures, and such a change would be out-of-scope for this PR anyway. So I think we can probably continue this elsewhere.

SGTM

Issue #242, opened 2023-05-18
[Feature Request] Error when a port is taken


lorensr:
### Is your feature request related to a problem? Please describe.

When I do `temporal server start-dev --ui-port 3000` and 3000 is already taken, I see no indication in the output that that's the case.

### Describe the solution you'd like

Output:

```
Error: port 3000 is taken. Select another port with the --ui-port option.
```

and exit. Or better yet, increment until we get an open port:

```
Port 3000 is taken. Using port 3001 instead.
🆙
Temporal Server is running at: localhost:7233
Web UI is running at: http://localhost:3001
```

Could do for Server port as well as UI port?

lucasmls:
Hey @lorensr, how you doing? Hope you're well :)

Can i get this one? Thanks!

lorensr:
Sorry for the delay—that would be great, thanks!

lucasmls:
Just so we're all aligned. I plan to open the PR by this weekend. Sorry about the delay.

Quinn-With-Two-Ns:
Testing on the rewrite branch it appears the command errors if the UI port is taken, but not the server port.
```
➜  temporal git:(cli-rewrite) ✗ ./temporal server start-dev --ui-port 3000
time=2024-03-01T08:53:58.904 level=ERROR msg="Failed to start gRPC listener" error="listen tcp 127.0.0.1:7233: bind: address already in use" service=frontend address=127.0.0.1:7233
Temporal server is running at: localhost:7233
Web UI is running at: http://localhost:3000
Metrics available at: http://localhost:50598/metrics
```

```
➜  temporal git:(cli-rewrite) ✗ ./temporal server start-dev --ui-port 3000
time=2024-03-01T08:54:18.517 level=ERROR msg="Failed to start gRPC listener" error="listen tcp 127.0.0.1:7233: bind: address already in use" service=frontend address=127.0.0.1:7233
{"time":"2024-03-01T08:54:18.520478-08:00","level":"FATAL","prefix":"echo","file":"server.go","line":"149","message":"listen tcp 127.0.0.1:3000: bind: address already in use"}
```

cretz:
I wonder if we should report this to the server. Server start should fail if frontend can't bind IMO.

bergundy:
Looks like the server does fail (pre CLI rewrite and with `make start-sqlite` in the server repo):

```
[...]

PR #373, opened 2023-10-28
If server or UI port is unavailable, try to find an available one

lucasmls:
## What was changed

If the server or UI port is already in use by another process when running the CLI, the CLI will currently crash. The aim of this pull request is to modify this behavior; instead of crashing, the CLI will increment the port number by 1 until it finds an available port.

Screenshot 2023-10-27 at 21 53 04

---
I know this is a naive approach. I think a better way to handle this would be to ask the user to manually input an available port. I chose to stick with the easiest solution for the problem. However, if we all agree that prompting the user for the port is a better approach, I'm open to developing it.

**Another question**
Loren suggested that it would be nice if we added a log like this: _"Port 3000 is taken. Using port 3001 instead."_
I find it nice, but at the context that the code of this PR is being executed i don't have access to the logger. Any tips on this?

## Checklist

**1. Closes** 
https://github.com/temporalio/cli/issues/242

**2. How was this tested:**

- I've added the unit test `TestUIPortAlreadyInUse`, which locks the default ports and their subsequent ones (**server**: `7233` and `7234`, **ui**: `8233` and `8234`). By utilizing the previously coded helper `assertServerHealth`, I ensure that the server is running healthily.
- I also tested it manually. I opened another process using the default ports and ensured that the server was able to start and operate in the newly allocated ports.
Screenshot 2023-10-27 at 22 09 00

**3. Any docs updates needed?**

This alters some existing behavior. Should we include this new behavior in the CLI command description?


server/commands.go
```diff
@@ -387,3 +388,30 @@ func getDynamicConfigValues(input []string) (map[dynamicconfig.Key][]dynamicconf
...
+		return false
+	}
+
+	if conn != nil {
+		_ = conn.Close()
+		return true
+	}
+
+	return false
+}
```
lucasmls, line 417: I've added this code here for the sake of simplicity. I questioned whether it makes sense to move it to the [PortProvider](https://github.com/temporalio/cli/blob/main/server/config/freeport.go#L40) struct. Alternatively, if there's a better place for it, please let me know. What do all of you think?

cretz:
>  The aim of this pull request is to modify this behavior; instead of crashing, the CLI will increment the port number by 1 until it finds an available port.

I am not sure it is desirable to use anything but the given/default port when starting the CLI (regardless of port purpose, UI, server, etc). There are programmatic expectations for ports when people start the CLI, we should not continue if we can't meet those expectations. Similarly, it that port is in use by another instance of the UI and someone starts the CLI, they will blindly go to the common UI port unaware it's another instance. We want users to solve the port ambiguity by picking another or stopping what is already using the port, not silently picking another for them.

If we want to suggest an alternative, like supporting `--ui-port random`, that may be more acceptable than changing existing behavior to instead swallow errors on fixed port ambiguity.

server/commands.go
josh-berry: I agree changing PortProvider, or just using it outright, makes sense. (I think it already has some functionality for picking a free port, so you don't have to reimplement this.)

josh-berry (commented):
Thanks for sending this in!

I agree with Chad that this should be opt-in, so as to avoid breaking any scripts etc. that may have been written around the assumption that starting the server will fail if any of the ports are in use.

Issue #197, opened 2023-04-17
[Feature Request] Better help strings


bergundy:
Some of the help strings the CLI prints out today are confusing and unhelpful.

* [ ] `temporal server`

Today's output:

    NAME:
       temporal server - Commands for managing the Temporal Server.
    
    USAGE:
       Server commands allow you to start and manage the [Temporal Server](/concepts/what-is-the-temporal-server) from the command line.
    
       Currently, `cli` server functionality extends to starting the Server.
    
       Server commands follow this syntax:
       `temporal server COMMAND`
    Suggestion:

    NAME:
       temporal server - Start a [Temporal Server](/concepts/what-is-the-temporal-server).
    
    USAGE:
       Use the subcommands such as `start-dev` to start the server:
       `temporal server COMMAND`* [ ] `temporal server start-dev`

Today's output:

```
NAME:
   temporal server start-dev - Start Temporal development server.

USAGE:
   The temporal server start-dev command starts the Temporal Server on localhost:7233.
   The results of any command run on the Server can be viewed at http://localhost:7233.
```

Suggestion:

    NAME:
       temporal server start-dev - Start a Temporal development server.
    
    USAGE:
       `temporal server start-dev [OPTIONS]`We need to review every single command so I decided to stop here but this is critical to fix before 1.0.0.

josh-berry:
Done in #561.

PR #393, opened 2023-11-18
Improve help text

lorensr:
Continuing from #304 working on #197


lorensr:
I'm not grokking these test failures, like this:

image

comes from this test

image

which doesn't seem to have anything I changed?

josh-berry:
@lorensr It looks like you duplicated a `--help` flag somewhere, because the panic in the test failure seems to be saying that `--help` was redefined (/defined multiple times).

cretz:

PR #399, opened 2023-12-09
Enhance task-queue describe

dnr:
## What was changed
- `temporal task-queue describe` prints the `TaskQueueStatus` parts of the `DescribeTaskQueue` response
- It now takes a `--partitions` flag to query for multiple partitions.

Partition fan-out is more properly done server-side, but that'll take longer and in the meantime (and on older server versions) this is still useful.

## Why?
More useful for debugging.

## Checklist

1. Closes 

2. How was this tested:
Manually

3. Any docs updates needed?

### DIFF
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -135,7 +135,7 @@ const (
 	// Task Queue flags
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
-	FlagPartitionsDefinition    = "Query for all partitions up to this number"
+	FlagPartitionsDefinition    = "Query for all partitions up to this number (experimental+temporary feature)"
 
 	// Namespace update flags
 	FlagActiveClusterDefinition           = "Active cluster name."
--- a/common/defs-flags.go
+++ b/common/defs-flags.go
@@ -135,6 +135,7 @@ const (
 	// Task Queue flags
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
+	FlagPartitionsDefinition    = "Query for all partitions up to this number"
 
 	// Namespace update flags
 	FlagActiveClusterDefinition           = "Active cluster name."
--- a/taskqueue/task_queue_commands.go
+++ b/taskqueue/task_queue_commands.go
@@ -1,6 +1,7 @@
 package taskqueue
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


common/defs-flags.go
```diff
@@ -135,6 +135,7 @@ const (
 	// Task Queue flags
 	FlagTaskQueueName           = "Name of the Task Queue."
 	FlagTaskQueueTypeDefinition = "Task Queue type [workflow|activity]"
+	FlagPartitionsDefinition    = "Query for all partitions up to this number"
```
bergundy, line 138: Since this may be removed in the future when implemented server-side, can we mark this flag as experimental and add a note?

taskqueue/task_queue_commands.go
```diff
@@ -9,37 +10,98 @@ import (
...
-	for _, e := range resp.Pollers {
-		items = append(items, e)
+	err = output.PrintItems(c, statuses, opts)
+	if err != nil {
+		return err
 	}
-	return output.PrintItems(c, items, opts)
+
+	opts = &output.PrintOptions{
+		Fields: []string{"Partition", "PollerInfo.Identity", "PollerInfo.LastAccessTime", "PollerInfo.RatePerSecond", "VersionCaps.BuildId", "VersionCaps.UseVersioning"},
```
bergundy, line 102: Can we spell `Capabilities` out or use a different field name?

taskqueue/task_queue_commands.go
```diff
@@ -9,37 +10,98 @@ import (
...
+
+	type statusWithPartition struct {
+		Partition int `json:"partition"`
+		taskqueuepb.TaskQueueStatus
+	}
+	type pollerWithPartition struct {
+		Partition int `json:"partition"`
+		taskqueuepb.PollerInfo
+		// copy this out to display nicer in table or card, but not json
+		VersionCaps *commonpb.WorkerVersionCapabilities `json:"-"`
```
bergundy, line 47: Is this still printed in the JSON view, just not flattened?

common/defs-flags.go
dnr: definitely

taskqueue/task_queue_commands.go
dnr: yeah, it's in pollerinfo. but the table+card output functions can't handle doubly-nested fields so I had to pull it out. but json can of course (if I handle it separately)

taskqueue/task_queue_commands.go
dnr: I would use `WorkerVersionCapabilities` but that would shadow the field in `PollerInfo`. How about `Versioning`?

note that the first part of the field name is not displayed in table view, so it's mostly not visible. it is in card view or as the input to `--fields`

dnr (commented):
btw, this is admittedly a bit of a hack: the output library is not meant to be used twice in one invocation, it won't work well with a pager. but there are already a bunch of pager-related bugs, and we'll probably rewrite this all soon. I just didn't want to throw away the code I used for debugging.

bergundy (approved):

PR #406, opened 2023-12-28
Fix the "delete namespace" prompt

### LINKED ISSUE #405: [Bug] Cannot delete namespace that has any upper-case letters
Create namespace (any would work as long as it has at least 1 upper-case letter):

          temporal operator namespace create MyNS

Try to delete namespace:

          temporal operator namespace delete MyNS

Namespace is not deleted.
Works fine if namespace includes only lower case letters and numbers

josh-berry:
We were improperly lower-casing the user's input when asking them to confirm the namespace name, and then providing no feedback to the user if the name didn't match, leaving them with the false impression that the namespace was deleted even though we didn't even try.

Closes #405


namespace/namespace_commands.go
```diff
@@ -303,7 +303,7 @@ func DeleteNamespace(c *cli.Context) error {
 
 	promptMsg := color.Red(c, "Are you sure you want to delete namespace %s? Type namespace name to confirm:", ns)
 	if !common.Prompt(promptMsg, c.Bool(common.FlagYes), ns) {
```
bergundy, line 305: We shouldn't print the prompt if the user wants to auto-confirm.

josh-berry: We print both the prompt and response, which I think is okay for this hotfix. I don't really want to expand the scope of this PR since we're working on a rewrite.

common/util.go
```diff
@@ -501,16 +501,16 @@ func Prompt(msg string, autoConfirm bool, expectedInputs ...string) bool {
...
+		text = strings.TrimRight(text, "\n")
 	}
 	fmt.Println()
 
 	if len(expectedInputs) == 0 {
 		return true
 	}
 
-	textLower := strings.ToLower(strings.TrimSpace(text))
 	for _, expectedInput := range expectedInputs {
```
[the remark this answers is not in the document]
josh-berry: Same as below, I don't really want to make any more changes than absolutely necessary since this is a hotfix.

bergundy (approved):
Nothing blocking. We can handle these comments in the rewrite.

@cretz FYI.

PR #408, opened 2024-01-03
Insecure CSRF cookie

ndtretyak:
## What was changed

CSRF cookie is now set without `Secure` attribute.

## Why?

Now CSRF cookie is `Secure` but development servers are usually used locally without HTTPS. For some reason, it works fine in Chrome (there is an exception for `localhost` domain in `Secure` handling) but does not work in Safari.

Since `start-dev` command starts a **development** server, I think it is safe to use insecure cookies here.

## Checklist

1. Closes 

2. How was this tested:

Checked locally with different browsers.

3. Any docs updates needed?

No.


josh-berry (approved):

josh-berry:
Thanks for the PR!

PR #415, opened 2024-01-18
Embed fallback tzdata

dnr:
## What was changed
Embed time zone database in case platform database can't be found.

## Why?
This ensures time zones are always available, even on Windows which does not supply a time zone database (at least in a format that Go understands).


bergundy (approved):
Can you confirm that this fixes the issue we saw?

bergundy:
I wouldn't go as far as requiring an automated test, just manual confirmation.

dnr:
> Can you confirm that this fixes the issue we saw?

I verified it by running the cli dev-server in an empty docker container so that it couldn't find a platform database. It failed without the embedded db and succeeded with.

PR #432, opened 2024-01-30
CLI Refresh: `workflow signal` command

cretz:
## What was changed

Add support for `workflow signal` command. The functionality is mostly unchanged from current CLI, but the output may be different.

Improvements over current CLI:

* JSON output support (empty for single-workflow, contains batch job ID for batch)
* Use non-success exit code if prompt confirmation fails on batch
* Clarify batch count is approximate (due to race between count request and actual batch)
* `--reason` not required for batch (same default format as UI)


temporalcli/commands.workflow.go
```diff
@@ -41,3 +94,68 @@ func (*TemporalWorkflowTraceCommand) run(*CommandContext, []string) error {
...
+
+	// Check query is set properly
+	if s.Query == "" {
+		return nil, nil, fmt.Errorf("must set either workflow ID or query")
+	} else if s.WorkflowId != "" {
+		return nil, nil, fmt.Errorf("cannot set workflow ID when query is set")
+	} else if s.RunId != "" {
+		return nil, nil, fmt.Errorf("cannot set run ID when query is set")
+	} else if s.Reason == "" {
+		return nil, nil, fmt.Errorf("reason required when query is set")
```
cretz, line 123: Removing this requirement by request and setting a default value

cretz: Done

temporalcli/commands_test.go
```diff
@@ -47,13 +50,17 @@ func (h *CommandHarness) Close() {
...
 	// Split into lines, then check each piece is present
 	lines := strings.Split(text, "\n")
 	for _, line := range lines {
 		foundAll := true
+		lastFoundPieceIndex := -1
 		for _, piece := range pieces {
-			if !strings.Contains(line, piece) {
+			if index := strings.Index(line, piece); index > lastFoundPieceIndex {
+				lastFoundPieceIndex = index
+			} else {
```
cretz, line 63: @dandavison - here is update for forcing order

temporalcli/commands.workflow.go
```diff
@@ -41,3 +95,77 @@ func (*TemporalWorkflowTraceCommand) run(*CommandContext, []string) error {
...
+
+	// Default the reason if not set
+	reason := s.Reason
+	if reason == "" {
+		username := "<unknown-user>"
+		if u, err := user.Current(); err != nil && u.Username != "" {
+			username = u.Username
+		}
+		reason = "Requested from CLI by " + username
+	}
```
cretz, line 147: Current CLI requires reason to be set by user. UI defaults to something like `Terminated from Web UI by foo@bar.com`, so we follow a similar format, but we just do "Requested" instead of specific verb.

temporalcli/commands_test.go
dandavison: Great, thanks for adding that! This looks like it will allow the pieces to match in an overlapping way -- I think we need to set `lastFoundPieceIndex` to the index of the end of `piece`.

cretz: I think it's ok to be overlapping, it's still in order by start of piece

dandavison: No, it's clear to me that when users make a call like `s.ContainsOnSameLine(out, "Namespace", "default")` they mean

> there exists a line which contains 'Namespace' and 'default', in that order (not overlapping).

I've never seen a test framework that allows them to be overlapping or out of order. I think that your `s.ContainsOnSameLine(out, "Namespace", "default")` call is nice and readable -- nicer to read than a regexp. But how about implementing it as a regular expression match call? That's what I'd do in most languages.

cretz: Updated

temporalcli/commands_test.go
```diff
@@ -47,13 +50,17 @@ func (h *CommandHarness) Close() {
...
 	// Split into lines, then check each piece is present
 	lines := strings.Split(text, "\n")
 	for _, line := range lines {
 		foundAll := true
+		lastEndIndex := -1
 		for _, piece := range pieces {
-			if !strings.Contains(line, piece) {
+			if index := strings.Index(line, piece); index > lastEndIndex {
+				lastEndIndex = index + len(piece)
+			} else {
```
dandavison, line 63: Ha, algorithms like this are hard to get right first time without unit tests! I suggest we use a regular-expression based implementation, i.e. form a regex by joining `pieces` with whatever separator that gives the separator semantics that we think callers will want in the common case. That way the behavior of the utility is clear to anyone who looks at the code, and they can use their own regex assertion if they want different behavior. I'd also be OK with deleting this utility and requiring callers to use a regex explicitly.

As it is it still has bugs. The most serious I'm seeing is that `ContainsOnSameLine("a b a", "b", "a")` will fail, when it should pass.

But also `ContainsOnSameLine("ab", "a", "b")` will fail (no gap between them) and yet `ContainsOnSameLine("axb", "a", "b")` will pass.

And related to the first bug, `ContainsOnSameLine("a a", "a", "a")` will fail.

cretz: This is why I wanted it simple and ask users to use regex if they needed more instead of increase the rules of this little helper. I regret not sticking to the original simple/looser need. At this point I think we should just document the limits and tell people to use regex if they want.

dandavison: The original version allowed the pieces to be in any order and overlapping, which isn't appropriate for a test harness assertion utility.

What do you see as the downside of implementing `ContainsOnSameLine` as a regex search for `pieces` joined by something like `.*` or `.+` or `.*\s.*`? However it's implemented, the resulting function will have _some_ behavior, which we need to make clear to callers. A regex is the best way to make that clear to callers as I see it.

Or, as I mentioned above, we could just delete the utility and let callers decide on their regex.

cretz: > What do you see as the downside of implementing ContainsOnSameLine as a regex search for pieces joined by something like .* or .+ or .*\s.*?

The continual requirements. "Can you strip colors too?", "Can you make it case insensitive?", "Can you make it only allow whole words instead of partial match?", "Can you allow this to work on adjacent lines instead of just one?", etc etc. "But it isn't appropriate for an stdout test harness utility to not strip out ANSI codes", "But it isn't appropriate for this test harness utility to not check whole words", etc. Each has their own opinion and it was just meant to be a little code reuse thing, but I didn't want to name it `ContainsAnywhereOnSameLineInAnyOrderMayOverlap`.

> Or, as I mentioned above, we could just delete the utility and let callers decide on their regex.

This may be best, but I still want some code reuse and I accept the limitations of the little helper.

bergundy (approved):
LGTM.

Not much to add.

Only thing I'm unsure of is if we want to print the success on non-json output via the logger to stderr.
Seems like the rest of the output for this command goes to stdout and if I want to grep for the job ID (yes some prefer that over jq), I have to pipe stderr which is a bit awkward and non-standard.

temporalcli/commands_test.go
dandavison: > The continual requirements. "Can you strip colors too?", "Can you make it case insensitive?", "Can you make it only allow whole words instead of partial match?", "Can you allow this to work on adjacent lines instead of just one?", etc etc. "But it isn't appropriate for an stdout test harness utility to not strip out ANSI codes", "But it isn't appropriate for this test harness utility to not check whole words", etc.

My suggestions that a string matching function should require order and not allow overlapping matches are perfectly reasonable and uncontroversial and it's not appropriate to attempt to ridicule them like that.

dandavison (changes requested):
As discussed, there are bugs in the test helper function that need fixing.

cretz:
> Only thing I'm unsure of is if we want to print the success on non-json output via the logger to stderr.

I will update the started-batch statement to be a stdout line instead of a log line

temporalcli/commands_test.go
cretz: Updated to support regex-based match

dandavison (approved):

temporalcli/commands_test.go
dandavison: Thanks, that looks great to me.

temporalcli/commands.workflow.go
```diff
@@ -22,8 +32,52 @@ func (*TemporalWorkflowResetBatchCommand) run(*CommandContext, []string) error {
...
+	defer cl.Close()
+
+	// Get input payloads
+	input, err := c.buildRawInputPayloads()
+	if err != nil {
+		return err
+	}
+
+	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl)
```
dandavison, line 48: Did this just happen to be the first command implemented for which batch was an option?

temporalcli/commands.workflow.go
```diff
@@ -22,8 +32,52 @@ func (*TemporalWorkflowResetBatchCommand) run(*CommandContext, []string) error {
...
+			SignalOperation: &batch.BatchOperationSignal{
+				Signal:   c.Name,
+				Input:    input,
+				Identity: clientIdentity(),
+			},
+		}
+		if err := startBatchJob(cctx, cl, batchReq); err != nil {
+			return err
+		}
+	}
```
dandavison, line 79: I believe that the contract of `workflowExecOrBatch` is such that one or other of the above conditions must be true, so I'd like the code not to appear to allow neither to be executed.

temporalcli/commands.workflow.go
```diff
@@ -41,3 +95,77 @@ func (*TemporalWorkflowTraceCommand) run(*CommandContext, []string) error {
 func (*TemporalWorkflowUpdateCommand) run(*CommandContext, []string) error {
 	return fmt.Errorf("TODO")
 }
+
```
dandavison, line 98: Although it's a private function, it would be good to document this something like
> This function guarantees that exactly one of the 3 returned values will be non-nil.

temporalcli/commands.workflow.go
cretz: Yes

temporalcli/commands.workflow.go
cretz: Yes, this is leftover from previous code where it could return all nils when you declined the prompt, but I decided to change that to be an error and did not change this.

temporalcli/commands.workflow.go
cretz: Concur

dandavison: OK great, I can make these minor changes.

PR #434, opened 2024-01-31
Expose schedule memo and search attributes in list

### LINKED ISSUE #433: Updating Schedule with memo option not updating memo
### Describe the bug

Doing `temporal schedule update --schedule-id ... --memo $KEY=$VALUE` does not update the memo of the Schedule.

### Minimal Reproduction

```
# Create Schedule
temporal schedule create --schedule-id foo --workflow-type Foo --task-queue default --memo bar=1

# Describe Schedule shows "Memo" with "bar": "1"
temporal schedule describe -schedule-id foo -o json

# Update Schedule
temporal schedule update --schedule-id foo --workflow-type Foo --task-queue default --memo bar=2

# Describe Schedule still shows "Memo" with "bar": "1" instead of "bar": "2"
temporal schedule describe -schedule-id foo -o json
```

### Environment/Versions
Running in Docker with this version
```
$ temporal -v
temporal version 0.10.7 (server 1.22.2) (ui 2.21.3)
```

### Additional context
Is there also any way to get the memo returned with `temporal schedule list`? Some SDKs like the Python one returns memo in the list results: https://python.temporal.io/temporalio.client.ScheduleListDescription.html#memo.

dnr:
## What was changed
Expose schedule's memo and search attributes in `schedule list` command.
Note these are not printed by default, you can use either `-o json` for a full dump or selectively enable them with `--fields ScheduleId,SearchAttributes,Memo`.
The values are always converted to strings for now. In the rewrite we can handle types better.

## Why?
Fixes #433 

## Checklist
2. How was this tested:
manually

### DIFF
--- a/schedule/schedule_commands.go
+++ b/schedule/schedule_commands.go
@@ -476,18 +476,8 @@ func DescribeSchedule(c *cli.Context) error {
 		item.LastRunActualTime = ra.ActualTime
 		item.LastRunExecution = ra.StartWorkflowResult
 	}
-	if fields := resp.Memo.GetFields(); len(fields) > 0 {
-		item.Memo = make(map[string]string, len(fields))
-		for k, payload := range fields {
-			item.Memo[k] = converter.GetDefaultDataConverter().ToString(payload)
-		}
-	}
-	if fields := resp.SearchAttributes.GetIndexedFields(); len(fields) > 0 {
-		item.SearchAttributes = make(map[string]string, len(fields))
-		for k, payload := range fields {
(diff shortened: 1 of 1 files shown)


cretz:
Hrmm, memo and search attributes aren't even shown when listing workflows. Should we do this there too?

(CLI refresh project has no concept of customizing list fields currently so we may have to add that concept)

dnr:
I'm not sure customizing list fields is that useful. The key point to me is that the list view should be fairly readable, though it may omit some available information, and if you want all the information, you can use the json view and format manually.

dnr:
Going to close this PR since we're rewriting things. This should be incorporated into #459.

PR #435, opened 2024-02-05
Implement `workflow cancel` command

tdeebswihart:
# What was changed

I implemented `temporal workflow cancel`, which should behave the same as in the existing CLI except for the same caveats/improvements as in #432


cretz (approved):
Looks great!

cretz:
May have to resolve conflicts and regen commands to fix CI

PR #436, opened 2024-02-05
Implement `workflow terminate` command

tdeebswihart:
# What was changed

I implemented `temporal workflow terminate`, which should behave the same as in the existing CLI except for the same caveats/improvements as in #432.

I extended the existing `SingleWorkflowOrBatch` functions to allow a command to opt-in to supplying a reason alongside a workflow id as terminate allows for this.

### DIFF
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -84,8 +84,40 @@ func (*TemporalWorkflowStackCommand) run(*CommandContext, []string) error {
 	return fmt.Errorf("TODO")
 }
 
-func (*TemporalWorkflowTerminateCommand) run(*CommandContext, []string) error {
-	return fmt.Errorf("TODO")
+func (c *TemporalWorkflowTerminateCommand) run(cctx *CommandContext, _ []string) error {
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, func(p *singleOrBatchParams) {
+		// You're allowed to specify a reason when terminating a workflow
+		p.AllowReasonWithWorkflowID = true
+	})
+	if err != nil {
+		return err
+	}
+
+	// Run single or batch
+	if exec != nil {
+		err = cl.TerminateWorkflow(cctx, exec.WorkflowId, exec.RunId, c.Reason)
+		if err != nil {
+			return fmt.Errorf("failed to terminate workflow: %w", err)
+		}
+		cctx.Printer.Println("Workflow terminated")
+	} else if batchReq != nil {
+		batchReq.Operation = &workflowservice.StartBatchOperationRequest_TerminationOperation{
+			TerminationOperation: &batch.BatchOperationTermination{
+				Identity: clientIdentity(),
+			},
+		}
+		if err := startBatchJob(cctx, cl, batchReq); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (*TemporalWorkflowTraceCommand) run(*CommandContext, []string) error {
@@ -96,16 +128,26 @@ func (*TemporalWorkflowUpdateCommand) run(*CommandContext, []string) error {
 	return fmt.Errorf("TODO")
 }
 
+type singleOrBatchParams struct {
+	AllowReasonWithWorkflowID bool
+}
+
 func (s *SingleWorkflowOrBatchOptions) workflowExecOrBatch(
 	cctx *CommandContext,
 	namespace string,
 	cl client.Client,
+	options ...func(*singleOrBatchParams),
 ) (*common.WorkflowExecution, *workflowservice.StartBatchOperationRequest, error) {
+	var p singleOrBatchParams
+	for _, opt := range options {
+		opt(&p)
+	}
+
 	// If workflow is set, we return single execution
 	if s.WorkflowId != "" {
 		if s.Query != "" {
 			return nil, nil, fmt.Errorf("cannot set query when workflow ID is set")
-		} else if s.Reason != "" {
+		} else if !p.AllowReasonWithWorkflowID && s.Reason != "" {
 			return nil, nil, fmt.Errorf("cannot set reason when workflow ID is set")
 		} else if s.Yes {
 			return nil, nil, fmt.Errorf("cannot set 'yes' when workflow ID is set")
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -45,7 +45,7 @@ func (c *TemporalWorkflowSignalCommand) run(cctx *CommandContext, args []string)
 		return err
 	}
 
-	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl)
+	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, singleOrBatchOverrides{})
 	if err != nil {
 		return err
 	}
@@ -91,9 +91,9 @@ func (c *TemporalWorkflowTerminateCommand) run(cctx *CommandContext, _ []string)
 	}
 	defer cl.Close()
 
-	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, func(p *singleOrBatchParams) {
+	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, singleOrBatchOverrides{
 		// You're allowed to specify a reason when terminating a workflow
-		p.AllowReasonWithWorkflowID = true
+		AllowReasonWithWorkflowID: true,
 	})
 	if err != nil {
 		return err
@@ -128,7 +128,7 @@ func (*TemporalWorkflowUpdateCommand) run(*CommandContext, []string) error {
 	return fmt.Errorf("TODO")
 }
 
-type singleOrBatchParams struct {
+type singleOrBatchOverrides struct {
 	AllowReasonWithWorkflowID bool
 }
 
@@ -136,18 +136,13 @@ func (s *SingleWorkflowOrBatchOptions) workflowExecOrBatch(
 	cctx *CommandContext,
 	namespace string,
 	cl client.Client,
-	options ...func(*singleOrBatchParams),
+	overrides singleOrBatchOverrides,
 ) (*common.WorkflowExecution, *workflowservice.StartBatchOperationRequest, error) {
-	var p singleOrBatchParams
-	for _, opt := range options {
-		opt(&p)
-	}
-
 	// If workflow is set, we return single execution
 	if s.WorkflowId != "" {
 		if s.Query != "" {
 			return nil, nil, fmt.Errorf("cannot set query when workflow ID is set")
-		} else if !p.AllowReasonWithWorkflowID && s.Reason != "" {
+		} else if s.Reason != "" && !overrides.AllowReasonWithWorkflowID {
 			return nil, nil, fmt.Errorf("cannot set reason when workflow ID is set")
 		} else if s.Yes {
 			return nil, nil, fmt.Errorf("cannot set 'yes' when workflow ID is set")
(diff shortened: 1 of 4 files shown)


temporalcli/commands.workflow.go
```diff
@@ -96,16 +128,26 @@ func (*TemporalWorkflowUpdateCommand) run(*CommandContext, []string) error {
...
 
+type singleOrBatchParams struct {
+	AllowReasonWithWorkflowID bool
+}
+
 func (s *SingleWorkflowOrBatchOptions) workflowExecOrBatch(
 	cctx *CommandContext,
 	namespace string,
 	cl client.Client,
+	options ...func(*singleOrBatchParams),
```
bergundy, line 139: Can you use an options struct instead of this mutator pattern? We typically don't use this pattern in our codebase.

temporalcli/commands.workflow_test.go
```diff
@@ -114,3 +114,100 @@ func (s *SharedServerSuite) testSignalBatchWorkflow(json bool) *CommandResult {
 	}
 	return res
 }
+func (s *SharedServerSuite) TestWorkflow_Terminate_SingleWorkflowSuccess() {
+	// Make workflow wait for termination and then return the context's error
```
bergundy, line 118: The workflow can't wait for termination. It's like kill -9.

temporalcli/commands.workflow_test.go
```diff
@@ -114,3 +114,100 @@ func (s *SharedServerSuite) testSignalBatchWorkflow(json bool) *CommandResult {
...
+		"workflow", "terminate",
+		"--address", s.Address(),
+		"-w", run.GetID(),
+		// Ensure that we can provide a reason
+		"--reason", "terminate-test",
+	)
+	s.NoError(res.Err)
+
+	// Confirm workflow was terminated
+	s.Contains(run.Get(s.Context, nil).Error(), "terminated")
```
bergundy, line 144: Check that the reason is propagated too please.

temporalcli/commands.workflow_test.go
```diff
@@ -114,3 +114,100 @@ func (s *SharedServerSuite) testSignalBatchWorkflow(json bool) *CommandResult {
...
+		args = append(args, "--yes", "-o", "json")
+	} else {
+		s.CommandHarness.Stdin.WriteString("y\n")
+	}
+	res := s.Execute(args...)
+	s.NoError(res.Err)
+
+	// Confirm that all workflows are terminated
+	for _, run := range runs {
+		s.Contains(run.Get(s.Context, nil).Error(), "terminated")
```
bergundy, line 210: Same here, would be nice to verify the reason was propagated.

temporalcli/commands.workflow_test.go
```diff
@@ -114,3 +114,100 @@ func (s *SharedServerSuite) testSignalBatchWorkflow(json bool) *CommandResult {
 	}
 	return res
 }
+func (s *SharedServerSuite) TestWorkflow_Terminate_SingleWorkflowSuccess() {
```
bergundy, line 117: Is reason optional for terminate? It should be and should default to the same reason we have for batch.
Worth testing.

temporalcli/commandsmd/commands.md
```diff
@@ -386,12 +386,29 @@ temporal workflow start \
 
 ### temporal workflow terminate: Terminate Workflow Execution by ID or List Filter.
 
-TODO
+The `temporal workflow terminate` command is used to terminate a [Workflow Execution](/concepts/what-is-a-workflow-execution). Canceling a running Workflow Execution records a `WorkflowExecutionTerminated` event as the closing Event in the workflow's Event History. No further command tasks may be scheduled after running this command.
```
bergundy, line 389:
```suggestion
The `temporal workflow terminate` command is used to terminate a [Workflow Execution](/concepts/what-is-a-workflow-execution). Canceling a running Workflow Execution records a `WorkflowExecutionTerminated` event as the closing Event in the workflow's Event History. Workflow code is oblivious to termination. Use `temporal workflow cancel` if you need to perform cleanup in your workflow.
```

temporalcli/commandsmd/commands.md
```diff
@@ -386,12 +386,29 @@ temporal workflow start \
...
+
+```
+temporal workflow terminate --query=MyQuery
+```
+
+Use the options listed below to change the behavior of this command.
+
+#### Options
+
+Includes options set for [single workflow or batch](#options-set-single-workflow-or-batch)
```
bergundy, line 406: The comment says `--reason` is only for batch (and mentions that it contains the time which is a lie), that should be fixed.

temporalcli/commands.workflow_test.go
tdeebswihart: It looks like our old code doesn't specify a reason for batch termination unless one is provided: https://github.com/temporalio/cli/blob/main/batch/batch_commands.go#L89 and https://github.com/temporalio/cli/blob/main/batch/batch_commands.go#L157

Is there a default reason we'd _like_ to use?

temporalcli/commands.workflow.go
tdeebswihart: We use it often in the server; good to know the SDK/CLI code doesn't. I'd hoped to use this so that we didn't need to change callsites which relied on the default behavior

I'll make that change though

temporalcli/commands.workflow_test.go
tdeebswihart: Done! Thanks for reminding me to do so

temporalcli/commands.workflow.go
bergundy: I wouldn't worry the other callsites in internal code.
Thanks for changing.

temporalcli/commands.workflow_test.go
bergundy: https://github.com/temporalio/cli/blob/8ad817da1f6cde7e48446e5c4e1ab7c7d4e0c4a8/temporalcli/commands.workflow.go#L183

tdeebswihart: Ah, I was looking at the old code :). I'll go add this to the non-batch code

bergundy: Make sure you fix the `--reason` help string too please.

tdeebswihart: Done!

temporalcli/commandsmd/commands.md
```diff
@@ -338,8 +338,7 @@ Includes options set for [payload input](#options-set-for-payload-input).
 * `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
 * `--query`, `-q` (string) - Start a batch to Signal Workflow Executions with given List Filter. Either this or
   Workflow Id must be set.
-* `--reason` (string) - Reason to perform batch. Only allowed if query is present. Defaults to message with user name
-  and time.
+* `--reason` (string) - Reason to perform batch. Only allowed if query is present unless the command specifies otherwise.
```
bergundy, line 341: I think reason needs to be moved out of here and added individually for every command.
I tried phrasing it generically but it comes out unhelpful.

tdeebswihart: Terminate seems to be the only command that needs it (from a brief grep of our protos), so I'm just going to break it out for this single command. All other batch ops can use this

temporalcli/commandsmd/commands.md
```diff
@@ -338,7 +338,7 @@ Includes options set for [payload input](#options-set-for-payload-input).
 * `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
 * `--query`, `-q` (string) - Start a batch to Signal Workflow Executions with given List Filter. Either this or
   Workflow Id must be set.
-* `--reason` (string) - Reason to perform batch. Only allowed if query is present unless the command specifies otherwise. 
+* `--reason` (string) - Reason to perform batch. Only allowed if query is present unless the command specifies otherwise. Defaults to message with the current user's name.
```
dnr, line 341: This still mentions batch?

temporalcli/commandsmd/commands.md
dnr: schedule pause/unpause also takes a reason. It could be a different flag though.

temporalcli/commandsmd/commands.md
tdeebswihart: You're looking at the help for `workflow signal` rather than `workflow terminate`. That command only allows a reason to be specified when invoked in batch mode. If you check the options for `workflow terminate` you'll see I updated it there already:

```
* `--reason` (string) - Reason to terminate this workflow. Defaults to message with the current user's name.
```

temporalcli/commandsmd/commands.md
```diff
@@ -386,12 +385,34 @@ temporal workflow start \
...
+```
+
+Use the options listed below to change the behavior of this command.
+
+#### Options
+
+* `--workflow-id`, `-w` (string) - Workflow Id. Either this or query must be set.
+* `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
+* `--query`, `-q` (string) - Start a batch to Signal Workflow Executions with given List Filter. Either this or Workflow Id must be set.
+* `--reason` (string) - Reason to terminate this workflow. Defaults to message with the current user's name.
```
bergundy, line 409: Applies to both batch and single workflow:

```suggestion
* `--reason` (string) - Reason for termination. Defaults to message with the current user's name.
```

temporalcli/commandsmd/commands.md
```diff
@@ -386,12 +385,34 @@ temporal workflow start \
...
+temporal workflow terminate --query=MyQuery
+```
+
+Use the options listed below to change the behavior of this command.
+
+#### Options
+
+* `--workflow-id`, `-w` (string) - Workflow Id. Either this or query must be set.
+* `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
+* `--query`, `-q` (string) - Start a batch to Signal Workflow Executions with given List Filter. Either this or Workflow Id must be set.
```
bergundy, line 408:
```suggestion
* `--query`, `-q` (string) - Start a batch to Terminate Workflow Executions with given List Filter. Either this or Workflow Id must be set.
```

bergundy (approved):

temporalcli/commandsmd/commands.md
```diff
@@ -386,12 +385,34 @@ temporal workflow start \
...
+
+Use the options listed below to change the behavior of this command.
+
+#### Options
+
+* `--workflow-id`, `-w` (string) - Workflow Id. Either this or query must be set.
+* `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
+* `--query`, `-q` (string) - Start a batch to terminate Workflow Executions with given List Filter. Either this or Workflow Id must be set.
+* `--reason` (string) - Reason for termination. Defaults to message with the current user's name.
+* `--yes`, `-y` (bool) - Confirm prompt to perform batch. Only allowed if query is present.
```
cretz, line 410:
```suggestion
Includes options set for [single workflow or batch](#options-set-for-single-workflow-or-batch).
```

Or are you making this a completely separate options set just because of the English description of reason?

temporalcli/commands.workflow.go
```diff
@@ -96,16 +141,29 @@ func (*TemporalWorkflowUpdateCommand) run(*CommandContext, []string) error {
...
+
+type singleOrBatchOverrides struct {
+	AllowReasonWithWorkflowID bool
+}
+
 func (s *SingleWorkflowOrBatchOptions) workflowExecOrBatch(
 	cctx *CommandContext,
 	namespace string,
 	cl client.Client,
+	overrides singleOrBatchOverrides,
```
cretz, line 160: Probably can toss all of this in a single struct instead of keeping overrides separate, but no biggie

cretz (commented):
Nice! Just a couple of notes (though I know it's already merged, just stuff to discuss)

temporalcli/commandsmd/commands.md
bergundy: Note that `#options-set-for-single-workflow-or-batch` mentions signal and reason is applied only to batch.

tdeebswihart: Here's the discussion Roey and I had about this @cretz: https://github.com/temporalio/cli/pull/436#discussion_r1480582673

cretz: We can make docs more generic if we need to, but I have no problem copy/pasting just for docs-only reasons if we have to like here.

Issue #66, opened 2022-12-31
[Feature Request] Provide easy way to hide output


lorensr:
### Is your feature request related to a problem? Please describe.
In order to be able to continue using my terminal, I have to add `> /dev/null 2>&1 &` like:

`temporal server start-dev > /dev/null 2>&1 &`

### Describe the solution you'd like
`temporal server start-dev -d`

like `docker-compose`'s -d flag:

```
  -d, --detach                    Detached mode: Run containers in the background
```

A Mac-only alternative solution is #10 

### Additional context
Previous discussion: https://github.com/temporalio/temporalite/issues/139

lorensr:
Another option: hide stdout unless they pass `-v / --verbose` or `--debug`

bergundy:
I don't think we need a detached mode, users can run the command in the background if they intend on running it the way you do.
For complete daeomonization support, we'd need PID files, a way to get the status of the "current" running server, a command to stop the server, etc..
I don't see a huge benefit in this when it's easy enough to run the server with:

```
temporal server start-dev &> /dev/null & disown
```

With docker you don't really have a choice but to provide a `-d` flag but for a standard applications that runs on the developer's machine there's far less value for built-in daemonization.

> Another option: hide stdout unless they pass `-v / --verbose` or `--debug`

There's already a configurable log level, we should make sure the log levels are tuned and possibly default to `warn`.

lorensr:
Roey requested I don't tell people to do:

`temporal server start-dev &> /dev/null & disown`

So I'm back onto improving the default output experience ☺️

@feedmeapples would it be easy to:

1. change default log level to warn
2. hide `Starting UI server...` and Echo output, including `⇨ http server started on 127.0.0.1:8233`
3. add this output:

```
> dist/temporal server start-dev
Temporal Server running at: localhost:7233
Web UI running at: http://localhost:8233
```

(would also be nice if `localhost:7233` and `http://localhost:8233` were green)

### current output

```
> dist/temporal server start-dev
{"level":"info","ts":"2023-03-24T16:05:53.195-0400","msg":"Successfully saved cluster metadata.","component":"metadata-initializer","cluster-name":"active","logging-call-at":"fx.go:700"}
{"level":"info","ts":"2023-03-24T16:05:53.196-0400","msg":"Use rpc address 127.0.0.1:7233 for cluster active.","component":"metadata-initializer","logging-call-at":"fx.go:840"}
[...]

PR #441, opened 2024-02-10
CLI Refresh: Output improvements on start-dev

cretz:
## What was changed

* Split CLI log level from server log level (server "info" is much noisier than CLI) and defaulted it to warn
* Put simple non-log text output for started server and UI server as requested at #66
* Defaulted IP to `localhost` and translate internally to 127.0.0.1 for use but keep as localhost for text output
* Once https://github.com/temporalio/ui/pull/1858 is available and potential future server shutdown changes, many of the spurios output will be removed

Now when you run `temporal server start-dev`, this is the output:

```
Temporal server is running at: localhost:7233
Web UI is running at: http://localhost:8233
```

(many suggestions came from @lorensr)

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -208,6 +208,7 @@ type TemporalServerStartDevCommand struct {
 	SqlitePragma       []string
 	DynamicConfigValue []string
 	LogConfig          bool
+	LogLevelServer     StringEnum
 }
 
 func NewTemporalServerStartDevCommand(cctx *CommandContext, parent *TemporalServerCommand) *TemporalServerStartDevCommand {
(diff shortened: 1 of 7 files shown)


temporalcli/commands.gen.go
```diff
@@ -229,13 +230,15 @@ func NewTemporalServerStartDevCommand(cctx *CommandContext, parent *TemporalServ
...
-	s.Command.Flags().StringVar(&s.Ip, "ip", "127.0.0.1", "IP address to bind the frontend service to.")
+	s.Command.Flags().StringVar(&s.Ip, "ip", "localhost", "IP address to bind the frontend service to.")
 	s.Command.Flags().StringVar(&s.UiIp, "ui-ip", "", "IP address to bind the Web UI to. Default is same as --ip.")
 	s.Command.Flags().StringVar(&s.UiAssetPath, "ui-asset-path", "", "UI custom assets path.")
 	s.Command.Flags().StringVar(&s.UiCodecEndpoint, "ui-codec-endpoint", "", "UI remote codec HTTP endpoint.")
 	s.Command.Flags().StringArrayVar(&s.SqlitePragma, "sqlite-pragma", nil, "Specify SQLite pragma statements in pragma=value format.")
 	s.Command.Flags().StringArrayVar(&s.DynamicConfigValue, "dynamic-config-value", nil, "Dynamic config value, as KEY=JSON_VALUE (string values need quotes).")
 	s.Command.Flags().BoolVar(&s.LogConfig, "log-config", false, "Log the server config being used in stderr.")
+	s.LogLevelServer = NewStringEnum([]string{"debug", "info", "warn", "error", "off"}, "warn")
+	s.Command.Flags().Var(&s.LogLevelServer, "log-level-server", "Log level for the server only.")
```
dandavison, line 241: Not related to this PR but I think it would be nice to show the user the available alternative enum values

```
      --log-level-server string            Log level for the server only. (default "warn")
```

dandavison: Does `--log-level off` have precedent elsewhere in Temporal/Go? If so then fine; if not, then I'll also throw in `--log-level never` as a candidate, since it's sort of consistent gramatically: `debug, info, warn, error, never`.

cretz: > Not related to this PR but I think it would be nice to show the user the available alternative enum values

Agreed. We currently have a general concept of trailing sentences being removed from the English part, and Cobra doesn't support any kind of enum listing. So we can, at code writing time, append the "Accepted values: <values...>" for all enums.

> Does --log-level off have precedent elsewhere in Temporal/Go?

No, only here I think, so I'd support changing this to `never` for for sure.

I can open a PR for both of these shortly...

dandavison:
This is a huge improvement.

PR #442, opened 2024-02-12
Change log-level from "off" to "never" and show enums in options help

cretz:
## What was changed

Change log-level from "off" to "never" and show enums in options help.


temporalcli/commandsmd/code.go
```diff
@@ -319,12 +319,26 @@ func (c *CommandOption) writeFlagBuilding(selfVar, flagVar string, w *codeWriter
...
+	if len(c.EnumValues) > 0 {
+		desc += " Accepted values: "
+		for i, enumVal := range c.EnumValues {
+			if i > 0 {
+				desc += ", "
+			}
+			desc += enumVal
+		}
+		desc += "."
+	}
```
dandavison, line 334:
```suggestion
	if len(c.EnumValues) > 0 {
		c.Desc += fmt.Sprintf(" Accepted values: %s.", strings.Join(c.EnumValues, ", "))
	}
```

dandavison (approved):
One code suggestion but otherwise LGTM

temporalcli/commandsmd/code.go
cretz: Yeah, I did this originally because I had "and" in there and it didn't look right, will change.

PR #443, opened 2024-02-12
CLI Refresh: Implement `workflow reset`

tdeebswihart:
## What was changed

I've implemented the `workflow reset` command and tests for all its major functionality.

Batch reset functionality will come in a second PR after this one


temporalcli/commands.workflow.go
```diff
@@ -252,3 +329,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+		}
+		for _, e := range resp.GetHistory().GetEvents() {
+			if e.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
+				workflowTaskEventID = e.GetEventId()
+				break
+			} else if e.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_TASK_SCHEDULED {
+				// if there is no task completed event, set it to first scheduled event + 1
+				workflowTaskEventID = e.GetEventId() + 1
+			}
+		}
```
ast2023, line 372: extract this into a function

temporalcli/commands.workflow.go
```diff
@@ -252,3 +329,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+		if len(resp.NextPageToken) != 0 {
+			req.NextPageToken = resp.NextPageToken
+		} else {
+			break
+		}
+	}
+	if workflowTaskEventID == 0 {
+		return "", 0, errors.New("unable to find any scheduled or completed task")
+	}
```
ast2023, line 421: I think this part is the same as in the previous function.

temporalcli/commands.workflow.go
```diff
@@ -53,8 +70,68 @@ func (*TemporalWorkflowQueryCommand) run(*CommandContext, []string) error {
...
+	if c.Type.Value != "" {
+		resetBaseRunID, eventID, err = getResetEventIDByType(cctx, c.Type.Value, c.Parent.Namespace, c.WorkflowId, c.RunId, wfsvc)
+		if err != nil {
+			return fmt.Errorf("getting reset event ID by type failed: %w", err)
+		}
+	}
+	username := "<unknown-user>"
+	if u, err := user.Current(); err != nil && u.Username != "" {
+		username = u.Username
+	}
```
ast2023, line 101: do not we have a function to establish username?

temporalcli/commands.workflow.go
```diff
@@ -53,8 +70,68 @@ func (*TemporalWorkflowQueryCommand) run(*CommandContext, []string) error {
...
+	defer cl.Close()
+
+	conn, err := c.Parent.ClientOptions.dialGRPC(cctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	wfsvc := workflowservice.NewWorkflowServiceClient(conn)
```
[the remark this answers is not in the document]
tdeebswihart: I wasn't aware that existed, nice!

temporalcli/commands.workflow.go
tdeebswihart: One does not exist and I didn't extract these three lines into one, no

temporalcli/commands.workflow.go
```diff
@@ -252,3 +329,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+			} else if e.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_TASK_SCHEDULED {
+				// if there is no task completed event, set it to first scheduled event + 1
+				workflowTaskEventID = e.GetEventId() + 1
+			}
+		}
+		if len(resp.NextPageToken) != 0 {
+			req.NextPageToken = resp.NextPageToken
+		} else {
+			break
+		}
```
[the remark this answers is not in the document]
tdeebswihart: If Go had a `do { } while (COND)` loop I would, but the NextPageToken starts with length 0 so we'd skip the loop entirely.

temporalcli/commands.workflow.go
ast2023: you have exactly same code in defaultReason. Extaract and use

temporalcli/commands.workflow.go
tdeebswihart: It is except one iterates forwards through history and the other in reverse order, see `GetWorkflowExecutionHistory` vs `GetWorkflowExecutionHistoryReverse`.

The logic in the `EVENT_TYPE_WORKFLOW_TASK_SCHEDULED` case also differs

temporalcli/commands.workflow.go
```diff
@@ -309,3 +377,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+		for _, e := range resp.GetHistory().GetEvents() {
+			if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
+				workflowTaskEventID = e.GetEventId()
+				break
+			} else if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_SCHEDULED {
+				// if there is no task completed event, set it to first scheduled event + 1
+				workflowTaskEventID = e.GetEventId() + 1
+			}
+		}
```
ast2023, line 420: You should extract a function which returns workflowTaskEventID

temporalcli/commands.workflow.go
```diff
@@ -309,3 +377,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+				if workflowTaskEventID == 0 {
+					workflowTaskEventID = e.GetEventId() + 1
+				}
+			}
+		}
+		if len(resp.NextPageToken) != 0 {
+			req.NextPageToken = resp.NextPageToken
+		} else {
+			break
+		}
```
ast2023, line 465: this looks similar to the previous function. There is obviously duplication.

temporalcli/commands.workflow.go
```diff
@@ -309,3 +377,147 @@ func startBatchJob(cctx *CommandContext, cl client.Client, req *workflowservice.
...
+			if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
+				workflowTaskCompletedID = e.GetEventId()
+			}
+		}
+		if len(resp.NextPageToken) != 0 {
+			req.NextPageToken = resp.NextPageToken
+		} else {
+			break
+		}
+	}
```
ast2023, line 518: this is also very similar to the previous function

temporalcli/commands.workflow.go
tdeebswihart: You must have missed my response to your earlier comment. There are two differences between the two functions:

1. The order of iteration and request/response types. This one iterates over history in order (`GetWorkflowExecutionHistory`) whereas the previous function iterates over it in reverse order (`GetWorkflowExecutionHistoryReverse`)
2. The logic to handle the case where event type is `EVENT_TYPE_WORKFLOW_TASK_SCHEDULED` differs because of that

There's very little duplicated between these two functions except the scaffolding of "iterate over history in some order with the same stopping condition"

temporalcli/commands.workflow.go
tdeebswihart: I'm not abstracting a single for loop into a function as everything outside of it differs.

temporalcli/commands.workflow.go
tdeebswihart: See my comments on the other variants of this logic

temporalcli/commands.workflow.go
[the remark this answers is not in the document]
tdeebswihart:
```
for true; ok = len(resp.NextPageToken) != 0; ok {
    // logic
}

or some variant thereof is not any more readable than the explicit break maybe 10 lines below the for in this function.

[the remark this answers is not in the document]
Comment (tdeebswihart) [reply temporalcli/commands.workflow.go:425]: I'll change it for the sake of not wanting to waste time arguing

File: temporalcli/client.go [hunk 9]
```diff
@@ -142,6 +147,27 @@ func payloadCodecInterceptor(namespace, codecEndpoint, codecAuth string) (grpc.U
 	)
 }
 
+func errorInterceptor() grpc.UnaryClientInterceptor {
```
cretz, line 150: Unused

temporalcli/client.go
```diff
@@ -11,13 +11,18 @@ import (
...
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/log"
+
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 )
 
+const localHostPort = "127.0.0.1:7233"
```
cretz, line 24: Unused

temporalcli/client.go
```diff
@@ -142,6 +147,27 @@ func payloadCodecInterceptor(namespace, codecEndpoint, codecAuth string) (grpc.U
...
 
+func errorInterceptor() grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		err = serviceerror.FromStatus(status.Convert(err))
+		return err
+	}
+}
+
+func headersProviderInterceptor(provider stringMapHeadersProvider) grpc.UnaryClientInterceptor {
```
cretz, line 158: Unused

temporalcli/commands.workflow.go
cretz, line 1: Don't have to, but if it makes more sense to you, feel free to break this into `commands.workflow_reset.go` or similar if you think there is a lot of reset-specific code cluttering other code

temporalcli/commands.workflow.go
```diff
@@ -110,8 +125,57 @@ func (c *TemporalWorkflowQueryCommand) run(cctx *CommandContext, args []string)
...
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	resetBaseRunID := c.RunId
+	eventID := int64(c.EventId)
+	if c.Type.Value != "" {
+		resetBaseRunID, eventID, err = getResetEventIDByType(cctx, c.Type.Value, c.Parent.Namespace, c.WorkflowId, c.RunId, cl.WorkflowService())
```
cretz, line 141:
```suggestion
		resetBaseRunID, eventID, err = c.getResetEventIDByType(cctx, cl)
```

Nit, not required, just can save params and reduce the package-level pollution by using a reasonable receiver here

temporalcli/commands.workflow.go
```diff
@@ -16,6 +18,19 @@ import (
 	"go.temporal.io/sdk/client"
 )
 
+var (
+	resetTypesMap = map[string]interface{}{
```
cretz, line 22: Unused

temporalcli/commands.workflow.go
```diff
@@ -16,6 +18,19 @@ import (
...
+		"FirstWorkflowTask":  "",
+		"LastWorkflowTask":   "",
+		"LastContinuedAsNew": "",
+	}
+	resetReapplyTypesMap = map[string]interface{}{
```
cretz, line 27: Unused

temporalcli/commands.workflow.go
```diff
@@ -110,8 +125,57 @@ func (c *TemporalWorkflowQueryCommand) run(cctx *CommandContext, args []string)
...
+
+	resp, err := cl.ResetWorkflowExecution(cctx, &workflowservice.ResetWorkflowExecutionRequest{
+		Namespace: c.Parent.Namespace,
+		WorkflowExecution: &common.WorkflowExecution{
+			WorkflowId: c.WorkflowId,
+			RunId:      resetBaseRunID,
+		},
+		Reason:                    fmt.Sprintf("%s: %s", username(), c.Reason),
+		WorkflowTaskFinishEventId: eventID,
+		RequestId:                 uuid.NewString(),
```
cretz, line 166: Done by default by the SDK (but fine here too)

temporalcli/commands.workflow.go
```diff
@@ -110,8 +125,57 @@ func (c *TemporalWorkflowQueryCommand) run(cctx *CommandContext, args []string)
...
+	reapplyType := enums.RESET_REAPPLY_TYPE_SIGNAL
+	if c.ReapplyType.Value != "All" {
+		reapplyType, err = enums.ResetReapplyTypeFromString(c.ReapplyType.Value)
+		if err != nil {
+			return err
+		}
+	}
+
+	if eventID > 0 {
+		cctx.Printer.Printlnf("Resetting workflow %s to event ID %d", c.WorkflowId, eventID)
```
cretz, line 155: What does it mean for event ID to be 0? Should we always say something is being done to text-based output users instead of the possibility of no output?

temporalcli/commands_test.go
```diff
@@ -303,6 +308,15 @@ func StartDevServer(t *testing.T, options DevServerOptions) *DevServer {
...
 	}()
 
+	// Dial workflowservice client
+	d.GRPCConn, err = grpc.Dial(d.Options.ClientOptions.HostPort, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer func() {
+		if !success {
+			d.GRPCConn.Close()
+		}
+	}()
```
cretz, line 318: Can we remove this?

temporalcli/commands.workflow_test.go
```diff
@@ -446,3 +452,195 @@ func (s *SharedServerSuite) testQueryWorkflow(json bool) {
...
+	require.NoError(s.T(), err)
+	var ignored any
+	s.NoError(run.Get(s.Context, &ignored))
+	s.Equal(1, oneExecutions)
+	s.Equal(1, twoExecutions)
+
+	// We want to reset to the last WFTCompleted event before the second activity, so we
+	// need to search the history for it. I could just pick the event ID, but I don't want
+	// this test to break if new event types are added in the future
+	wfsvc := workflowservice.NewWorkflowServiceClient(s.GRPCConn)
```
cretz, line 601: Can you use the regular client already on the suite?

temporalcli/client.go
tdeebswihart: I'd wondered what all would shake out from deleting `dialGRPC`. Thanks for catching these

temporalcli/commands.workflow.go
tdeebswihart: That's probably for the best; there's enough here with the helpers that I'll separate it out

temporalcli/commands_test.go
tdeebswihart: Yes we can

temporalcli/commands.workflow_test.go
tdeebswihart: Yep. I forgot about this when I made the change to the actual logic

temporalcli/commands_test.go
tdeebswihart: Note: I'm also going to change these to use `t.Cleanup()` in a followup PR so we don't need to track success

temporalcli/commands.workflow.go
tdeebswihart: I had added that to debug some tests, but it shouldn't be possible for the eventID to be zero.

The old CLI didn't announce what event ID we were resetting to but it may be useful to know what eventID the CLI has decided to use when --type is specified

temporalcli/commands.workflow.go
tdeebswihart: Done

temporalcli/commands.workflow_reset.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if workflowTaskEventID == 0 {
+		return "", 0, errors.New("unable to find any scheduled or completed task")
+	}
+	return
+}
+
+// Returns id of the first workflow task completed event or if it doesn't exist then id of the event after task scheduled event.
+func getFirstWorkflowTaskEventID(ctx context.Context, namespace, wid, rid string, wfsvc workflowservice.WorkflowServiceClient) (resetBaseRunID string, workflowTaskEventID int64, err error) {
+	resetBaseRunID = rid
```
ast2023, line 125: I do not see you ever change `resetBaseRunID` or `rid` in the function.

temporalcli/commands.workflow_reset.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	for more := true; more; more = len(req.NextPageToken) != 0 {
+		resp, err := wfsvc.GetWorkflowExecutionHistory(ctx, &req)
+		if err != nil {
+			return "", 0, fmt.Errorf("failed to get workflow execution history: %w", err)
+		}
+		for _, e := range resp.GetHistory().GetEvents() {
+			if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
+				workflowTaskEventID = e.GetEventId()
+				return resetBaseRunID, workflowTaskEventID, nil
```
ast2023, line 143: is it possible that `e.GetEventId()` is 0 and `e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED`? May be we can just break here, just like in the GetLastWorkflowTaskEventID? Then these loops will be exactly the same?

temporalcli/commands.workflow_reset.go
tdeebswihart: Yep. All three variants (these are lightly-cleaned up version of the original logic) use the same API even though only `getLastContinueAsNewID` changes it. We could change two of the three calls in `getResetEventIDByType` to do, say, `return rid, getLastWorkflowTaskEventID(...)` but I didn't

temporalcli/commands.workflow_reset.go
tdeebswihart: As far as I'm aware event IDs are _always_ greater than zero, and even if they began at zero the first WorkflowTaskCompleted event must be preceded by three other events so still could not be zero:
1. WorkflowExecutionStarted              
2. WorkflowTaskScheduled                 
3. WorkflowTaskStarted 
4. **WorkflowTaskCompleted**

tdeebswihart: The logic here is "set workflowTaskEventID to right after the first WorkflowTaskScheduledEvent _but_ override it if we encounter WorkflowTaskCompleted"

temporalcli/commands.workflow_reset.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		for _, e := range resp.GetHistory().GetEvents() {
+			if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
+				workflowTaskEventID = e.GetEventId()
+				break
+			} else if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_SCHEDULED {
+				// if there is no task completed event, set it to first scheduled event + 1
+				workflowTaskEventID = e.GetEventId() + 1
+			}
+		}
```
ast2023, line 114: This `for` loop is premature optimization. You could define a function: 
```
func getFirstEventOfType(events, eventType) event{
   for _, e := range events {
       if e.GetEventType() == eventType {
          return e
      }
  }
}
```
and then write code as follow:
```
    events = resp.GetHistory().getEvents()
    event = getFirstEventOfType(events, enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED)
    if event == nil {
        event = getFirstEventOfType(events, enums.EVENT_TYPE_WORKFLOW_TASK_SCHEDULED)
    } 
```
and in the `getLastContinuedAs new:
```
   event = getFirstEventOfType(reverse(events), enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED)
```

tdeebswihart: `reverse(events)` is not the same as iterating in reverse order as this is a paginated query.

temporalcli/commands.workflow_reset.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
+)
+
+var (
+	resetTypesMap = map[string]interface{}{
```
cretz, line 16: Unused

cretz (approved):

cretz:
Overall LGTM!

temporalcli/commands.workflow_reset.go
tdeebswihart: Looks like I managed to remove that as well, excellent. Removed in the next PR

PR #444, opened 2024-02-12
Don't run tests in verbose mode in CI

tdeebswihart:
## What was changed

CI will now run tests without `-v`.

## Why?
When you run in verbose mode you need to dig through _all_ test outputs and logging to find even a single test failure. In https://github.com/temporalio/cli/actions/runs/7877883209/job/21494958729?pr=443 we have 1000 lines of output and only 20 lines were relevant to the actual failing test.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
           go-version: '1.21'
 
       - name: Test
-        run: go test -v ./...
+        run: go test ./...
 
       - name: Regen code, confirm unchanged
         if: ${{ matrix.checkGenCodeTarget }}


.github/workflows/ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           go-version: '1.21'
 
       - name: Test
-        run: go test -v ./...
+        run: go test ./...
```
cretz, line 31: I don't think that you have to dig through lots of output should be that big of a deal. Granted Go makes it hard to see when test output begins, but we put in work to separate the pieces. Just because the action you're using as an impetus for this PR didn't require verbose output to see the problem doesn't mean other people's test failures don't. I don't think the logs are that egregious.

cretz: Update: To clarify, since verbose is automatically applied on the test that fails, I can be ok with this change. I would like to hear others' opinion here real quick before approving though.

tdeebswihart: I'm only proposing this for that exact reason: failing tests will always be dumped verbosely :)

cretz (approved):
Looks like nobody came with another opinion, marking approved

PR #445, opened 2024-02-12
Implement activity commands

dandavison:
## What was changed
Implement `temporal activity fail` and `temporal activity complete`

## How was this tested
- PR includes end-to-end tests
- I modified [`samples-python:hello/hello_activity.py`](https://github.com/temporalio/samples-python/blob/main/hello/hello_activity.py) to sleep for a long time in the activity, issued CLI commands (examples below), and confirmed results in web UI.

<details>

```diff
commit 5519021cac3a1d0b9818d903882e0e053ad3df8b (HEAD -> update-reapply)
Author: Dan Davison <dan.davison@temporal.io>
Date:   Mon Feb 12 18:02:34 2024 -0500

    Test cli activity commands

diff --git a/hello/hello_activity.py b/hello/hello_activity.py
index 6615b55..a8a8e54 100644
--- a/hello/hello_activity.py
+++ b/hello/hello_activity.py
@@ -21,6 +21,7 @@ class ComposeGreetingInput:
 @activity.defn
 async def compose_greeting(input: ComposeGreetingInput) -> str:
     activity.logger.info("Running activity with parameter %s" % input)
+    await asyncio.sleep(0xFFFFFFFF)
     return f"{input.greeting}, {input.name}!"
 
 
@@ -33,7 +34,8 @@ class GreetingWorkflow:
         return await workflow.execute_activity(
             compose_greeting,
             ComposeGreetingInput("Hello", name),
-            start_to_close_timeout=timedelta(seconds=10),
+            start_to_close_timeout=timedelta(seconds=0xFFFFFFFF),
+            activity_id="MyActivityId",
         )
```

</details>

## Example commands tested
```
temporal-new activity complete --activity-id MyActivityId --workflow-id MyWorkflowId --result '{"resultkey": "resultval"}' --identity 'MyIdentity' --run-id e592b7a1-a15d-4de1-83b4-32ec070706fa
temporal-new activity fail --activity-id MyActivityId --workflow-id MyWorkflowId --detail '{"detailkey": "detailval"}' --run-id d9049fc2-ad6c-4f3e-9f04-52b287a7e913 --reason 'top-level-reason' --identity 'MyIdentity'
```

### DIFF
--- a/temporalcli/commands.activity.go
+++ b/temporalcli/commands.activity.go
@@ -1,0 +1,73 @@
+package temporalcli
+
+import (
+	"fmt"
+
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/api/failure/v1"
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+func (c *TemporalActivityCompleteCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	metadata := map[string][]byte{"encoding": []byte("json/plain")}
+	resultPayloads, err := CreatePayloads([][]byte{[]byte(c.Result)}, metadata, false)
+	if err != nil {
+		return err
+	}
+
+	_, err = cl.WorkflowService().RespondActivityTaskCompletedById(cctx, &workflowservice.RespondActivityTaskCompletedByIdRequest{
+		Namespace:  c.Parent.Namespace,
+		WorkflowId: c.WorkflowId,
+		RunId:      c.RunId,
+		ActivityId: c.ActivityId,
+		Result:     resultPayloads,
+		Identity:   c.Identity,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to complete Activity: %w", err)
+	}
+	return nil
+}
+
+func (c *TemporalActivityFailCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	metadata := map[string][]byte{"encoding": []byte("json/plain")}
+	var detailPayloads *common.Payloads
+	detailPayloads = nil
+	if len(c.Detail) > 0 {
+		detailPayloads, err = CreatePayloads([][]byte{[]byte(c.Detail)}, metadata, false)
+		if err != nil {
+			return err
+		}
+	}
+	_, err = cl.WorkflowService().RespondActivityTaskFailedById(cctx, &workflowservice.RespondActivityTaskFailedByIdRequest{
+		Namespace:  c.Parent.Namespace,
+		WorkflowId: c.WorkflowId,
+		RunId:      c.RunId,
+		ActivityId: c.ActivityId,
+		Failure: &failure.Failure{
+			Message: c.Reason,
+			Source:  "CLI",
+			FailureInfo: &failure.Failure_ApplicationFailureInfo{ApplicationFailureInfo: &failure.ApplicationFailureInfo{
+				NonRetryable: true,
+				Details:      detailPayloads,
+			}},
+		},
+		Identity: c.Identity,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to fail Activity: %w", err)
+	}
+	return nil
+}
--- a/temporalcli/commands.activity.go
+++ b/temporalcli/commands.activity.go
@@ -42,10 +42,9 @@ func (c *TemporalActivityFailCommand) run(cctx *CommandContext, args []string) e
 	}
 	defer cl.Close()
 
-	metadata := map[string][]byte{"encoding": []byte("json/plain")}
 	var detailPayloads *common.Payloads
-	detailPayloads = nil
 	if len(c.Detail) > 0 {
+		metadata := map[string][]byte{"encoding": []byte("json/plain")}
 		detailPayloads, err = CreatePayloads([][]byte{[]byte(c.Detail)}, metadata, false)
 		if err != nil {
 			return err
--- a/temporalcli/commandsmd/commands.md
+++ b/temporalcli/commandsmd/commands.md
@@ -57,6 +57,43 @@ This document has a specific structure used by a parser. Here are the rules:
 * `--color` (string-enum) - Set coloring. Options: always, never, auto. Default: auto.
 * `--no-json-shorthand-payloads` (bool) - Always all payloads as raw payloads even if they are JSON.
 
+### temporal activity: Complete or fail an activity.
+
+#### Options
+
+Includes options set for [client](#options-set-for-client).
+
+
+### temporal activity complete: Complete an activity.
+
+Complete an Activity Execution.
+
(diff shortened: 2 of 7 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -57,6 +57,43 @@ This document has a specific structure used by a parser. Here are the rules:
 * `--color` (string-enum) - Set coloring. Options: always, never, auto. Default: auto.
 * `--no-json-shorthand-payloads` (bool) - Always all payloads as raw payloads even if they are JSON.
 
+### temporal activity: Complete or fail an activity.
+
```
cretz, line 61:
```suggestion

Activity commands operate on [Activity Executions](/concepts/what-is-an-activity-execution).

Activity commands follow this syntax:
`temporal activity [command] [command options]`
```

Just stole it from old CLI, but I would say it's a bit trivial (we hope to come back and clean these)

temporalcli/commandsmd/commands.md
```diff
@@ -57,6 +57,43 @@ This document has a specific structure used by a parser. Here are the rules:
...
+### temporal activity complete: Complete an activity.
+
+Complete an Activity Execution.
+
+`temporal activity complete --activity-id=MyActivityId --workflow-id=MyWorkflowId --result='{"MyResultKey": "MyResultVal"}'`
+
+#### Options
+
+* `--activity-id` (string) - The Activity to be completed. Required.
+* `--identity` (string) - Identity of operator.
```
cretz, line 76: Previous CLI required this, but I am ok not requiring it

temporalcli/commandsmd/commands.md
```diff
@@ -57,6 +57,43 @@ This document has a specific structure used by a parser. Here are the rules:
...
+* `--result` (string) - The result with which to complete the Activity (JSON). Required.
```
cretz, line 77: This works for me to meet parity. There are a few things we may want in the future to be as robust as `--input` for workflows:

* `--result` as a `string[]` to accept multiple results
* `--result-file` to accept files instead of strings
* `--result-meta` to support payload metadata
* `--result-base64` bool to support base64 binary strings

I wonder if we should create an issue here (same for `--detail` on failure)

temporalcli/payload.go
cretz, line 1: Not sure this deserves its own file but meh

temporalcli/payload.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"go.temporal.io/api/common/v1"
+)
+
+func CreatePayloads(data [][]byte, metadata map[string][]byte, isBase64 bool) (*common.Payloads, error) {
```
cretz, line 12:
```suggestion
func createPayloads(data [][]byte, metadata map[string][]byte, isBase64 bool) (*common.Payloads, error) {
```

Probably don't need to export (though it doesn't really matter in the context of an application vs a library so there's no rules here, so nothing required to change)

temporalcli/commandsmd/commands.md
cretz, line 76: It was a bit inconsistent for us to have our CLI accept identity here but not on other commands like start workflow. I wonder if we should accept `--identity` higher up.

temporalcli/commands.activity.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	defer cl.Close()
+
+	metadata := map[string][]byte{"encoding": []byte("json/plain")}
+	resultPayloads, err := CreatePayloads([][]byte{[]byte(c.Result)}, metadata, false)
+	if err != nil {
+		return err
+	}
+
+	_, err = cl.WorkflowService().RespondActivityTaskCompletedById(cctx, &workflowservice.RespondActivityTaskCompletedByIdRequest{
```
cretz, line 24: While this is the approach of the old SDK, `cl.CompleteActivityByID` would have worked here too (you'd wrap the payload in `rawValue`). Arguably it'd be cleaner/clearer to use the high-level SDK client, but don't need to change now, but we might consider using the high-level API in the future.

temporalcli/commands.activity.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	metadata := map[string][]byte{"encoding": []byte("json/plain")}
+	var detailPayloads *common.Payloads
+	detailPayloads = nil
+	if len(c.Detail) > 0 {
+		detailPayloads, err = CreatePayloads([][]byte{[]byte(c.Detail)}, metadata, false)
```
cretz, line 49:
```suggestion
	var detailPayloads *common.Payloads
	if len(c.Detail) > 0 {
		metadata := map[string][]byte{"encoding": []byte("json/plain")}
		detailPayloads, err = CreatePayloads([][]byte{[]byte(c.Detail)}, metadata, false)
```

Little cleanup

cretz (approved):
Some notes for now and future, but nothing blocking. The existing CLI sure was inconsistent/unclean with these two commands.

temporalcli/commandsmd/commands.md
dandavison: Yes, I'm not in love with that text from the old CLI. I think the help output is pretty clear as-is:

```
$ temporal activity -h
Complete or fail an activity.

Usage:
  temporal activity [command]

Available Commands:
  complete    Complete an activity.
  fail        Fail an activity.
```

and these Markdown links `[Activity Executions](/concepts/what-is-an-activity-execution)` don't link to a valid target (unless I'm missing something?)

temporalcli/commandsmd/commands.md
dandavison: https://github.com/temporalio/cli/issues/446

temporalcli/payload.go
dandavison: I wasn't sure how people would like the code arranged. Where would you put it out of interest?

temporalcli/commandsmd/commands.md
dandavison: Agreed; I'd like consistency across all protos that have an `identity` field. I was just going for parity here.

temporalcli/commands.activity.go
dandavison: OK I'm happy to make that change and it would be easy, however, it looks like it doesn't support `identity` so I'm going to leave it for now.

temporalcli/payload.go
dandavison: OK thanks, my Go style is a work-in-progress... when I uppercased it I was thinking that I was working in a separate "util" package from the main commands, but as you point out that's not the case.

temporalcli/commands.activity.go
dandavison: Thanks!

dandavison:
Thanks @cretz! Merged to keep the velocity up but there are some threads there we can follow-up in future PRs.

temporalcli/commandsmd/commands.md
cretz: Not missing anything, I too am not a fan, but I was just leaving many docs things as is to come back and be cleaned up later. Can just leave off too.

temporalcli/payload.go
cretz: Probably client, but it doesn't matter. As the utils grow we can move/consolidate as needed

PR #447, opened 2024-02-13
Emit JSONL from `workflow list --output json`

dandavison:
## What was changed
Change `workflow list --output json` to emit JSONL format.

## Why?
Currently it emits one multiline JSON object per workflow, so the result is (in the case of >1 workflow) neither valid JSON nor valid JSONL. Most tools will not be able to parse this (although as it happens, `jq` does parse this).


cretz:

PR #450, opened 2024-02-13
Override SDK-set headers for client name and others

### LINKED ISSUE #440: CLI Refresh: Need client-name and client-version headers as CLI with version

Need gRPC interceptor to set these two headers

cretz:
## What was changed

Add `client-name`, `client-version`, `supported-server-versions`, and `caller-type` headers on outbound client calls in CLI refresh code. Previous CLI used outgoing gRPC metadata on the context, but my research has shown we overwrite that anyways. So up until now we may not have been setting these headers properly (making CLI appear as Go SDK).

## Checklist

1. Closes #440

### DIFF
--- a/temporalcli/client.go
+++ b/temporalcli/client.go
@@ -15,6 +15,7 @@ import (
 	"go.temporal.io/sdk/converter"
 	"go.temporal.io/sdk/log"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error) {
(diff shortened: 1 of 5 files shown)


temporalcli/client.go
```diff
@@ -92,6 +101,25 @@ func (c *ClientOptions) tlsConfig() (*tls.Config, error) {
...
+	// The SDK sets some values on the outgoing metadata that we can't override
+	// via normal headers, so we have to replace directly on the metadata
+	md, _ := metadata.FromOutgoingContext(ctx)
+	if md == nil {
+		md = metadata.MD{}
+	}
+	md.Set("client-name", "temporal-cli")
+	md.Set("client-version", Version)
+	md.Set("supported-server-versions", ">=1.0.0 <2.0.0")
+	md.Set("caller-type", "operator")
```
Sushisource, line 118: What is this one for?

Sushisource (approved):

temporalcli/client.go
bergundy: It gives calls made by the CLI higher priority over application code.
This helps troubleshoot issues when you're getting rate limited for example.

bergundy (approved):

temporalcli/commands.workflow_exec_test.go
```diff
@@ -149,6 +153,52 @@ func (s *SharedServerSuite) TestWorkflow_Execute_SimpleFailure() {
...
+		) error {
+			lastHeadersLock.Lock()
+			lastHeadersClient, _ = metadata.FromOutgoingContext(ctx)
+			lastHeadersLock.Unlock()
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}),
+	)
+
+	// Capture from server
+	// TODO(cretz): Pending fix on server for gRPC interceptors
```
cretz, line 176: https://github.com/temporalio/temporal/pull/5414 is the fix for those curious

PR #458, opened 2024-02-15
Implement `workflow show`

Sushisource:
## What was changed
Implement the `workflow show` command

## Why?
Part of cli rewrite

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.workflow_exec.go
+++ b/temporalcli/commands.workflow_exec.go
@@ -391,19 +391,10 @@ type structuredHistoryEvent struct {
 var structuredHistoryEventType = reflect.TypeOf(structuredHistoryEvent{})
 
 func (s *structuredHistoryIter) Next() (any, error) {
-	// Load iter
-	if s.iter == nil {
-		s.iter = s.client.GetWorkflowHistory(
-			s.ctx, s.workflowID, s.runID, s.follow, enums.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
-	}
-	if !s.iter.HasNext() {
-		return nil, nil
-	}
-	event, err := s.iter.Next()
+	event, err := s.NextRawEvent()
 	if err != nil {
 		return nil, err
 	}
-
 	// Build data
 	data := structuredHistoryEvent{
 		ID:   event.EventId,
@@ -426,4 +417,20 @@ func (s *structuredHistoryIter) Next() (any, error) {
 		s.iter = nil
 	}
 	return data, nil
+}
+
+func (s *structuredHistoryIter) NextRawEvent() (*history.HistoryEvent, error) {
+	// Load iter
+	if s.iter == nil {
+		s.iter = s.client.GetWorkflowHistory(
+			s.ctx, s.workflowID, s.runID, s.follow, enums.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+	}
+	if !s.iter.HasNext() {
+		return nil, nil
+	}
+	event, err := s.iter.Next()
+	if err != nil {
+		return nil, err
+	}
+	return event, nil
 }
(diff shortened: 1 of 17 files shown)


temporalcli/commands.workflow_view.go
```diff
@@ -265,6 +265,51 @@ func (*TemporalWorkflowCountCommand) run(*CommandContext, []string) error {
...
+		workflowID:     c.WorkflowId,
+		runID:          c.RunId,
+		includeDetails: true,
+		follow:         c.Follow,
+		color:          !cctx.JSONOutput && !color.NoColor,
+	}
+	if !cctx.JSONOutput {
+		cctx.Printer.Println(color.MagentaString("Progress:"))
+		// TODO: Do we need to do the "time elapsed" thing? Also needs some redoing to print
+		//   result at the end, since final event is not accessible from here.
```
Sushisource, line 289: Looking for any thoughts on this

temporalcli/commands.workflow_exec.go
```diff
@@ -404,7 +411,13 @@ func (s *structuredHistoryIter) Next() (any, error) {
 	data := structuredHistoryEvent{
 		ID:   event.EventId,
 		Time: event.EventTime.AsTime().Format(time.RFC3339),
-		Type: coloredEventType(event.EventType),
+	}
+	// TODO: This is pretty wrong - the printer should be able to determine if color should be used
+	//   or not, but that would mean adding typed color information to the structured data.
```
Sushisource, line 416: This is hacky, but fixing it is sort of nonobvious. We would need to stop using strings and start using something better-typed for color, so that the printer can then print what it gets from `Next` smartly.

Not sure it's worth it.

cretz: Color is turned on/off at a global level which is how the https://pkg.go.dev/github.com/fatih/color library works (I hate globals as much as the next guy, but it's how they did it which is ok here). So calling the color calls is a no-op when it is disabled. You don't have to worry about it as a developer, use the color calls directly and just know they are not applied when color is disabled,

temporalcli/commands.workflow_view.go
cretz: ON `workflow execute` I think, there was this "Progress" with a counting number of seconds which I removed, and if that's what this was here, I think we can remove it here too.

temporalcli/commands.workflow_view.go
```diff
@@ -265,6 +265,51 @@ func (*TemporalWorkflowCountCommand) run(*CommandContext, []string) error {
...
+		if err := iter.print(cctx.Printer); err != nil {
+			return fmt.Errorf("displaying history failed: %w", err)
+		}
+	} else {
```
cretz, line 293: It is important to note that IIUC `workflow show` is the primary way people dump workflow history for use in replayers. So for JSON, we need to dump exact history format (i.e. create the `history.History` proto object with its events and send that to `PrintStructured`, which is close to what's here but has to be the full thing)

temporalcli/commands.workflow_exec.go
Sushisource: Then it's not working right. I was consistently getting color escape codes inside the JSON for even types

temporalcli/commands.workflow_view.go
Sushisource: This is exactly how the old CLI did it FYI. But, making it be a literal history proto makes more sense anyway.

cretz: > But, making it be a literal history proto makes more sense anyway.

They used that in `printReplayableHistory` which makes sense (it's simple code just for JSON), but for the non-JSON, yeah the lazy stuff for text is a bit more difficult. Feel free to steal anything from `workflow execute` which does similar.

temporalcli/commandsmd/commands.md
```diff
@@ -438,8 +438,8 @@ Use the options listed below to change the command's behavior.
 
 #### Options
 
-* `--reset-points` (bool) - Only show auto-reset points.
```
cretz, line 441: Looks like old CLI accepted this but never used it, :+1: removing

cretz (approved):
LGTM

PR #459, opened 2024-02-15
Implement `schedule` commands

dnr:
## What was changed
Implement schedule commands for new CLI.

## Why?
Feature parity.

## Checklist

1. Closes 

2. How was this tested:

new tests

3. Any docs updates needed?

- The format of `temporal schedule describe -o json` changed to the `--raw` format of the previous cli.


temporalcli/commandsmd/commands.md
```diff
@@ -377,6 +377,163 @@ Search Attribute commands enable operations for the creation, listing, and remov
...
+
+Any combination of `--calendar`, `--interval`, and `--cron` is supported.
+Actions will be executed at any time specified in the Schedule.
+
+#### Options set for schedule configuration:
+
+* `--calendar` (string[]) - Calendar specification in JSON, e.g. `{"dayOfWeek":"Fri","hour":"17","minute":"5"}`.
+* `--catchup-window` (duration) - Maximum allowed catch-up time if server is down.
+* `--cron` (string[]) - Calendar spec in cron string format, e.g. `3 11 * * Fri`.
+* `--end-time` (timestamp) - Overall schedule end time.
```
cretz, line 447: Nice, can you document this new field type in the big xml/markdown comment near the top of this file?

temporalcli/commands.schedule.go
```diff
@@ -0,0 +1,484 @@
...
+	IntervalSpecs    []any `cli:",cardOmitEmpty"`
+	WorkflowId       any
+	WorkflowType     any
+	Paused           any
+	Notes            any `cli:",cardOmitEmpty"`
+	NextRunTime      any
+	LastRunTime      any
+	RunningWorkflows []any
+	SearchAttributes any `cli:",cardOmitEmpty"`
+	Memo             any `cli:",cardOmitEmpty"`
```
cretz, line 30: Not required, but I figure at least some of these could have stronger field types

temporalcli/commands.schedule.go
```diff
@@ -0,0 +1,484 @@
...
+		out.LastRunTime = last.ScheduleTime
+	}
+	return out
+}
+
+func specToPrintable(out *printableSchedule, spec *client.ScheduleSpec) {
+	for _, cal := range spec.Calendars {
+		// TODO: This is not quite right for either json or text:
+		// For json: this doesn't come out as proper protojson, the field names are
+		// day_of_month instead of dayOfMonth.
```
cretz, line 84: This is the cost of mixing proto and non-proto JSON but there's not an east solution. What we do in some other commands is separate the JSON representation structs from the text ones, and for JSON, make this be something like `[]json.Message` instead where we pre-JSON it. It became too confusing for complicated models to try to satisfy JSON and non-JSON needs with the same model (but if you must reuse models, consider more explicit JSON models then that meet the JSON field name needs).

cretz (commented):
Looks good. Didn't pour over every output detail or anything.

temporalcli/commands.schedule.go
Sushisource: Big +1 here

temporalcli/commands.schedule.go
```diff
@@ -0,0 +1,484 @@
...
+	NextRunTime      any
+	LastRunTime      any
+	RunningWorkflows []any
+	SearchAttributes any `cli:",cardOmitEmpty"`
+	Memo             any `cli:",cardOmitEmpty"`
+}
+
+func describeResultToPrintable(id string, desc *client.ScheduleDescription) *printableSchedule {
+	// TODO: should we include any other fields here, e.g. jitter, time zone, start/end time
+	out := &printableSchedule{
```
Sushisource, line 35: I think so. I'd expect describing to return everything available.

temporalcli/commands.schedule.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	return strings.Join(fields, " "), nil
+}
+
+func toIntervalSpec(str string) (client.ScheduleIntervalSpec, error) {
+	var spec client.ScheduleIntervalSpec
+	var err error
+	parts := strings.Split(str, "/")
+	if len(parts) > 2 {
+		return spec, errors.New("invalid interval string")
```
Sushisource, line 160: Would be good to say why here

Sushisource:
Also appears this will close https://github.com/temporalio/cli/issues/433 - if that's true can you update the description to close that when this is merged?

dnr:
> Also appears this will close #433 - if that's true can you update the description to close that when this is merged?

Not really, it just prints memo/search attrs in list, but not update them.

temporalcli/commands_test.go
```diff
@@ -216,12 +218,23 @@ func (s *SharedServerSuite) TearDownTest() {
...
 	}
 	s.CommandHarness = nil
+	s.testVars = nil
 }
 
 func (s *SharedServerSuite) T() *testing.T                 { return s.Suite.T() }
 func (s *SharedServerSuite) SetT(t *testing.T)             { s.Suite.SetT(t) }
 func (s *SharedServerSuite) SetS(suite suite.TestingSuite) { s.Suite.SetS(suite) }
 
+func (s *SharedServerSuite) testVar(key string) string {
```
cretz, line 228: I don't understand this suite-level helper, can't a test make whatever variables it needs?

temporalcli/commands.schedule_test.go
```diff
@@ -0,0 +1,361 @@
...
+	"fmt"
+	"io"
+	"regexp"
+	"time"
+
+	"github.com/temporalio/cli/temporalcli"
+	"go.temporal.io/sdk/workflow"
+)
+
+func (s *SharedServerSuite) schedId() string   { return s.testVar("sched") }
```
cretz, line 16:
```suggestion
func (s *SharedServerSuite) schedID() string   { return s.testVar("sched") }
```

But I don't understand this compared to just making vars in your tests as needed (or returning an ID from `createSchedule` or whatever)

temporalcli/commands.schedule_test.go
```diff
@@ -0,0 +1,361 @@
...
+		"-s", s.schedId(),
+	)
+	s.Error(res.Err)
+}
+
+func (s *SharedServerSuite) TestSchedule_Describe() {
+	res := s.createSchedule("--interval", "2s")
+	s.NoError(res.Err)
+
+	time.Sleep(4 * time.Second) // run at least once
```
cretz, line 86: Can we avoid sleeps in tests? Can we do some kind of manual trigger/backfill here?

temporalcli/commands.schedule_test.go
```diff
@@ -0,0 +1,361 @@
...
+
+	s.Eventually(func() bool {
+		res = s.Execute(
+			"schedule", "list",
+			"--address", s.Address(),
+		)
+		s.NoError(res.Err)
+		out := res.Stdout.String()
+		return AssertContainsOnSameLine(out, s.schedId(), "DevWorkflow", "false") == nil
+	}, 10*time.Second, time.Second)
```
cretz, line 152: Can we increase the interval to like 100ms or something? Same for all other `Eventually` uses in this PR. Ideally no test even takes near a second (though I know it happens sometimes)

cretz (commented):
Looks good, just want to bring that test time down. And it looks like the CLI already has some flaky tests in CI, ug.

temporalcli/commands.schedule.go
```diff
@@ -0,0 +1,597 @@
...
+	action := &client.ScheduleWorkflowAction{
+		ID:                       opts.ID,
+		Workflow:                 sw.Type,
+		TaskQueue:                opts.TaskQueue,
+		WorkflowExecutionTimeout: opts.WorkflowExecutionTimeout,
+		WorkflowRunTimeout:       opts.WorkflowRunTimeout,
+		WorkflowTaskTimeout:      opts.WorkflowTaskTimeout,
+		// RetryPolicy not supported yet
+		// TODO: buildStartOptions doesn't fill in TypedSearchAttributes, this doesn't work yet:
+		TypedSearchAttributes: opts.TypedSearchAttributes,
```
dnr, line 267: @cretz @Quinn-With-Two-Ns What do you think is the best way to handle this right now?

The situation appears to be: `buildStartOptions` builds search attributes in the form `map[string]any` (very easy to do from json) and puts them in `StartWorkflowOptions.SearchAttributes`. This is deprecated, but it still works for starting workflows.

For `ScheduleWorkflowAction` though, apparently untyped support is gone completely? The comment on `UntypedSearchAttributes` says `This should never be used for create.` Even if I could use it, it's a `map[string]*commonpb.Payload` which means I'd have to convert to `Payload`s (a little more work but not a huge deal).

To set `TypedSearchAttributes`, though, it looks like I'd have to do a lot of work. Basically it would have to look a lot like `convertToTypedSearchAttributes` from `internal/internal_search_attributes.go`, except with reflection?

I understand the goal of typed search attributes is to make them hard to use in a generic way, but the cli kinda needs to use them in a generic way 🫤  (So does "modify schedule search attributes" btw, except that's from workflow context so the code has to be different again.)

temporalcli/commands.schedule_test.go
cretz: This is great, thanks!

temporalcli/commands_test.go
cretz: I don't _hate_ it, it can stay, though I don't see ever using it myself (I think passing typesafe state around via explicitly named things to be much clearer and a more natural programming thing to do)

temporalcli/commands.schedule.go
cretz: Ug, yeah, we've hit this a few times now with generic tools. If the `UntypedSearchAttributes` does work for now I think it's ok to use here and just assume that warning is for users only. The problem is technically a search attribute key for setting is supposed to be defined by its name _and_ type, but we have no reasonable interface for that in CLI.

Note the starting of workflows sets untyped search attributes too, we were just forced to remove the either-or option on schedules because this same type is used for update and we couldn't tell which was the intended field on update (it was a long discussion internally before we made this backwards compatibility break and we documented it in release notes).

dnr: I confirmed `UntypedSearchAttributes` in the start workflow action works for create schedule so I used that for now (sorry Quinn) and added a test.

temporalcli/commands_test.go
dnr: It is type-safe, it just happens to be all strings (we use strings as identifiers everywhere so it's useful even limited to strings). And I think it's fine for tests to use a more implicit style than I would want in non-test code.

But in this case you're right it's easy to do it the straightfoward way for these tests. That pattern has more value with more complex tests.

cretz (approved):
If there are obvious flag incompatibilities with today's CLI schedule command, can you mention those in the PR description? We're going to go across the PRs and collect them when we make our first release.

dnr:
> If there are obvious flag incompatibilities with today's CLI schedule command, can you mention those in the PR description? We're going to go across the PRs and collect them when we make our first release.

I'm not aware of any big incompatibilities, though there may be some small ones. The non-raw describe json output is different, but that's why `--raw` was advised before. Now it's just `-o json`

cretz:
:+1: No concerns about output compatibility

PR #461, opened 2024-02-15
Use testing.T.Cleanup() in new cli test framework

tdeebswihart:
## What was changed
All manual calls to stop shared resources were replaced by calling `thing.Stop` as a cleanup function

## Why?
We don't need to track this ourselves, we can let the testing library do it for us

### DIFF
--- a/temporalcli/commands_test.go
+++ b/temporalcli/commands_test.go
@@ -170,23 +170,9 @@ type SharedServerSuite struct {
 func (s *SharedServerSuite) SetupSuite() {
 	s.DevServer = StartDevServer(s.Suite.T(), DevServerOptions{})
 	// Stop server if we fail later
-	success := false
-	defer func() {
-		if !success {
-			s.Server.Stop()
-		}
-	}()
+	s.T().Cleanup(s.Server.Stop)
+
 	s.Worker = s.DevServer.StartDevWorker(s.Suite.T(), DevWorkerOptions{})
-	success = true
-}
-
-func (s *SharedServerSuite) TearDownSuite() {
-	s.Stop()
-}
-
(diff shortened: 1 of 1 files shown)


temporalcli/commands_test.go
```diff
@@ -317,15 +293,9 @@ func StartDevServer(t *testing.T, options DevServerOptions) *DevServer {
 		require.NoError(t, err)
 	}
 
-	success = true
 	return d
 }
 
-func (d *DevServer) Stop() {
```
cretz, line 324: If I choose to invoke `StartDevServer` I should be able to choose when to stop it. I don't think we should assume lifetimes of things we didn't start.

temporalcli/commands_test.go
```diff
@@ -411,13 +381,10 @@ func (d *DevServer) StartDevWorker(t *testing.T, options DevWorkerOptions) *DevW
 	w.Worker.RegisterActivity(ops.DevActivity)
 	// Start worker or fail
 	require.NoError(t, w.Worker.Start(), "failed starting worker")
+	t.Cleanup(w.Worker.Stop)
 	return w
 }
 
-func (d *DevWorker) Stop() {
```
cretz, line 417: Same here, if I as the caller choose to start the worker, I should be able to choose to stop it

temporalcli/commands_test.go
Quinn-With-Two-Ns: Yes, I need this functionality for tests I am adding

tdeebswihart: Then I'll close this. Until your work it wasn't necessary :)

tdeebswihart:
Closing as we actually will want to close these manually in future tests

PR #462, opened 2024-02-15
CLI Refresh: `workflow update` command

pdoerner:
## What was changed

Add support for `workflow update` command. The functionality should be unchanged but the output may be different.

### DIFF
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -5,11 +5,12 @@ import (
 	"os/user"
 
 	"github.com/google/uuid"
-	"github.com/temporalio/cli/temporalcli/internal/printer"
 	"go.temporal.io/api/batch/v1"
 	"go.temporal.io/api/common/v1"
 	"go.temporal.io/api/workflowservice/v1"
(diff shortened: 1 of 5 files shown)


temporalcli/commands.workflow.go
```diff
@@ -223,8 +224,45 @@ func (*TemporalWorkflowTraceCommand) run(*CommandContext, []string) error {
...
+		FirstExecutionRunID: c.FirstExecutionRunId,
+		Args:                input,
+	}
+
+	updateHandle, err := cl.UpdateWorkflowWithOptions(cctx, request)
+	if err != nil {
+		return fmt.Errorf("unable to update workflow: %w", err)
+	}
+
+	var valuePtr interface{}
```
cretz, line 253: I think elsewhere I did:

```suggestion
	var valuePtr json.RawMessage
```

Though I think it's basically the same thing here (it'll just be a `map[string]any` for an object which is turned right back to JSON) so no worries.

cretz (approved):
Looks great, merge at will

Issue #448, opened 2024-02-13
[CLI Refresh] Output valid JSON and/or JSONL for `workflow list`


dandavison:
Commands such as `workflow list` conceptually output a sequence of objects. Currently `cli-rewrite` JSON output is formatting this as a sequence of multiline JSON objects. This means that the overall result (when there is >1 workflow) is neither valid JSON nor valid [JSONL](https://jsonlines.org/).

A natural requirement is that the CLI should produce either valid JSON or valid JSONL. Possible designs satisfying that requirement:

### A

- `--output json` always produces valid JSON (i.e. we add `[`, commas, and `]` to the `workflow list` output).
- We will also support `--output jsonl` meaning: convert all top-level arrays to newline-separated objects; otherwise do not emit any newlines in JSON.

  Thus `--output jsonl` would apply consistently to all commands (but mainly be useful for commands such as `workflow list` and `env get` that produce top-level arrays, although people could conceivably use it to produce compact output for other commands).

### B

- `--output json` produces valid JSON everywhere except for certain commands (perhaps only `workflow list`) where it produces valid JSONL.

  The only change required here would be eliminating the newlines within `workflow list` items. A downside here is that it will no longer be possible to produce easily human-readable output from `workflow list` without piping into something like `jq`.

cretz:
`batch list` seems to apply here too

cretz:
`cluster list` seems to apply here too (though not really, it should probably be a fixed list since it'll never be big)

cretz:
`operator namespace list` seems to apply here too (though not really, it should probably be a fixed list since it'll never be big)

cretz:
Ok, after discussion, it looks like we're going with Option A here

cretz:
Option A complete

PR #465, opened 2024-02-16
Add more operator cluster commands

Quinn-With-Two-Ns:
Add `operator cluster list`, `operator cluster upsert` and `operator cluster remove`.

To write the tests I had to expose a bunch of cluster configs in the Test server and make sure the test could actually read them to use and verify the information.

### DIFF
--- a/temporalcli/commands.operator_cluster.go
+++ b/temporalcli/commands.operator_cluster.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"go.temporal.io/api/operatorservice/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
 )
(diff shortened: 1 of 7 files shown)


temporalcli/commands.operator_cluster.go
```diff
@@ -86,3 +87,102 @@ func (c *TemporalOperatorClusterDescribeCommand) run(cctx *CommandContext, args
...
+func (c *TemporalOperatorClusterListCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	var nextPageToken []byte
+	var execsProcessed int
+	for pageIndex := 0; ; pageIndex++ {
```
cretz, line 100: This is one of those cases I wonder if we never expect a large list and therefore should just collect the list and dump all at once. Regardless, I added to #448 to revisit.

cretz (approved):
Ug, sorry you had to dig up this stuff. At least you know more about cluster stuff now, heh.

Issue #468, opened 2024-02-17
[CLI Refresh] Implement `temporal task-queue` build-ID commands (`update-` and `get-`)


antlai-temporal:
Implement task-queue commands in the new CLI:
* get-build-ids
* get-build-ids-reachability
* update-build-ids

cretz:
Closing this issue as completed (for whatever reason, PRs are not closing related issues)

PR #469, opened 2024-02-17
[CLI Refresh] Add build id commands

antlai-temporal:
## What was changed
Implements

./temporal task-queue get-build-id-reachability
./temporal task-queue get-build-ids
./temporal task-queue update-build-ids add-new-compatible
./temporal task-queue update-build-ids add-new-default
./temporal task-queue update-build-ids promote-id-in-set
./temporal task-queue update-build-ids promote-set

## Checklist

1. Closes 
#468


antlai-temporal:
wrong target brainch

temporalcli/commands.taskqueue_get_build_id.go
```diff
@@ -0,0 +1,126 @@
...
+		MaxSets:   c.MaxSets,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to get task queue build ids: %w", err)
+	}
+
+	type rowtype struct {
+		BuildIds      []string
+		DefaultForSet string
+		IsDefaultSet  bool
```
cretz, line 30: You'll probably want to provide `"json:buildIds"` type of struct tags for these since we usually want our JSON as camelCase (even if old CLI wasn't as kind/consistent).

temporalcli/commands.taskqueue_get_build_id.go
```diff
@@ -0,0 +1,126 @@
...
+			BuildIds:      s.BuildIDs,
+			IsDefaultSet:  ix == len(sets.Sets)-1,
+			DefaultForSet: s.BuildIDs[len(s.BuildIDs)-1],
+		}
+		items = append(items, row)
+	}
+
+	if cctx.JSONOutput {
+		return cctx.Printer.PrintStructured(items, printer.StructuredOptions{})
+	}
```
cretz, line 44: Can just get rid of this. `Println` doesn't print when JSON set, and `Table` is ignored when JSON set, so the code below will do what you want. So basically `PrintStructured` is abstracted for reuse for just this kind of use case.

temporalcli/commands.taskqueue_update_build_ids.go
```diff
@@ -0,0 +1,80 @@
+package temporalcli
+
+import (
+	"fmt"
+
+	"go.temporal.io/sdk/client"
+)
+
+func updateBuildIds(cctx *CommandContext, parent *TemporalTaskQueueCommand, options *client.UpdateWorkerBuildIdCompatibilityOptions) error {
+	cl, err := parent.ClientOptions.dialClient(cctx)
```
cretz, line 10:
```suggestion
func (c *TemporalTaskQueueUpdateBuildIdsCommand) updateBuildIds(cctx *CommandContext, options *client.UpdateWorkerBuildIdCompatibilityOptions) error {
	cl, err := c.Parent.ClientOptions.dialClient(cctx)
```

Can make the common piece the receiver if you wanted to just `c.Parent.updateBuildIds(cctx, options)` in other commands below without having a top-level function. But not required or anything.

cretz (commented):
Looks great, only minor comments, and only the JSON field-name one I think is really required.

temporalcli/commands.taskqueue_get_build_id.go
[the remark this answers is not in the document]
antlai-temporal: I was wondering about that, but the commands are at different levels, keeping all the sub commands of `update_build_id` in a separate file makes it a bit easier to find...

temporalcli/commands.taskqueue_get_build_id.go
antlai-temporal: Done

temporalcli/commands.taskqueue_get_build_id.go
```diff
@@ -0,0 +1,126 @@
...
+
+	cctx.Printer.Println(color.MagentaString("Version Sets:"))
+	return cctx.Printer.PrintStructured(items, printer.StructuredOptions{Table: &printer.TableOptions{}})
+}
+
+// Missing in the SDK?
```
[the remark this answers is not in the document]
antlai-temporal: This code is going away with the versioning rewrite, but I will make sure we have something like that after the rewrite...

temporalcli/commands.taskqueue_get_build_id.go
antlai-temporal: Great, much cleaner thanks!

temporalcli/commands.taskqueue_update_build_ids.go
antlai-temporal: Didn't realize I had a base command build by the parser... That's great thanks!

temporalcli/commands.taskqueue_get_build_id.go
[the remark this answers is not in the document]
cretz: Sure, though we don't over-optimize for discoverability or we'd put every command in its own file (or force every sub command to be in a file with its adjacent ones). But this is fine. You may want to separate the tests too. For discoverability, often you prefer to test `foo.go` in `foo_test.go`.

cretz (approved):
One minor thing about test file name at https://github.com/temporalio/cli/pull/469#discussion_r1502944716, nothing big, LGTM

PR #470, opened 2024-02-18
Add operator search-attribute commands

yiminc:
## What was changed
Add operator search-attribute commands

## Why?
Add commands to operate search attributes

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+# Binaries for programs and plugins
+/temporal
+/temporal.exe
+
+# Used by IDE
+/.idea
+/.vscode
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,10 @@ Uses normal `go test`, e.g.:
 
 See other tests for how to leverage things like the command harness and dev server suite.
 
+Example to run a single test case:
+
+    go test ./... -run TestSharedServerSuite/TestOperator_SearchAttribute
+
 ## Adding/updating commands
 
 First, update [commands.md](temporalcli/commandsmd/commands.md) following the rules in that file. Then to regenerate the
--- a/temporalcli/commands.operator_search_attribute.go
+++ b/temporalcli/commands.operator_search_attribute.go
@@ -1,0 +1,141 @@
+package temporalcli
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/operatorservice/v1"
+)
+
+func (c *TemporalOperatorSearchAttributeCreateCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	// Name and Type are required parameters, and there must be the same number of them.
+
+	listReq := &operatorservice.ListSearchAttributesRequest{
+		Namespace: c.Parent.Parent.Namespace,
+	}
+	existingSearchAttributes, err := cl.OperatorService().ListSearchAttributes(cctx, listReq)
+	if err != nil {
+		return fmt.Errorf("unable to get existing search attributes: %w", err)
+	}
+
+	searchAttributes := make(map[string]enums.IndexedValueType, len(c.Type))
+	for i, saType := range c.Type {
+		saName := c.Name[i]
+		typeInt, err := searchAttributeTypeStringToEnum(saType)
+		if err != nil {
+			return fmt.Errorf("unable to parse search attribute type %s: %w", saType, err)
+		}
+		existingSearchAttributeType, searchAttributeExists := existingSearchAttributes.CustomAttributes[saName]
+		if searchAttributeExists && existingSearchAttributeType != typeInt {
+			return fmt.Errorf("search attribute %s already exists and has different type %s", saName, existingSearchAttributeType.String())
+		}
+		searchAttributes[saName] = typeInt
+	}
+
+	request := &operatorservice.AddSearchAttributesRequest{
+		SearchAttributes: searchAttributes,
+		Namespace:        c.Parent.Parent.Namespace,
+	}
+
+	_, err = cl.OperatorService().AddSearchAttributes(cctx, request)
+	if err != nil {
+		return fmt.Errorf("unable to add search attributes: %w", err)
+	}
+	cctx.Printer.Println(color.GreenString("Search attributes have been added"))
+	return nil
+}
+
+func searchAttributeTypeStringToEnum(search string) (enums.IndexedValueType, error) {
+	for k, v := range enums.IndexedValueType_shorthandValue {
+		if strings.EqualFold(search, k) {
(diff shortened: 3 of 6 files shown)


temporalcli/commands.operator_search_attribute.go
```diff
@@ -0,0 +1,167 @@
...
+
+	_, err = cl.OperatorService().AddSearchAttributes(cctx, request)
+	if err != nil {
+		return fmt.Errorf("unable to add search attributes: %w", err)
+	}
+	cctx.Printer.Println(color.GreenString("Search attributes have been added"))
+	return nil
+}
+
+func stringToEnum(search string) (enums.IndexedValueType, error) {
```
cretz, line 59: If you must handle case insensitive, you can loop over the `enums.IndexedValueType_shorthandValue` and check the key to match this instead of a switch here. But if you do want to keep this helper, can you name it something more specific since it's a package level function visible for all writing code in this package?

temporalcli/commands.operator_search_attribute.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+func (c *TemporalOperatorSearchAttributeCreateCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	if len(c.Name) != len(c.Type) {
```
cretz, line 20: Not required, but may want a check that neither is zero (i.e. that at least one SA was provided)

.gitignore
cretz, line 1: Can we leave all of the extra gitignore stuff out? A version of this file with just the following may make sense if we're concerned about committing binaries:

```
/temporal
/temporal.exe
```

CONTRIBUTING.md
```diff
@@ -14,6 +14,10 @@ Uses normal `go test`, e.g.:
 
 See other tests for how to leverage things like the command harness and dev server suite.
 
+Example to run a single test case:
+
+    go test -v github.com/temporalio/cli/temporalcli -run SharedServerSuite -testify.m TestOperator_SearchAttribute
```
cretz, line 19: I think something like this is simpler
```
go test ./... -run TestSharedServerSuite/TestOperator_SearchAttribute
```

temporalcli/commandsmd/commands.md
```diff
@@ -184,6 +184,32 @@ Cluster commands follow this syntax: `temporal operator cluster [command] [comma
...
+
+`temporal operator search-attribute list` displays a list of all Search Attributes.
+
+### temporal operator search-attribute remove: Removes custom search attribute metadata only
+
+`temporal operator search-attribute remove` command removes custom Search Attribute metadata.
+
+#### Options
+
+* `--name` (string[]) - Search Attribute name.
```
cretz, line 210:
```suggestion
* `--name` (string[]) - Search Attribute name. Required.
```

temporalcli/commands.operator_search_attribute.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (c *TemporalOperatorSearchAttributeRemoveCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	yes, err := cctx.promptYes(
+		fmt.Sprintf("You are about to remove search attributes %v? y/N", c.Name), c.Yes)
```
cretz, line 88:
```suggestion
		fmt.Sprintf("You are about to remove search attribute %q? y/N", c.Name), c.Yes)
```

(pedantic, sorry, just was confusing w/ the plural)

temporalcli/commands.operator_search_attribute.go
```diff
@@ -0,0 +1,167 @@
...
+		SearchAttributes: names,
+		Namespace:        namespace,
+	}
+
+	_, err = cl.OperatorService().RemoveSearchAttributes(cctx, request)
+	if err != nil {
+		return fmt.Errorf("unable to remove search attributes: %w", err)
+	}
+
+	// response contains nothing
```
cretz, line 112: Would be ok w/ a simple println telling users it was removed similar to one we have for when one is added, but up to you.

temporalcli/commands.operator_search_attribute.go
```diff
@@ -0,0 +1,167 @@
...
+	var saSchemaMapping []saStorageSchema
+	for saName, saSchema := range resp.StorageSchema {
+		saSchemaMapping = append(saSchemaMapping, saStorageSchema{Name: saName, Schema: saSchema})
+	}
+
+	cctx.Printer.PrintStructured(sas, printer.StructuredOptions{Table: &printer.TableOptions{}})
+	if len(saSchemaMapping) > 0 {
+		cctx.Printer.Println(color.MagentaString("Storage Schema Mapping:"))
+		return cctx.Printer.PrintStructured(saSchemaMapping, printer.StructuredOptions{Table: &printer.TableOptions{}})
+	}
```
cretz, line 165: Is this a new feature we're adding to CLI as part of this PR?

cretz (commented):
LGTM, just little things

temporalcli/commands.operator_search_attribute.go
yiminc: will use more specific name

.gitignore
yiminc: There are files generated for IDE that we don't want to see them checked in.

cretz: We haven't really been bloating gitignore with all possible IDE files like done in older repos (e.g. https://github.com/temporalio/sdk-python/blob/main/.gitignore and https://github.com/temporalio/sdk-dotnet/blob/main/.gitignore). Can we just limit this to build artifacts and not editor ones?

temporalcli/commands.operator_search_attribute.go
cretz: It's not future proof for future enum values and it's more lines of code than if you just did a simple map iteration, but up to you I guess.

temporalcli/commands.operator_search_attribute.go
yiminc: Didn't realize this was not exposed in existing CLI. Remove it for now.

CONTRIBUTING.md
```diff
@@ -14,6 +14,10 @@ Uses normal `go test`, e.g.:
 
 See other tests for how to leverage things like the command harness and dev server suite.
 
+Example to run a single test case:
+
+    go test ./... -run SharedServerSuite/TestOperator_SearchAttribute
```
cretz, line 19:
```suggestion
    go test ./... -run TestSharedServerSuite/TestOperator_SearchAttribute
```

cretz (commented):
LGTM, one minor note about test name on `go test` example (didn't test with unqualified name)

cretz (approved):

CONTRIBUTING.md
yiminc: weird, both works.

temporalcli/commands.operator_search_attribute.go
yiminc: updated to use enums.IndexedValueType_shorthandValue

PR #473, opened 2024-02-21
Add support for batch resets.

ast2023:
## What was changed
Adding support for batch reset commands.

## Why?
Supported by older tools.

## Checklist
How was this tested:
implemented automated feature test.

### DIFF
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -7,13 +7,14 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/google/uuid"
-	"github.com/temporalio/cli/temporalcli/internal/printer"
 	"go.temporal.io/api/batch/v1"
 	"go.temporal.io/api/common/v1"
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/query/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
+
+	"github.com/temporalio/cli/temporalcli/internal/printer"
 )
 
 func (c *TemporalWorkflowCancelCommand) run(cctx *CommandContext, args []string) error {
@@ -56,10 +57,6 @@ func (*TemporalWorkflowDeleteCommand) run(*CommandContext, []string) error {
 func (c *TemporalWorkflowQueryCommand) run(cctx *CommandContext, args []string) error {
 	return queryHelper(cctx, c.Parent, c.PayloadInputOptions,
 		c.Type, c.RejectCondition, c.WorkflowReferenceOptions)
-}
-
-func (*TemporalWorkflowResetBatchCommand) run(*CommandContext, []string) error {
-	return fmt.Errorf("TODO")
 }
 
 func (c *TemporalWorkflowSignalCommand) run(cctx *CommandContext, args []string) error {
--- a/temporalcli/commands.workflow_reset.go
+++ b/temporalcli/commands.workflow_reset.go
@@ -5,23 +5,68 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"github.com/google/uuid"
+	"go.temporal.io/api/batch/v1"
 	"go.temporal.io/api/common/v1"
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
+
+	"github.com/temporalio/cli/temporalcli/internal/printer"
 )
 
 func (c *TemporalWorkflowResetCommand) run(cctx *CommandContext, _ []string) error {
-	if c.Type.Value == "" && c.EventId <= 0 {
-		return errors.New("must specify either valid event id or reset type")
+	validateArguments, doReset := c.getResetOperations()
+	if err := validateArguments(); err != nil {
+		return err
 	}
 	cl, err := c.Parent.ClientOptions.dialClient(cctx)
 	if err != nil {
 		return err
 	}
 	defer cl.Close()
+	return doReset(cctx, cl)
+}
 
+func (c *TemporalWorkflowResetCommand) getResetOperations() (validate func() error, doReset func(*CommandContext, client.Client) error) {
+	if c.WorkflowId != "" {
+		validate = c.validateWorkflowResetArguments
+		doReset = c.doWorkflowReset
+	} else {
+		validate = c.validateBatchResetArguments
+		doReset = c.runBatchReset
+	}
+	return validate, doReset
+}
+
+func (c *TemporalWorkflowResetCommand) validateWorkflowResetArguments() error {
+	if c.Type.Value == "" && c.EventId <= 0 {
+		return errors.New("must specify either valid event id or reset type")
+	}
+	if c.WorkflowId == "" {
+		return errors.New("must specify workflow id")
+	}
+	return nil
+}
+
+func (c *TemporalWorkflowResetCommand) validateBatchResetArguments() error {
+	if c.Type.Value == "" {
+		return errors.New("must specify reset type")
+	}
+	if c.RunId != "" {
+		return errors.New("must not specify run Id")
+	}
+	if c.EventId != 0 {
+		return errors.New("must not specify event Id")
+	}
+	if c.Type.Value == "BuildId" && c.BuildId == "" {
+		return errors.New("must specify build Id for BuildId based batch reset")
+	}
+	return nil
+}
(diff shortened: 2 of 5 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -464,21 +464,23 @@ temporal workflow reset --workflow-id=meaningful-business-id --type=LastContinue
...
-
+For batch reset only FirstWorkflowTask, LastWorkflowTask or BuildId can be used. Workflow Id, run Id and event Id 
+should not be set.
 Use the options listed below to change reset behavior.
 
 #### Options
 
-* `--workflow-id`, `-w` (string) - Workflow Id. Required.
+* `--workflow-id`, `-w` (string) - Workflow Id.
 * `--run-id`, `-r` (string) - Run Id.
```
cretz, line 474: May want to clarify that workflow ID is required if query is not present and run ID can only be present if workflow ID is present. (ug, I wish we could do `Includes options set for [single workflow or batch](#options-set-single-workflow-or-batch)` but I understand that `reason` is required makes us copy most of it)

temporalcli/commands.workflow.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	"github.com/google/uuid"
-	"github.com/temporalio/cli/temporalcli/internal/printer"
 	"go.temporal.io/api/batch/v1"
 	"go.temporal.io/api/common/v1"
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/query/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
+
+	"github.com/temporalio/cli/temporalcli/internal/printer"
```
cretz, line 17: Why'd this move?

temporalcli/commands.workflow_reset.go
```diff
@@ -5,23 +5,68 @@ import (
...
 	}
 	cl, err := c.Parent.ClientOptions.dialClient(cctx)
 	if err != nil {
 		return err
 	}
 	defer cl.Close()
+	return doReset(cctx, cl)
+}
+
+func (c *TemporalWorkflowResetCommand) getResetOperations() (validate func() error, doReset func(*CommandContext, client.Client) error) {
```
cretz, line 31: This is over-encapsulation IMO. Just inline most this logic that is broken out to these multiple tiny single-use methods. Returning callbacks to do stuff instead of just normal if statements in the main command body I think is unnecessarily complicated. (and even if, you can have short circuited returns and such)

temporalcli/commands.workflow_reset_test.go
```diff
@@ -207,3 +214,283 @@ func (s *SharedServerSuite) TestWorkflow_Reset_ToEventID() {
...
+}
+
+func newSystemUnderTest(suite *SharedServerSuite) batchResetTestData {
+	counters := make([]atomic.Int32, 4)
+	sut := batchResetTestData{
+		t:         suite.T(),
+		counters:  counters,
+		assert:    suite,
+		client:    suite.Client,
+		namespace: suite.Namespace(),
```
cretz, line 286: Can't you just pass the suite as part of the test data instead of a bunch of separate fields?

temporalcli/commands.workflow_reset_test.go
```diff
@@ -207,3 +214,283 @@ func (s *SharedServerSuite) TestWorkflow_Reset_ToEventID() {
...
+	}
+	sut.tq = suite.T().Name()
+	sut.buildPrefix = uuid.NewString()[:6] + "-"
```
cretz, line 289: It's just test code so it doesn't matter much, but a bit inconsistent on which fields you choose to set in the struct literal and which after.

temporalcli/commands.workflow_reset_test.go
```diff
@@ -207,3 +214,283 @@ func (s *SharedServerSuite) TestWorkflow_Reset_ToEventID() {
...
+	sut.ctx, sut.ctxCancelF = context.WithTimeout(context.Background(), 30*time.Second)
+	return sut
+}
+
+func (sut *batchResetTestData) internalVersionFor(v string) string {
+	return sut.buildPrefix + v
+}
+
+func (sut *batchResetTestData) firstActivity() error {
+	return increment(&sut.counters[firstActivity])
```
cretz, line 303: Why a whole separate `increment` function?

temporalcli/commands.workflow_reset_test.go
```diff
@@ -207,3 +214,283 @@ func (s *SharedServerSuite) TestWorkflow_Reset_ToEventID() {
...
+	Eventually(func() bool, time.Duration, time.Duration, ...interface{})
+}
+
+type batchResetTestData struct {
+	t           *testing.T
+	buildPrefix string
+	assert      assertions
+	client      client.Client
+	tq          string
+	counters    []atomic.Int32
```
cretz, line 271:
```suggestion
	counters    []*atomic.Int32
```

It makes more sense to pass around pointers to this

temporalcli/commandsmd/commands.md
```diff
@@ -464,21 +464,23 @@ temporal workflow reset --workflow-id=meaningful-business-id --type=LastContinue
...
 * `--event-id`, `-e` (int) - The Event Id for any Event after `WorkflowTaskStarted` you want to reset to (exclusive). It can be `WorkflowTaskCompleted`, `WorkflowTaskFailed` or others.
 * `--reason` (string) - The reason why this workflow is being reset. Required.
 * `--reapply-type` (string-enum) - Event types to reapply after the reset point. Options: All, Signal, None. Default: All.
-* `--type`, `-t` (string-enum) - Event type to which you want to reset. Options: FirstWorkflowTask, LastWorkflowTask, LastContinuedAsNew.
+* `--type`, `-t` (string-enum) - Event type to which you want to reset. Options: FirstWorkflowTask, LastWorkflowTask, LastContinuedAsNew, BuildId.
+* `--build-id` (string) - Only used if type is BuildId. Reset the first workflow task processed by this build id. Note that by default, this reset is allowed to be to a prior run in a chain of continue-as-new.
+* `--query`, `-q` (string) - Start a batch reset to operate on Workflow Executions with given List Filter.
+* `--yes`, `-y` (bool) - Confirm prompt to perform batch. Only allowed if query is present.
 
-### temporal workflow reset-batch: Reset a batch of Workflow Executions by reset type.
```
cretz, line 479: Can you make a note in the description how this is a breaking change because we are completely removing a command that exists in CLI today?

temporalcli/commands.workflow_reset_test.go
ast2023: No, you do not need a pointer here. The protocol for batchResetTestData is by sharing already. I should have return pointer from newSystemUnderTest to make it clearer. Will fix it.

cretz: Correct you do not _need_ it if you're sure you'll never `append`, but it's just a common pattern for atomic values to be used and passed around by pointers. Too many times people create slices for structs only to later `append` and be surprised when the underlying array when they could have just used pointers in a reasonable way when all operations require the pointer receiver. So in general in Go it's bad practice to take a pointer to a indexed slice value since a slice's underlying array can be mutated (even if it technically works here).

temporalcli/commands.workflow.go
ast2023: order of imports is different for CLI and OSS. My IDE is configured for OSS correct order. I wonder if we can/should have  project specific configurations.

temporalcli/commands.workflow_reset.go
ast2023: This level of encapsulation allows me to write code in declarative style Right now the `run` is very stright forward.  Understand how to do, validate, reset.

temporalcli/commands.workflow.go
cretz: Yeah, I am not a fan of the OSS approach that groups some non-stdlib imports separate from other non-stdlib imports. But it's not a big deal, it's pedantic.

temporalcli/commandsmd/commands.md
ast2023: What command do we remove? You mean we are changing reset_batch with reset with batch options?
Where should I add this note?

cretz: > What command do we remove?

https://docs.temporal.io/cli/workflow#reset-batch is the removed command. We're just trying to keep track of breaking changes we're making against current CLI.

>  Where should I add this note?

Just somewhere in the PR description is good. We'll be going back over PR descriptions for "known incompatibilities" when we build our first release notes.

cretz (approved):

PR #476, opened 2024-02-26
Adding namespace commands to cli

prathyushpv:
## What was changed
Adding operator namespace commands to cli. These are the commands that are added
* operator namespace create
* operator namespace delete
* operator namespace describe
* operator namespace list
* operator nemspace update

## Breaking changes
* Removed --verbose option from `namespace update` command. This option is used to print update request payload.
* Previous cli was printing historyArchivalUri and visibilityArchivalUri fields only if the option `--fields` is set to `long` when executing `namespace describe` command. Here we are always printing those fields since we dont have `--fields` option.

## Checklist

2. How was this tested:
Unit tests.

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.operator_namespace.go
+++ b/temporalcli/commands.operator_namespace.go
@@ -17,7 +17,7 @@ import (
 )
 
 func (c *TemporalOperatorNamespaceCreateCommand) run(cctx *CommandContext, args []string) error {
-	namespaceName := args[0]
+	nsName := args[0]
 	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
 	if err != nil {
 		return err
@@ -42,7 +42,7 @@ func (c *TemporalOperatorNamespaceCreateCommand) run(cctx *CommandContext, args 
 	}
 
 	_, err = cl.WorkflowService().RegisterNamespace(cctx, &workflowservice.RegisterNamespaceRequest{
-		Namespace:                        namespaceName,
+		Namespace:                        nsName,
 		Description:                      c.Description,
 		OwnerEmail:                       c.Email,
 		WorkflowExecutionRetentionPeriod: durationpb.New(retention),
@@ -58,42 +58,46 @@ func (c *TemporalOperatorNamespaceCreateCommand) run(cctx *CommandContext, args 
 	if err != nil {
 		return fmt.Errorf("unable to create namespace: %w", err)
 	}
-	cctx.Printer.Println(color.GreenString("Namespace %s successfully registered.", namespaceName))
+	cctx.Printer.Println(color.GreenString("Namespace %s successfully registered.", nsName))
 	return nil
 }
 
 func (c *TemporalOperatorNamespaceDeleteCommand) run(cctx *CommandContext, args []string) error {
-	namespaceName := args[0]
+	nsName := args[0]
 	yes, err := cctx.promptString(
-		color.RedString("Are you sure you want to delete namespace %s? Type namespace name to confirm:", namespaceName),
-		namespaceName,
+		color.RedString("Are you sure you want to delete namespace %s? Type namespace name to confirm:", nsName),
+		nsName,
 		c.Yes)
 	if err != nil {
 		return err
 	}
+
 	if !yes {
 		return fmt.Errorf("user denied confirmation")
 	}
+
 	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
 	if err != nil {
 		return err
 	}
 	defer cl.Close()
+
 	resp, err := cl.OperatorService().DeleteNamespace(cctx, &operatorservice.DeleteNamespaceRequest{
-		Namespace: namespaceName,
+		Namespace: nsName,
 	})
 	if err != nil {
 		return fmt.Errorf("unable to delete namespace: %w", err)
 	}
+
 	if cctx.JSONOutput {
 		return cctx.Printer.PrintStructured(resp, printer.StructuredOptions{})
 	}
-	cctx.Printer.Println(color.GreenString("Namespace %s has been deleted.", namespaceName))
+	cctx.Printer.Println(color.GreenString("Namespace %s has been deleted.", nsName))
 	return nil
 }
 
 func (c *TemporalOperatorNamespaceDescribeCommand) run(cctx *CommandContext, args []string) error {
-	namespaceName, namespaceID, err := getNamespaceFromIDArgs(cctx, c.NamespaceId, args)
+	nsName, nsID, err := getNamespaceFromIDArgs(cctx, c.NamespaceId, args)
 	if err != nil {
 		return err
 	}
@@ -103,13 +107,15 @@ func (c *TemporalOperatorNamespaceDescribeCommand) run(cctx *CommandContext, arg
 		return err
 	}
 	defer cl.Close()
+
 	resp, err := cl.WorkflowService().DescribeNamespace(cctx, &workflowservice.DescribeNamespaceRequest{
-		Namespace: namespaceName,
-		Id:        namespaceID,
+		Namespace: nsName,
+		Id:        nsID,
 	})
 	if err != nil {
 		return fmt.Errorf("unable to describe namespace: %w", err)
 	}
+
 	if cctx.JSONOutput {
 		return cctx.Printer.PrintStructured(resp, printer.StructuredOptions{})
 	}
@@ -131,12 +137,14 @@ func (c *TemporalOperatorNamespaceListCommand) run(cctx *CommandContext, args []
 		if err != nil {
 			return fmt.Errorf("failed listing namespaces: %w", err)
 		}
+
 		if cctx.JSONOutput {
 			// For JSON we are going to dump one line of JSON per execution
 			_ = cctx.Printer.PrintStructured(resp.Namespaces, printer.StructuredOptions{})
 		} else {
 			_ = printNamespaces(cctx, resp.Namespaces...)
 		}
+
 		nextPageToken = resp.GetNextPageToken()
 		if len(nextPageToken) == 0 {
 			return nil
@@ -145,7 +153,7 @@ func (c *TemporalOperatorNamespaceListCommand) run(cctx *CommandContext, args []
 }
 
 func (c *TemporalOperatorNamespaceUpdateCommand) run(cctx *CommandContext, args []string) error {
-	ns := args[0]
+	nsName := args[0]
 	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
 	if err != nil {
 		return err
@@ -156,9 +164,9 @@ func (c *TemporalOperatorNamespaceUpdateCommand) run(cctx *CommandContext, args 
 
 	if c.PromoteGlobal {
 		fmt.Printf("Will promote local namespace to global namespace for:%s, other flag will be omitted. "+
-			"If it is already global namespace, this will be no-op.\n", ns)
+			"If it is already global namespace, this will be no-op.\n", nsName)
 		updateRequest = &workflowservice.UpdateNamespaceRequest{
-			Namespace:        ns,
+			Namespace:        nsName,
 			PromoteNamespace: true,
 		}
 	} else if len(c.ActiveCluster) > 0 {
@@ -167,12 +175,12 @@ func (c *TemporalOperatorNamespaceUpdateCommand) run(cctx *CommandContext, args 
 			ActiveClusterName: c.ActiveCluster,
 		}
 		updateRequest = &workflowservice.UpdateNamespaceRequest{
-			Namespace:         ns,
+			Namespace:         nsName,
 			ReplicationConfig: replicationConfig,
 		}
 	} else {
 		resp, err := cl.WorkflowService().DescribeNamespace(cctx, &workflowservice.DescribeNamespaceRequest{
-			Namespace: ns,
+			Namespace: nsName,
 		})
 		if err != nil {
 			switch err.(type) {
@@ -193,6 +201,7 @@ func (c *TemporalOperatorNamespaceUpdateCommand) run(cctx *CommandContext, args 
 		if len(c.Email) > 0 {
 			ownerEmail = c.Email
 		}
+
 		data := map[string]string{}
 		if len(c.Data) > 0 {
 			data, err = stringKeysValues(c.Data)
@@ -200,6 +209,7 @@ func (c *TemporalOperatorNamespaceUpdateCommand) run(cctx *CommandContext, args 
 				return err
 			}
 		}
+
 		if len(c.Retention) > 0 {
 			ret, err := timestamp.ParseDurationDefaultDays(c.Retention)
 			if err != nil {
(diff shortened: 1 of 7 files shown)


temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+	yes, err := cctx.promptString(
+		color.RedString("Are you sure you want to delete namespace %s? Type namespace name to confirm:", nsName),
+		nsName,
+		c.Yes)
+	if err != nil {
+		return err
+	}
+
+	if !yes {
+		return fmt.Errorf("user denied confirmation")
```
cretz, line 76: Or mistyped the name

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+	switch input {
+	case "disabled":
+		return enums.ARCHIVAL_STATE_DISABLED
+	case "enabled":
+		return enums.ARCHIVAL_STATE_ENABLED
+	}
+	return enums.ARCHIVAL_STATE_UNSPECIFIED
+}
+
+func getNamespaceFromIDArgs(cctx *CommandContext, nsID string, args []string) (string, string, error) {
```
cretz, line 320: Can you inline this instead of making a single-use package-level method for all?

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+		default:
+			return fmt.Errorf("namespace update failed: %w", err)
+		}
+	} else {
+		cctx.Printer.Println(color.GreenString("namespace %s update succeeded.", nsName))
+	}
+	return nil
+}
+
+func printNamespaces(cctx *CommandContext, responses ...*workflowservice.DescribeNamespaceResponse) error {
```
cretz, line 272:
```suggestion
func printNamespaceDescriptions(cctx *CommandContext, responses ...*workflowservice.DescribeNamespaceResponse) error {
```

Pedantic, don't have to change, it's just that this top-level function is available for everyone in this package even though it's very clearly only for operator describe/list use

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+		resp, err := cl.WorkflowService().ListNamespaces(cctx, &workflowservice.ListNamespacesRequest{
+			NextPageToken: nextPageToken,
+			PageSize:      100,
+		})
+		if err != nil {
+			return fmt.Errorf("failed listing namespaces: %w", err)
+		}
+
+		if cctx.JSONOutput {
+			// For JSON we are going to dump one line of JSON per execution
```
cretz, line 142: Hrmm, I wonder if this is needed as a lazy stream, but it is fine. I have added this to #448.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	var updateRequest *workflowservice.UpdateNamespaceRequest
+
+	if c.PromoteGlobal {
+		fmt.Printf("Will promote local namespace to global namespace for:%s, other flag will be omitted. "+
```
cretz, line 166: Do not use printing directly, this will mess up JSON output. Use a logger or the printer.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+	nsName := args[0]
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	var updateRequest *workflowservice.UpdateNamespaceRequest
+
+	if c.PromoteGlobal {
```
cretz, line 165: If these are mutually exclusive options, we should probably error if multiple are provided

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+			Namespace: nsName,
+		})
+		if err != nil {
+			switch err.(type) {
+			case *serviceerror.NamespaceNotFound:
+				return err
+			default:
+				return fmt.Errorf("namespace update failed: %w", err)
+			}
+		}
```
cretz, line 192: Kinda strange Go here. This is clearer IMO:

```suggestion
		if _, ok := err.(*serviceerror.NamespaceNotFound); ok {
			return err
		} else if err != nil {
			return fmt.Errorf("namespace update failed: %w", err)
		}
```

Then again, not sure why we can't just `return fmt.Errorf("failed describing namespace for update: %w", err)` and not have the if/else. I think the old CLI had a different exit code for not found which we don't do anymore. Same for this same logic on the error later.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+		updateRequest = &workflowservice.UpdateNamespaceRequest{
+			Namespace:         nsName,
+			UpdateInfo:        updateInfo,
+			Config:            updateConfig,
+			ReplicationConfig: replicationConfig,
+		}
+	}
+
+	if c.Verbose {
```
cretz, line 254: I vote we break compat and remove `verbose` option for this call. If we want client request/response verbosity, we can do it at a higher level (i.e. via gRPC interceptor layer) and we'd use the logger not the printer.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+		_ = cctx.Printer.PrintStructured(updateRequest, printer.StructuredOptions{})
+	}
+
+	_, err = cl.WorkflowService().UpdateNamespace(cctx, updateRequest)
```
cretz, line 258: I think if JSON is enabled we should dump the response

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		switch err.(type) {
+		case *serviceerror.NamespaceNotFound:
+			return err
+		default:
+			return fmt.Errorf("namespace update failed: %w", err)
+		}
+	} else {
+		cctx.Printer.Println(color.GreenString("namespace %s update succeeded.", nsName))
+	}
```
cretz, line 268:
```suggestion
	}
	cctx.Printer.Println(color.GreenString("Namespace %s update succeeded.", nsName))
```

No need for `else`, and capitalizing success message as you have in other commands

temporalcli/commands.operator_namespace_test.go
```diff
@@ -0,0 +1,204 @@
...
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/temporalio/cli/temporalcli"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+func (s *SharedServerSuite) TestOperator_Namespace_Create_List_And_Describe() {
```
cretz, line 12:
```suggestion
func (s *SharedServerSuite) TestOperator_Namespace_CreateListAndDescribe() {
```

No need to underscore every word IMO (but pedantic, no need to change if you don't want)

temporalcli/commands.operator_namespace_test.go
```diff
@@ -0,0 +1,204 @@
...
+	s.NoError(res.Err)
+	var describeResp workflowservice.DescribeNamespaceResponse
+	s.NoError(temporalcli.UnmarshalProtoJSONWithOptions(res.Stdout.Bytes(), &describeResp, true))
+
+	s.Equal("description after", describeResp.NamespaceInfo.Description)
+	s.Equal("email@after", describeResp.NamespaceInfo.OwnerEmail)
+	s.Equal(48*time.Hour, describeResp.Config.WorkflowExecutionRetentionTtl.AsDuration())
+}
+
+func (s *SharedServerSuite) TestNamespaceUpdate_Data() {
```
cretz, line 105: Wonder if you want to move this with your create-list-and-describe so you don't mess with the `default` namespace everyone else uses. Can move delete over there too. I understand it can seem a bit wrong to test so many separate things in a single test, but in these cases of integration with several parts it makes sense.

cretz (commented):
Looks good! Mostly pedantic/minor things.

temporalcli/commands.operator_namespace_test.go
prathyushpv: Right. I moved this to TestNamespaceUpdate.

temporalcli/commandsmd/commands.md
```diff
@@ -209,6 +210,107 @@ Cluster commands follow this syntax: `temporal operator cluster [command] [comma
...
+### temporal operator namespace describe [namespace]: Describe a Namespace by its name or ID.
+
+The temporal operator namespace describe command provides Namespace information.
+Namespaces are identified by Namespace ID.
+
+`temporal operator namespace describe --namespace-id=some-namespace-id`
+`temporal operator namespace describe example-namespace-name`
+
+<!--
+* maximum-args=1
```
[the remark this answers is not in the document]
prathyushpv: I have already added this under exact-args documentation.

temporalcli/commandsmd/commands.md
```diff
@@ -209,6 +210,107 @@ Cluster commands follow this syntax: `temporal operator cluster [command] [comma
...
+
+* `--active-cluster` (string) - Active cluster name.
+* `--cluster` (string[]) - Cluster names.
+* `--data` (string) - Namespace data in key=value format. Use JSON for values.
+* `--description` (string) - Namespace description.
+* `--email` (string) - Owner email.
+* `--global` (bool) - Whether the namespace is a global namespace.
+* `--history-archival-state` (string-enum) - History archival state. Options: disabled, enabled. Default: disabled.
+* `--history-uri` (string) - Optionally specify history archival URI (cannot be changed after first time archival is enabled).
+* `--retention` (string) - Length of time (in days) a closed Workflow is preserved before deletion. Default: 3.
```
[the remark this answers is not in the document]
prathyushpv: Right! I did not notice that we have duration time.

temporalcli/commandsmd/commands.md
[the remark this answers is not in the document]
prathyushpv: A flag for namespace name would have been better.

But we don't use the global --namespace flag in this command as it has a different meaning. In that case, we will need a new flag like --namespace-name.

Also, the previous cli will fail if `--namespace` is set when using namespace commands. But I could not add that check here because we set it to default namespace.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,335 @@
...
+
+func (c *TemporalOperatorNamespaceCreateCommand) run(cctx *CommandContext, args []string) error {
+	nsName := args[0]
+	cl, err := c.Parent.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	retention, err := timestamp.ParseDurationDefaultDays(c.Retention)
```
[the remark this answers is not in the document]
prathyushpv: this line is now removed.

temporalcli/commands.operator_namespace.go
prathyushpv: Oh! I copied it from the old cli code and forgot to change it.

temporalcli/commands.operator_namespace.go
prathyushpv: Yeah right. I just removed this check.

temporalcli/commandsmd/commands.md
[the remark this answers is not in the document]
cretz: I agree, will discuss with team about flag vs option here.

cretz (commented):
This is looking great! Only reason I am not approving is because we are still discussing internally on the use of positionals vs `--namespace`/`-n` here. It's starting to lean towards the latter which would be a compat break, we'll see. Will get answer soon hopefully.

Also, we need to make sure we document all compat breaks (e.g. removing `--verbose`) in the PR description.

prathyushpv:
> This is looking great! Only reason I am not approving is because we are still discussing internally on the use of positionals vs `--namespace`/`-n` here. It's starting to lean towards the latter which would be a compat break, we'll see. Will get answer soon hopefully.
> 
> Also, we need to make sure we document all compat breaks (e.g. removing `--verbose`) in the PR description.

Thank you! I have updated the description with breaking changes. Let me know when the decision is made about namespace name.

cretz:
@prathyushpv - after internal discussion, let's go with the way you have it now (which is the way the CLI works today. But if you don't mind, I would like this to way on #487 and put the `StartList` and `EndList` calls like you see in that PR on the `operator namespace list` command.

cretz:
@prathyushpv - you can merge main and add the `StartList`/`EndList` stuff now. Can leave positional params as they are (we don't like it, but we don't have a better answer right now).

prathyushpv:
> @prathyushpv - you can merge main and add the `StartList`/`EndList` stuff now. Can leave positional params as they are (we don't like it, but we don't have a better answer right now).

@cretz Thank you! I have added `StartList` and`EndList` to listNamespace command.

temporalcli/commands.operator_namespace.go
```diff
@@ -0,0 +1,318 @@
...
+		resp, err := cl.WorkflowService().ListNamespaces(cctx, &workflowservice.ListNamespacesRequest{
+			NextPageToken: nextPageToken,
+			PageSize:      100,
+		})
+		if err != nil {
+			return fmt.Errorf("failed listing namespaces: %w", err)
+		}
+
+		for _, ns := range resp.GetNamespaces() {
+			if cctx.JSONOutput {
```
cretz, line 149: Might as well put this condition around the loop instead of inside, but doesn't matter much

cretz (approved):
Looks great, thanks! May have to fix failing CI test.

temporalcli/commands.operator_namespace.go
prathyushpv: If it is printed one by one as a card, it is not putting a new line between cards. Is there any option for getting a new line after each card is printed? Or do you suggest printing a new line after each card here?

prathyushpv:
> Looks great, thanks! May have to fix failing CI test.

Thank you!

temporalcli/commands.operator_namespace.go
prathyushpv: Let me merge this and try to fix that in a different PR.

prathyushpv: It is already around the inner loop. If I put it outside the outer loop, then we will have to create a list of all namespaces in the cluster.

cretz: To clarify, change:

```go
		for _, ns := range resp.GetNamespaces() {
			if cctx.JSONOutput {
				_ = cctx.Printer.PrintStructured(ns, printer.StructuredOptions{})
			}
		}
```
to:
```go
		if cctx.JSONOutput {
			for _, ns := range resp.GetNamespaces() {
				_ = cctx.Printer.PrintStructured(ns, printer.StructuredOptions{})
			}
		}
```

No reason to run a loop you'll never need

cretz:
:+1: Merge whenever you're ready

Issue #342, opened 2023-09-25
Ability to set authorization via Envar


Smithx10:
In tctl we can set TEMPORAL_CLI_AUTH="Bearer $token",  in temporal cli from my quick look we have to continuously use ```--grpc-meta```. 

From common/flags.go I couldn't find any way to set such a thing.

```go
	&cli.StringSliceFlag{
		Name:     FlagMetadata,
		Usage:    FlagMetadataDefinition,
		Category: CategoryGlobal,
	},
```

josh-berry:
Duly noted, will keep this in mind for a future release!

metavee:
It would be great to see this implemented! It's not ideal to be passing credentials back and forth on the command line.

josh-berry:
@cretz is planning to add a --api-key option that will have an associated environment variable in #478, so we can probably drop this one in favor of that.

Issue #478, opened 2024-02-27
CLI Refresh: API key client option


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/426

cretz:
Note, this will include an env var for this `--api-key` setting.

PR #480, opened 2024-02-28
Allow setting the auth token via an env var

Sushisource:
## What was changed
Allow setting `Authorization` header via an env var

## Why?
Existed in tctl, request from user.

## Checklist

1. Closes https://github.com/temporalio/cli/issues/342
2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/temporalcli/client.go
+++ b/temporalcli/client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"maps"
 	"net/http"
 	"os"
 	"os/user"
@@ -33,17 +34,17 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error) 
 	}
 
 	// Headers
+	headers := &headersProvider{setByArgs: make(map[string]string)}
 	if len(c.GrpcMeta) > 0 {
-		headers := make(stringMapHeadersProvider, len(c.GrpcMeta))
 		for _, kv := range c.GrpcMeta {
 			pieces := strings.SplitN(kv, "=", 2)
 			if len(pieces) != 2 {
 				return nil, fmt.Errorf("gRPC meta of %q does not have '='", kv)
 			}
-			headers[pieces[0]] = pieces[1]
+			headers.setByArgs[pieces[0]] = pieces[1]
 		}
-		clientOptions.HeadersProvider = headers
 	}
+	clientOptions.HeadersProvider = headers
 
 	// Remote codec
 	if c.CodecEndpoint != "" {
@@ -155,10 +156,19 @@ func clientIdentity() string {
 	return "temporal-cli:" + username + "@" + hostname
 }
 
-type stringMapHeadersProvider map[string]string
+type headersProvider struct {
+	setByArgs map[string]string
+}
 
-func (s stringMapHeadersProvider) GetHeaders(context.Context) (map[string]string, error) {
-	return s, nil
+func (h *headersProvider) GetHeaders(context.Context) (map[string]string, error) {
+	retme := make(map[string]string)
+	maps.Copy(retme, h.setByArgs)
+	if _, ok := retme["Authorization"]; !ok {
+		if authHeader := os.Getenv(`TEMPORAL_CLI_AUTHORIZATION_TOKEN`); authHeader != "" {
+			retme["Authorization"] = authHeader
+		}
+	}
+	return retme, nil
 }
 
 var dataConverter = converter.NewCompositeDataConverter(
--- a/temporalcli/client.go
+++ b/temporalcli/client.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"maps"
 	"net/http"
 	"os"
 	"os/user"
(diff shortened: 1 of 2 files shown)


temporalcli/client.go
```diff
@@ -155,10 +156,19 @@ func clientIdentity() string {
...
 
-func (s stringMapHeadersProvider) GetHeaders(context.Context) (map[string]string, error) {
-	return s, nil
+func (h *headersProvider) GetHeaders(context.Context) (map[string]string, error) {
+	retme := make(map[string]string)
+	maps.Copy(retme, h.setByArgs)
+	if _, ok := retme["Authorization"]; !ok {
+		if authHeader := os.Getenv(`TEMPORAL_CLI_AUTHORIZATION_TOKEN`); authHeader != "" {
+			retme["Authorization"] = authHeader
+		}
```
tdeebswihart, line 169: You could simplify this with a `newStringMapHeadersProvider` function that pre-fills the Auth header, while retaining all the old logic

That'd also allow folks to override the environment variable with an argument

tdeebswihart (approved):

temporalcli/client.go
Sushisource: Goo point. I was thinking it might be nice to change the env var at runtime but... then I remembered that's not really a thing.

temporalcli/client.go
```diff
@@ -157,6 +157,14 @@ func clientIdentity() string {
 
 type stringMapHeadersProvider map[string]string
 
+func newStringMapHeadersProvider() stringMapHeadersProvider {
+	rmap := make(map[string]string)
+	if authHeader := os.Getenv(`TEMPORAL_CLI_AUTHORIZATION_TOKEN`); authHeader != "" {
```
dnr, line 162: aren't we using `TEMPORAL_X` for env vars instead of `TEMPORAL_CLI_X`? I already switched all my scripts and everything...

Sushisource: Good question. There aren't any in the CLI rewrite so far, and the only one in the repo's main branch is https://github.com/temporalio/cli/blob/main/common/defs.go#L38

Which uses `CLI`. I assume the env vars you're talking about are for server itself? Seems reasonable to draw the distinction here, plus this is the same name used for this concept back in `tctl`, not that that necessarily matters much.

I'm fine either way. Will avoid merging this for a bit to see if anyone else has opinions.

dnr: For tctl they were all `TEMPORAL_CLI_`, for temporal (old new cli) they were all changed to `TEMPORAL_`. It caused a small amount of hassle but now that we've done that everywhere, I'd much rather keep them `TEMPORAL_`

dnr: There are a bunch in the main branch: https://github.com/search?q=repo%3Atemporalio%2Fcli+TEMPORAL_&type=code

cretz: > There aren't any in the CLI rewrite so far

There are, but they are all tied to command options, so you'll find them in `commands.md`. They are just `TEMPORAL_X`. I would recommend `TEMPORAL_AUTH_HEADER` here.

I personally think a better approach to this whole thing is to do https://github.com/temporalio/cli/issues/478 and give that an env var for that option (ideally waiting on Go SDK to implement it first). Almost everyone will be doing `Bearer <token>` here anyways.

cretz: > There aren't any in the CLI rewrite so far

There are, but they are all tied to command options, so you'll find them in `commands.md`. They are just `TEMPORAL_X`. I would recommend `TEMPORAL_AUTH_HEADER` here.

I personally think a better approach to this whole thing is to do https://github.com/temporalio/cli/issues/478 and give that an env var for that option (ideally waiting on Go SDK to implement it first). Almost everyone will be doing `Bearer <token>` here anyways.

cretz, line 162: Can you use `CommandContext.Options.LookupEnv` here? That's out common abstraction for accessing external things like env vars so that tests can set it.

temporalcli/client.go
cretz, line 1: I have a general suggestion to take a different approach:

* Implement https://github.com/temporalio/sdk-go/issues/1401
* Implement https://github.com/temporalio/cli/issues/478 as `--api-key` but accept `TEMPORAL_API_KEY` env var that passes to that client
* Accept that this means we are assume the `Authorization` header value is prefixed with `Bearer ` but that is the most common use anyways
  * If others need more flexible header setting, we should add env var support to array-based options and should support a `TEMPORAL_GRPC_META` env var (or multiple with that prefix) for `--grpc-meta` so that everyone can set headers via env vars of any name/value

We have to support API keys anyways, so lets just add an env var to that.

temporalcli/client.go
Sushisource: Ah, OK. I was wondering why I found so few things.

temporalcli/client.go
Sushisource: OK, works for me :+1:

cretz: Awesome. Feel free to knock out the Go one yourself, or otherwise we should just block this CLI feature until that appears on the Go SDK client (it's a tctl-parity feature not a current-cli-parity feature).

Sushisource: I doubt it'll be a problem but I'll finish the other CLI things I have before coming back to this to make sure it's all agreed

Sushisource:
Dropping this in favor of a slightly different solution as mentioned here https://github.com/temporalio/cli/issues/478

Issue #309, opened 2023-08-08
[Feature Request] Log metrics endpoint's port number for visibility


taonic:
### Is your feature request related to a problem? Please describe.
In `server start-dev` command, when the `--metrics-port` is omitted, the cli will [assign a free port automatically](https://github.com/temporalio/cli/blob/f8c84edae78379ad3716ed51c84120c923db9388/server/config/config.go#L152).

However, unlike the frontend or pprof port, the automatically assigned metrics port is never advertised in the log.

```
{"level":"info","ts":"2023-08-08T12:17:07.286+1000","msg":"PProf listen on ","port":63884,"logging-call-at":"pprof.go:73"}
⇨ http server started on 127.0.0.1:8233
```

### Describe the solution you'd like
I'd like the metrics endpoint to be advertised when starting the dev server.

### Additional context
Happy to contribute a PR.

PR #482, opened 2024-02-28
Log metrics port

Sushisource:
## What was changed
Log the port metrics uses on startup

## Why?
Useful

## Checklist

1. Closes https://github.com/temporalio/cli/issues/309
2. How was this tested:
Manually

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.server.go
+++ b/temporalcli/commands.server.go
@@ -60,6 +60,10 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 			_, _ = cctx.Options.Stderr.Write(b)
 		}
 	}
+	// Grab a free port for metrics ahead-of-time so we know what port is selected
+	if opts.MetricsPort == 0 {
+		opts.MetricsPort = devserver.MustGetFreePort()
+	}
 
 	// Start, wait for context complete, then stop
 	s, err := devserver.Start(opts)
@@ -76,6 +80,7 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 	if !t.Headless {
 		cctx.Printer.Printlnf("Web UI is running at: http://%v:%v", friendlyIP, opts.UIPort)
 	}
+	cctx.Printer.Printlnf("Metrics available at: http://%v:%v/metrics", friendlyIP, opts.MetricsPort)
 	<-cctx.Done()
 	cctx.Printer.Println("Stopping server...")
 	return nil


Sushisource:
Created https://github.com/temporalio/temporal/issues/5461 since I had to work around this a bit

temporalcli/commands.server.go
```diff
@@ -60,6 +60,10 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 			_, _ = cctx.Options.Stderr.Write(b)
 		}
 	}
+	// Grab a free port for metrics ahead-of-time so we know what port is selected
+	if opts.MetricsPort == 0 {
+		opts.MetricsPort = devserver.MustGetFreePort()
+	}
```
cretz, line 66: Hrmm, so looking at the dev server code, I am not seeing that metrics config is even set if no metrics port is given by the user which I wonder if that means that no metrics endpoint is even created/bound. Can you confirm what happens today if metrics port is unset (i.e. 0)?

Sushisource: It is indeed created and started still - but only with LiteServer (which is the one that gets used here, right?) https://github.com/temporalio/temporal/blob/a2b17bebadc914f6445e9bd85be126f94730b25f/internal/temporalite/lite_server.go#L121

cretz: > which is the one that gets used here, right?

I don't think so. But I wonder if I introduced a regression copying over dev server code. Current CLI does default a metrics port: https://github.com/temporalio/cli/blob/7267df668277a9b43c50d454af50ff5e2b9f8e90/server/config/config.go#L153-L155. But new CLI does not. I kinda like the behavior of not starting/binding a metrics server if they don't ask for it now that I think about it. But I think your current code re-institutes what was in the CLI before.

Can be convinced either way (can also get others' opinion on whether we should always start a metrics server if you want).

Sushisource: I figure we might as well. I'm just not sure I see any real reason not too, and it's nice to have the metrics there if at some point after starting it you realize you wanted to see something.

Sushisource: Eh, I think the discoverability is a nice thing actually. A curious person might open it up, go "oh hey this is nice" who otherwise never would've known.

cretz (approved):
Not a big fan of forcing a metrics server (and output line) for something most will never use, but not a big deal

Issue #293, opened 2023-07-27
[Feature Request] Log notice when --env is getting overrode by env var


lorensr:
### Is your feature request related to a problem? Please describe.

User thought `--env` wasn't working. Issue was it was being overridden by `TEMPORAL_ADDRESS` env var.

### Describe the solution you'd like

If `--env prod` is provided but gets overrode, log:

```
Using TEMPORAL_ADDRESS=127.0.0.1:7233 (overrides prod.address)
```

### Alternative
Have `--env` override env var? I think I actually like this better?

KevM:
I think the _Alternative_ proposed is what the user is expecting. They explicitly asked for a environment they previously provided.

Sushisource:
Github not auto-closing properly.

I've added the log but not the "env file always takes precedence" behavior. I think you can make a pretty good argument in both directions there.

cretz:
Update here. With the coming `config` concept and `env`'s eventual deprecation, it will actually be very normal to override config-file-set values with environment variables, so we will no longer be logging a notice here.

PR #483, opened 2024-02-28
Log env vars which override the env file

Sushisource:
## What was changed
Log any time an env var overrides an env file setting

## Why?
It can be a bit confusing if you've forgotten you've set an env var. Helpful reminder.

## Checklist

1. Closes https://github.com/temporalio/cli/issues/293
2. How was this tested:
Added unit test

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -229,10 +229,13 @@ func UnmarshalProtoJSONWithOptions(b []byte, m proto.Message, jsonShorthandPaylo
 	return opts.Unmarshal(b, m)
 }
 
-func (c *CommandContext) populateFlagsFromEnv(flags *pflag.FlagSet) error {
+// Set flag values from environment file & variables. Returns a map of flags->values which were
+// overridden by environment variables.
+func (c *CommandContext) populateFlagsFromEnv(flags *pflag.FlagSet) (func(*slog.Logger), error) {
 	if flags == nil {
-		return nil
+		return func(logger *slog.Logger) {}, nil
 	}
+	var logCalls []func(*slog.Logger)
 	var flagErr error
 	flags.VisitAll(func(flag *pflag.Flag) {
 		// If the flag was already changed by the user, we don't overwrite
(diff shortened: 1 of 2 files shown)


temporalcli/commands.go
```diff
@@ -229,10 +229,13 @@ func UnmarshalProtoJSONWithOptions(b []byte, m proto.Message, jsonShorthandPaylo
 	return opts.Unmarshal(b, m)
 }
 
-func (c *CommandContext) populateFlagsFromEnv(flags *pflag.FlagSet) error {
+// Set flag values from environment file & variables. Returns a map of flags->values which were
+// overridden by environment variables.
```
cretz, line 233:
```suggestion
// Set flag values from environment file & variables.
```

Doesn't return a map (though may read better if it did return a list of changes instead of a logger callback, but this is a private function so it's fine)

temporalcli/commands.go
```diff
@@ -254,11 +257,21 @@ func (c *CommandContext) populateFlagsFromEnv(flags *pflag.FlagSet) error {
 						flag.Name, anns[0], envVal, err)
 					return
 				}
+				if flag.Changed {
+					logCalls = append(logCalls, func(l *slog.Logger) {
+						l.Info("Env var overrode --env setting", "env_var", anns[0], "flag", flag.Name, "value", envVal)
```
cretz, line 262: Hrmm, not sure we ever clarified the format of log keys (i.e. snake_case or what)

cretz, line 262:
```suggestion
						l.Info("Env var overrode --env setting", "env_var", anns[0], "flag", flag.Name)
```

I think we should leave off the value, this can be confidential information and not sure it's needed.

cretz (commented):
LGTM, only thing I really care about before approving is to remove the value from the log

temporalcli/commands.go
Sushisource: Good point

temporalcli/commands.go
Sushisource: Yeah forgot to change this, I didn't want to make a new struct type just to return what was altered

cretz (approved):

PR #486, opened 2024-02-29
defs-cmds: fix typo in help string

ikonst:


josh-berry (approved):

cretz:
Thanks, will merge on CI pass! (note most of this code is being replaced and as part of that docs as a whole will be revisited)

cretz:
Actually, we also need a CLA approval. Let me get that setup on this repo.

cretz:

PR #487, opened 2024-02-29
JSONL support

### LINKED ISSUE #448: [CLI Refresh] Output valid JSON and/or JSONL for `workflow list`

cretz:
## What was changed

* Add `--output jsonl` option which removes indentation/newlines from JSON output
* Add printer `StartList` and `EndList`
  * In `--output json` mode, this will add `[` at the beginning, commas before newline of each successive object, and `]` at the end so that it will parse as an array
  * In `--output jsonl` mode, this will do nothing special (indention is already off, just allows normal use of repeated JSON values)
* Added `StartList` and `EndList` to `workflow list`, `batch list`, and `cluster list`

(note, there may be an outstanding flaky batch test we are researching)

## Checklist

1. Closes #448


temporalcli/commands.batch.go
```diff
@@ -70,6 +70,10 @@ func (c TemporalBatchListCommand) run(cctx *CommandContext, args []string) error
 	}
 	defer cl.Close()
 
+	// This is a listing command subject to json vs jsonl rules
+	cctx.Printer.StartList()
+	defer cctx.Printer.EndList()
```
dandavison, line 75: I'm thinking maybe we should use the term "array" rather than "list" in the code since that's the correct JSON terminology.

cretz: `list` is what we call the commands and I don't want users to even have to worry about whether JSON is enabled or not. I want people to say whether their jsonl list is starting/ending too. Often people don't think of arrays as lazy/streaming. 

Having said this, if the word "array" is important enough, I can change to get PR approval.

temporalcli/internal/printer/printer.go
```diff
@@ -50,6 +54,42 @@ func (p *Printer) Printlnf(s string, v ...any) {
 	p.Println(fmt.Sprintf(s, v...))
 }
 
+// When called for JSON with indent, this will create an initial bracket and
+// make sure all [Printer.PrintStructured] calls get commas properly to appear
+// as a list (but the indention and multiline posture of the JSON remains). When
+// called for JSON without indent, this will make sure all
+// [Printer.PrintStructured] is on its own line (i.e. JSONL mode). When called
+// for non-JSON, this is a no-op.
```
Sushisource, line 62: This (and the if checks below) says to me that rather than using an empty `JSONIndent` as a special indication to do JSONL, we should instead just have the `JSON` field be `JSON | JSONL` and just be explicit about what mode we're in.

cretz: > we should instead just have the `JSON` field be `JSON | JSONL`

More like `JSON | JSONL | Neither`, but it's ugly in Go. I considered this (and I considered two booleans), but `JSONIndent` is how Go does it (where empty string means no multiline, ref https://pkg.go.dev/encoding/json#Encoder.SetIndent). Throughout the Go JSON printing code they check whether indentation is set. And this option gives the caller the freedom to set the indent they want, even though we're the caller. However, if the way Go did this is too confusing, I can make this a Go enum and add a `isJSON()` that checks for those two values.

Sushisource: I'd prefer that I think but it's not a big deal. I'll approve and leave it up to you. Lack up sum types :cry:

Sushisource (approved):

cretz:
Merging, but I do not hold the use of empty-jsonindent-means-jsonl nor the use of list vs array as a strong opinion if anyone really wants them changed.

temporalcli/commands.batch.go
dandavison: Thanks, that's fine, especially if there's a connection with the use of "list" as a verb in the CLI.

PR #491, opened 2024-03-06
 Adding task-queue list-partition command

prathyushpv:
## What was changed
Adding task-queue list-partition command

## Breaking changes
* This command had --output option which can take values table, card and json. We will not have card and table options. Option text will print a table view.
* --output json was printing some text line in current cli. Here we will print response object without that.

## Why?
cli rewrite


temporalcli/commands.taskqueue.go
```diff
@@ -99,3 +99,44 @@ func (c *TemporalTaskQueueDescribeCommand) run(cctx *CommandContext, args []stri
...
+	}
+
+	var items []*taskqueue.TaskQueuePartitionMetadata
+	cctx.Printer.Println(color.MagentaString("Workflow Task Queue Partitions\n"))
+	for _, e := range resp.WorkflowTaskQueuePartitions {
+		items = append(items, e)
+	}
+	_ = cctx.Printer.PrintStructured(items, printer.StructuredOptions{Table: &printer.TableOptions{}})
+
+	items = nil
```
cretz, line 134:
```suggestion
	items = items[:0]
```

Don't need to change this, very pedantic, but just thought I'd pass along a nice little Go trick for slice reuse w/ mem savings if you find yourself in other more performance-minded code

temporalcli/commandsmd/commands.md
```diff
@@ -466,6 +466,14 @@ Fetch the sets of compatible build IDs associated with a Task Queue and associat
 * `--task-queue`, `-t` (string) - Task queue name. Required.
 * `--max-sets` (int) - Limits how many compatible sets will be returned. Specify 1 to only return the current default major version set. 0 returns all sets. (default: 0). Default: 0.
 
+###  temporal task-queue list-partition: Lists the Task Queue's partitions and the matching nodes they are assigned to.
```
cretz, line 469:
```suggestion
### temporal task-queue list-partition: Lists the Task Queue's partitions and the matching nodes they are assigned to.
```

cretz (approved):
Looks great! Just one minor change wrt spacing on the markdown, but it's not technically required because we trim during parse.

temporalcli/commands.taskqueue.go
prathyushpv: Nice! I was not aware of that..

prathyushpv:
Thanks @cretz !

PR #496, opened 2024-03-12
CLI Refresh: `workflow delete` command

carlydf:
## What was changed
Add support for workflow delete command. The functionality is mostly unchanged from current CLI, but the output may be different.

## Why?
For the CLI refresh

## Checklist

1. How was this tested:
Tested in commands.workflow_test.go and via CLI with local dev environment.

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -3,6 +3,7 @@
 package temporalcli
 
 import (
+	"fmt"
 	"github.com/mattn/go-isatty"
 
 	"github.com/spf13/cobra"
@@ -1357,6 +1358,7 @@ func NewTemporalWorkflowCountCommand(cctx *CommandContext, parent *TemporalWorkf
 type TemporalWorkflowDeleteCommand struct {
 	Parent  *TemporalWorkflowCommand
 	Command cobra.Command
+	SingleWorkflowOrBatchOptions
 }
 
 func NewTemporalWorkflowDeleteCommand(cctx *CommandContext, parent *TemporalWorkflowCommand) *TemporalWorkflowDeleteCommand {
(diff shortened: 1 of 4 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -570,7 +570,26 @@ TODO
...
+
+#### Options set for single workflow or batch:
+
+* `--workflow-id`, `-w` (string) - Workflow Id. Either this or query must be set.
+* `--run-id`, `-r` (string) - Run Id. Cannot be set when query is set.
+* `--query`, `-q` (string) - Start a batch to operate on Workflow Executions with given List Filter. Either this or
+  Workflow Id must be set.
+* `--reason` (string) - Reason to perform batch. Only allowed if query is present unless the command specifies
+  otherwise. Defaults to message with the current user's name.
+* `--yes`, `-y` (bool) - Confirm prompt to perform batch. Only allowed if query is present.
```
cretz, line 592:
```suggestion
#### Options

Includes options set for [single workflow or batch](#options-set-single-workflow-or-batch)
```

Should be able to just reference the existing one (we should probably add parser checks to ensure there are not two options sets for the same name)

temporalcli/commands.workflow.go
```diff
@@ -91,7 +121,7 @@ func (c *TemporalWorkflowSignalCommand) run(cctx *CommandContext, args []string)
 			return fmt.Errorf("failed signalling workflow: %w", err)
 		}
 		cctx.Printer.Println("Signal workflow succeeded")
-	} else if batchReq != nil {
+	} else { // batchReq != nil
```
cretz, line 124: Seems fine to me. There was a time where we didn't return `err` when people declined the prompt, so technically all three results of `workflowExecOrBatch` could have been nil, but we changed that and didn't come back and change this, so all good.

temporalcli/commands.workflow_test.go
```diff
@@ -119,6 +119,49 @@ func (s *SharedServerSuite) testSignalBatchWorkflow(json bool) *CommandResult {
...
+			Namespace: s.Namespace(),
+		})
+		s.NoError(err)
+
+		if len(wfs.GetExecutions()) == 1 && wfs.GetExecutions()[0].GetExecution().GetWorkflowId() == "3" {
+			return true
+		}
+
+		return false
+	}, 10*time.Second, time.Second, "timed out awaiting for workflows termination")
```
cretz, line 162:
```suggestion
	}, 5*time.Second, 100 * time.Millisecond, "timed out awaiting for workflows termination")
```

I know list is eventually consistent, but since it's a local server only, we can call it a bit faster and have a faster-completing test

temporalcli/commands.gen.go
cretz, line 1: Did you hand-edit this file? It is generated so it will be overwritten next generation.

carlydf: i did hand-edit it before beginning to suspect that it was supposed to be generated by the commands.md file. I'll try generating it instead

carlydf: done :)

temporalcli/commands.workflow.go
carlydf: changed it on delete, cancel, and terminate as well to match

cretz (approved):
Looks great, merge whenever

PR #501, opened 2024-03-19
Cloud Login

### LINKED ISSUE #198: [Feature Request] SSO Authentication via the CLI

It is time consuming to distribute certificates to developers so they can use temporal CLI. It is also a potential security concern as these are long lived credentials stored on developers laptops.

Ideally you would be able to authenticate using the browser SSO to collect a short term token that would allow authentication to the Temporal Server. This would make the user experience much smoother. Similar to how gcloud auth or aws sso works.

cretz:
## What was changed

* Added `temporal cloud login` which by default opens browser and persists token to disk
* Added `temporal cloud logout` which by default removes file and opens browser
* Added `--cloud` to all client calls which takes the place of `--address`, and then wired up auth

⚠️ This PR is in draft because cloud data plane API does not accept web tokens at this time

## Checklist

1. Closes #198


chaptersix:
this PR appears to be stale. Closing for now.

PR #504, opened 2024-03-22
Add `temporal workflow fix-history-json` subcommand.

chronos-tachyon:
## What was changed
* New command: `temporal workflow fix-history-json`

## Why?
This command reads an event history JSON object using the `client.HistoryFromJSON` API, then serializes it back out using the `protojson` API.  `HistoryFromJSON` is backward compatible with both the standard protobuf JSON format and with GoGoProto's format, which differ in their handling of enum values.

## Checklist
1. Closes internal JIRA tickets SDK-1570 and OSS-1658, which do not seem to have equivalent issues in the public GitHub.

2. How was this tested:
I downloaded a sample event history JSON from my dev server, then ran it through a few conversions.

3. Any docs updates needed?
If there's a manual step in publishing changes to `commands.md`, then it will need to be run.


temporalcli/commandsmd/commands.md
```diff
@@ -635,6 +635,29 @@ temporal workflow execute
 Includes options set for [workflow start](#options-set-for-workflow-start).
 Includes options set for [payload input](#options-set-for-payload-input).
 
+### temporal workflow history: Interact directly with a workflow's event history.
+
+Workflow history commands use this syntax: `temporal workflow history COMMAND [ARGS]`.
+
+### temporal workflow history convert: Converts an event history between formats.
```
cretz, line 642: I wonder if this should be a command called `convert-history` instead of a new subcommand type

temporalcli/commandsmd/commands.md
```diff
@@ -635,6 +635,29 @@ temporal workflow execute
...
+		--input-file original.json \
+                --output-format json \
+		--output-file reserialized.json
+```
+
+Use the options listed below to change the command's behavior.
+
+#### Options
+
+* `--source-format` (string-enum) - Format of the input file.  Options: json, proto, pb, prototext, pbtxt.  Default: json.
```
cretz, line 656: I think the only use case is conversion from old to new, not for different proto formats. I know it can be seen as a "why not", but there's really only one Temporal history format, it just happens to be proto JSON.

I don't think this command really even needs to exist, but if it does, I wonder if we should call it `fix-history-format` or something.

temporalcli/commandsmd/commands.md
```diff
@@ -635,6 +635,29 @@ temporal workflow execute
...
+* `--source-file` (string) - Path to the input file.  Required.
+* `--target-format` (string-enum) - Format of the output file.  Options: json, proto, pb, prototext, pbtxt.  Default: json.
```
cretz, line 658: There is only one history format for Temporal (happens to be proto JSON). However there is a future where workflow export which wraps as proto binary may have use, but I don't think it's needed now.

temporalcli/commandsmd/commands.md
```diff
@@ -635,6 +635,29 @@ temporal workflow execute
...
+                --output-format json \
+		--output-file reserialized.json
+```
+
+Use the options listed below to change the command's behavior.
+
+#### Options
+
+* `--source-format` (string-enum) - Format of the input file.  Options: json, proto, pb, prototext, pbtxt.  Default: json.
+* `--source-file` (string) - Path to the input file.  Required.
```
cretz, line 657:
```suggestion
* `--source-file` (string) - Path to the input file. Required.
```

temporalcli/commands.workflow_history.go
```diff
@@ -0,0 +1,104 @@
...
+	switch c.SourceFormat.Value {
+	case historyFormatJSON:
+		history, err = client.HistoryFromJSON(bytes.NewReader(raw), client.HistoryJSONOptions{})
+		if err != nil {
+			return err
+		}
+
+	case historyFormatProtoText:
+		fallthrough
+	case historyFormatPBTXT:
```
cretz, line 38:
```suggestion
	case historyFormatProtoText, historyFormatPBTXT:
```

(and for other fallthrough cases in the PR)

temporalcli/commands.workflow_history.go
```diff
@@ -0,0 +1,104 @@
...
+	case "":
+		fallthrough
+	case "-":
+		_, err = cctx.Options.Stdout.Write(raw)
+		if err != nil {
+			return err
+		}
+
+	default:
+		err = os.WriteFile(c.TargetFile, raw, 0o666)
```
cretz, line 97:
```suggestion
		return os.WriteFile(c.TargetFile, raw, 0600)
```

`o` not needed (`0` prefix assumes octal), but can leave in if it reads better for you. Also, technically history can have some private data, wonder if we want owner-only.

temporalcli/commandsmd/commands.md
chronos-tachyon: Is there an existing command for extracting the history JSON using the CLI?  Because I couldn't find one and had to use the dev server web UI.  If we don't, I think we should add it as `temporal workflow history get`.  If we do, then I think you're probably right, but maybe we should make the history JSON extraction easier to find.

(`temporal workflow show -o json` *almost* does this, but it gives the decoded payloads, not the raw payloads, and `HistoryFromJSON` can't read JSON with decoded payloads.)

cretz: > Is there an existing command for extracting the history JSON using the CLI?

Yes, though poorly named `workflow show`

> temporal workflow show -o json almost does this, but it gives the decoded payloads, not the raw payloads, and HistoryFromJSON can't read JSON with decoded payloads.

Not following exactly, what do you mean decoded payloads in this case? To you mean applying a payload codec?

cretz: Oh, I think I found it, we should fix this for `workflow show` specifically. Today you need to call `temporal workflow show -o json --no-json-shorthand-payloads`.

temporalcli/commands.workflow_history.go
chronos-tachyon: I prefer `0o` because I think it's what C should have used for octal all along.

As for `666` vs `600`, I'm always a little cautious about overruling the user's umask, but if you think it's reasonable here (when `temporal workflow show -o json > file.json` would just use the umask) we can do that.

temporalcli/commands.workflow_history.go
chronos-tachyon: Done (in the one remaining use of fallthrough).

I kind of got into the habit of liberal use of fallthrough because I'm always thinking ahead to "when I have to convert this from switch-on-value to switch-on-first-matching-expression..."

temporalcli/commandsmd/commands.md
chronos-tachyon: Done.

temporalcli/commandsmd/commands.md
chronos-tachyon: I mean, fair.  I didn't have a specific use case in mind, I just thought maybe it would be nice to have a command to do that if needed.  Maybe the CLI isn't the right place for that, though, since it's meant for consumer use.

temporalcli/commandsmd/commands.md
chronos-tachyon: Ah, okay.  I went with `temporal workflow fix-history-json`.

temporalcli/commandsmd/commands.md
```diff
@@ -635,6 +635,21 @@ temporal workflow execute
 Includes options set for [workflow start](#options-set-for-workflow-start).
 Includes options set for [payload input](#options-set-for-payload-input).
 
+### temporal workflow fix-history-json: Updates an event history JSON file to the current format.
```
cretz, line 638: I like this naming. I hope we can remove it one day, heh.

temporalcli/commands.workflow_history.go
cretz: Works for me, I think what you have here is fine

cretz (commented):
LGTM. May want to make a change to PR title and description just for people scrolling by.

cretz (approved):

PR #508, opened 2024-03-27
Allow release GH workflow to be executed manually

cretz:
## What was changed

Allow releaser workflow to be run manually


cretz:
This needs to be added to `main`

PR #510, opened 2024-03-27
Replace release workflow with goreleaser from main

cretz:
## What was changed

Replace release GH workflow with goreleaser from main so that it will run on release and can be manually triggered. I suspect once cli-rewrite is on main we may make the name more generic to support things like docker publish

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -1,21 +1,25 @@
-name: Release
+name: goreleaser
 
 on:
+  workflow_dispatch:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - '**'
 
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
-          submodules: recursive
+          fetch-depth: 0
 
-      - name: Setup Go
-        uses: actions/setup-go@v4
+      - name: Set up Go
+        uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
+        with:
+          go-version-file: "go.mod"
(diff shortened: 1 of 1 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -1,21 +1,25 @@
-name: Release
+name: goreleaser
 
 on:
+  workflow_dispatch:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - '**'
```
josh-berry, line 7: If this is a regex, shouldn't it be `.*`?

.github/workflows/goreleaser.yml
```diff
@@ -38,7 +42,7 @@ jobs:
         run: echo "::set-output name=go::$(go version | cut -d ' ' -f 3)"
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v5
+        uses: goreleaser/goreleaser-action@336e29918d653399e599bfca99fadc1d7ffbc9f7 # v4.3.0
```
josh-berry, line 45: Any reason not to use v5 of goreleaser? (I feel like we should probably just bite the bullet and pull all the updates / work out all the kinks in the release process now with the RC but that's not a strong opinion)

josh-berry (approved):
lgtm, just a couple dumb questions.

.github/workflows/goreleaser.yml
cretz: I just copied `main` to keep it safe to run as main runs. I don't even like the file name much less the contents. We should revisit this (and include docker publish), but I want initial release to get done.

.github/workflows/goreleaser.yml
cretz: The doc train I followed from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs -> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore -> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet made me

PR #511, opened 2024-03-27
Make argument and flag use consistent across commands

### LINKED ISSUE #343: Make argument and flag use consistent across commands
Some commands have required flags—we should take a pass at every command and see if there are required flags that should instead be positional arguments.

Exceptions might be situations where it's not obvious which order arguments should be provided in, or what the arguments actually are when reading a particular CLI invocation (since script writers will probably appreciate having named flags for documentation purposes).

josh-berry:
Closes #343. See the issue for the full discussion and rationale behind the specific changes made.

### DIFF
--- a/temporalcli/commands.env.go
+++ b/temporalcli/commands.env.go
@@ -13,7 +13,7 @@ func (c *TemporalEnvCommand) envNameAndKey(cctx *CommandContext, args[] string, 
 		cctx.Logger.Warn("Arguments to env commands are deprecated; please use -E and/or -k instead.")
 
 		if (c.Parent.Env != "default" || keyFlag != "") {
-			return "", "", fmt.Errorf("cannot specify both an argument and -E/-k; please use -e and/or -k")
+			return "", "", fmt.Errorf("cannot specify both an argument and -E/-k; please use -E and/or -k")
 		}
 
 		keyPieces := strings.Split(args[0], ".")
--- a/temporalcli/commands.env.go
+++ b/temporalcli/commands.env.go
@@ -10,10 +10,10 @@ import (
 
 func (c *TemporalEnvCommand) envNameAndKey(cctx *CommandContext, args[] string, keyFlag string) (string, string, error) {
 	if len(args) > 0 {
-		cctx.Logger.Warn("Arguments to env commands are deprecated; please use -E and/or -k instead.")
+		cctx.Logger.Warn("Arguments to env commands are deprecated; please use --env and --key (or -k) instead")
 
 		if (c.Parent.Env != "default" || keyFlag != "") {
-			return "", "", fmt.Errorf("cannot specify both an argument and -E/-k; please use -E and/or -k")
+			return "", "", fmt.Errorf("cannot specify both an argument and flags; please use flags instead")
 		}
 
 		keyPieces := strings.Split(args[0], ".")
--- a/temporalcli/commands.env.go
+++ b/temporalcli/commands.env.go
@@ -8,36 +8,66 @@ import (
 	"github.com/temporalio/cli/temporalcli/internal/printer"
 )
 
-func (c *TemporalEnvDeleteCommand) run(cctx *CommandContext, args []string) error {
-	keyPieces := strings.Split(args[0], ".")
-	if len(keyPieces) > 2 {
-		return fmt.Errorf("env property key to delete cannot have more than one dot")
+func (c *TemporalEnvCommand) envNameAndKey(cctx *CommandContext, args[] string, keyFlag string) (string, string, error) {
+	if len(args) > 0 {
+		cctx.Logger.Warn("Arguments to env commands are deprecated; please use -E and/or -k instead.")
+
+		if (c.Parent.Env != "default" || keyFlag != "") {
+			return "", "", fmt.Errorf("cannot specify both an argument and -E/-k; please use -e and/or -k")
+		}
+
+		keyPieces := strings.Split(args[0], ".")
+		switch len(keyPieces) {
+		case 0:
+			return "", "", fmt.Errorf("no env or property name specified")
+		case 1:
+			return keyPieces[0], "", nil
+		case 2:
+			return keyPieces[0], keyPieces[1], nil
+		default:
+			return "", "", fmt.Errorf("property name may not contain dots")
+		}
 	}
-	// Env must be present (but flag itself doesn't have to be)
-	env, ok := cctx.EnvConfigValues[keyPieces[0]]
+
+	if strings.Contains(keyFlag, ".") {
+		return "", "", fmt.Errorf("property name may not contain dots")
+	}
+
+	return c.Parent.Env, keyFlag, nil
+}
+
+func (c *TemporalEnvDeleteCommand) run(cctx *CommandContext, args []string) error {
+	envName, key, err := c.Parent.envNameAndKey(cctx, args, c.Key)
+	if err != nil {
+		return err
+	}
+
+	// Env must be present
+	env, ok := cctx.EnvConfigValues[envName]
 	if !ok {
-		return fmt.Errorf("env %q not found", keyPieces[0])
+		return fmt.Errorf("env %q not found", envName)
 	}
 	// User can remove single flag or all in env
-	if len(keyPieces) > 1 {
-		cctx.Logger.Info("Deleting env property", "env", keyPieces[0], "property", keyPieces[1])
-		delete(env, keyPieces[1])
+	if key != "" {
+		cctx.Logger.Info("Deleting env property", "env", envName, "property", key)
+		delete(env, key)
 	} else {
-		cctx.Logger.Info("Deleting env", "env", keyPieces[0])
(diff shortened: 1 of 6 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -49,7 +49,7 @@ This document has a specific structure used by a parser. Here are the rules:
 
 #### Options
 
-* `--env` (string) - Environment to read environment-specific flags from. Default: default. Env: TEMPORAL_ENV.
+* `--env`, `-E` (string) - Environment to read environment-specific flags from. Default: default. Env: TEMPORAL_ENV.
```
cretz, line 52: Not sure we should add a new global alias for this. I think the previous way of changing environment from default was good enough.

temporalcli/commands.env.go
```diff
@@ -8,45 +8,75 @@ import (
 	"github.com/temporalio/cli/temporalcli/internal/printer"
 )
 
+func (c *TemporalEnvCommand) envNameAndKey(cctx *CommandContext, args[] string, keyFlag string) (string, string, error) {
+	if len(args) > 0 {
+		cctx.Logger.Warn("Arguments to env commands are deprecated; please use -E and/or -k instead.")
+
+		if (c.Parent.Env != "default" || keyFlag != "") {
+			return "", "", fmt.Errorf("cannot specify both an argument and -E/-k; please use -e and/or -k")
```
cretz, line 16: Should not assume user is using aliases. And, as often happens with these aliases, this statement is unclear what each is referring to and there's an typo (i.e. `-e` does not exist).

temporalcli/commandsmd/commands.md
```diff
@@ -139,61 +139,86 @@ For future reference, provide a reason for terminating the Batch Job.
...
 
 <!--
-* exact-args=2
+* maximum-args=2
 -->
 
+#### Options
+
+* `--key`, `-k` (string) - The name of the property.
+* `--value`, `-v` (string) - The value to set the property to.
```
cretz, line 220:
```suggestion
* `--key`, `-k` (string) - The name of the property. Required.
* `--value`, `-v` (string) - The value to set the property to. Required.
```

Also, I don't think there's value in adding aliases (but we have been against over-aliasing commands in general in this project, especially in cases where they may be used for something else in another command)

temporalcli/commands.operator_namespace.go
```diff
@@ -14,8 +14,25 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
+func (c *TemporalOperatorCommand) getNSFromFlagOrArg0(cctx *CommandContext, args []string) (string, error) {
+	if len(args) > 0 && c.Namespace != "default" {
+		return "", fmt.Errorf("namespace was provided as both an argument (%s) and a flag (-n %s); please specify namespace only with -n", args[0], c.Namespace)
+	}
+
+	if len(args) > 0 {
+		cctx.Logger.Warn("Passing the namespace as an argument is now deprecated; please switch to using -n instead.\n")
```
cretz, line 23: Should not assume use of alias IMO

temporalcli/commands.operator_namespace.go
```diff
@@ -14,8 +14,25 @@ import (
...
+		return args[0], nil
+	} else {
+		return c.Namespace, nil
+	}
```
cretz, line 27:
```suggestion
	}
	return c.Namespace, nil
```

(pedantic, no need to change)

temporalcli/commandsmd/commands.md
josh-berry: The alternative is to be inconsistent and have a different `--env` flag for `temporal env`. I think this is worse than simply adding a global alias.

I also like the global alias because it uses less visual and cognitive space for what is essentially contextual information. So overall I feel pretty strongly about keeping it.

temporalcli/commands.env.go
josh-berry: Oops, thought I caught all the `-e`s when I changed to `-E`—will fix.

I don't think any assumption is being made here—we're _telling_ the user to use `-E` and `-k`. If this is confusing (which it probably will not be for most people, since they can guess that `E` stands for `environment`), they can always look at `--help`.

temporalcli/commandsmd/commands.md
josh-berry: These cannot be required until we remove the old/deprecated way of specifying the env. If they are, the old way will stop working because Cobra will throw errors before even reaching command code.

temporalcli/commands.operator_namespace.go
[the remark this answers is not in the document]
josh-berry: Whups, I think this may have been a Print() before. Will fix.

temporalcli/commandsmd/commands.md
josh-berry: To more directly address your point about aliases: I think requiring users to spell out `--key` and `--value` is unfriendly. Since we could not reach consensus on using positionals consistently, I think having short aliases is the next least-bad option, and so I feel strongly about keeping them.

temporalcli/commandsmd/commands.md
cretz: I don't think it'll get used very much and I think the bar for our second-ever global alias should be a bit higher

temporalcli/commandsmd/commands.md
cretz: > These cannot be required until we remove the old/deprecated way of specifying the env

Ah, my mistake, sorry

> I think having short aliases is the next least-bad option, and so I feel strongly about keeping them

To be consistent with ourselves will we now need to apply this logic for short aliases for every command? Or is this strong feeling only about these new changes? We don't have that many aliases (even being a required option is not enough of a reason).

temporalcli/commandsmd/commands.md
josh-berry: This specific one. I think the other options are much less likely to be used, whereas I could see `-E` being used quite commonly/consistently, especially in ops scenarios where an operator is dealing with multiple environments. (I also agree with having `-o` for the output format.)

temporalcli/commandsmd/commands.md
josh-berry: > To be consistent with ourselves will we now need to apply this logic for short aliases for every command?

No, I think there is some subjectivity as to which options are commonly-used/important enough to merit short-form aliases. It can also sometimes be a deliberate design decision to _not_ have a short-form alias (e.g. for something obscure or dangerous).

temporalcli/commands.operator_namespace.go
josh-berry: If you feel strongly about this, I can reword (I suspect most people will use the short-form though).

temporalcli/commandsmd/commands.md
cretz: This flag has been here for years and years sans alias, I don't think it's that common, it has value forcing a user to be clear they are changing environments, and I wouldn't expect an alias to change profile/env/context any more than I'd expect `aws` CLI to have an alias for `--profile` or `kubectl`/`docker` to have an alias for `--context`.

cretz: > You are ignoring the fact that we are now using it in a different way.

Not if we're discussing the global alias which is my primary concern. While I don't like it for `temporal env` subcommands either, I think we can agree we are _not_ using it in a different way for the vast majority of commands.

> And I would find [...] annoyingly verbose

This is subjective and not shared by the other example binaries I linked there, so at the least they didn't agree. This new global alias applies to your words from earlier:

> It can also sometimes be a deliberate design decision to not have a short-form alias (e.g. for something obscure or _dangerous_)

Emphasis mine. It is dangerous to change env/profile/context. Regardless, I think `aws` or `kubectl` or `docker` or those similar tools with contextual pieces like this can provide guidance here.

temporalcli/commands.operator_namespace.go
cretz: I think our docs should always show long form. We used to go overboard with aliases where you could even have the command aliased, so you could `tctl w l`. People fought to keep that lack of clarity too. Even with the aliases, I hope our documentation (and error messages) don't require alias lookup.

temporalcli/commandsmd/commands.md
josh-berry: > Regardless, I think `aws` or `kubectl` or `docker` or those similar tools with contextual pieces like this can provide guidance here.

You are correct about `aws`, however:

- `kubectl` has `-s` for which API server to connect to, though I agree `--context` has no short-form
- `docker` has `-c` for which context to use and `-H` for which host to use

So the record is rather mixed here.

Regardless, we've exhausted my time budget on this, so I will concede because it's easier to add it in later than it is to remove it if we later decide we don't want it.

cretz: :+1: I will also concede the `-k` and `-v` since the blast radius of your change there is limited if you want to add those back. My bigger concern was this new global alias we were adding.

cretz (approved):

PR #516, opened 2024-04-02
Fix racy batch test

josh-berry:
This test occasionally fails because the batch job completes before the test has a chance to check its status. Allow for this case in the test.

### DIFF
--- a/temporalcli/commands_test.go
+++ b/temporalcli/commands_test.go
@@ -61,6 +61,10 @@ func (h *CommandHarness) ContainsOnSameLine(text string, pieces ...string) {
 	h.NoError(AssertContainsOnSameLine(text, pieces...))
 }
 
+func (h *CommandHarness) AnyLineMatchesRegexp(text string, pattern string) {
+	h.NoError(AssertAnyLineMatchesRegexp(text, pattern))
+}
+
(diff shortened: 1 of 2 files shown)


josh-berry:
hm... I'm realizing this is incomplete; the terminate test also has the same problem. It's less obvious how to fix that one, though, because success is indistinguishable from failure; if the batch job is already completed, it can't be terminated. (I'm seeing both failures on my Windows box pretty reliably.)

I think this one can still be merged as-is, but someone (maybe/probably me) will need to do a follow-up PR for the terminate test.

temporalcli/commands_test.go
```diff
@@ -61,6 +61,10 @@ func (h *CommandHarness) ContainsOnSameLine(text string, pieces ...string) {
 	h.NoError(AssertContainsOnSameLine(text, pieces...))
 }
 
+func (h *CommandHarness) AnyLineMatchesRegexp(text string, pattern string) {
```
cretz, line 64: Can you just use `h.Regexp`?

josh-berry: I... did not realize that was a thing, oops. Yes, will fix.

cretz (approved):

PR #507, opened 2024-03-27
Minor refactor and fixes to workflow trace

sebneira:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

PR #520, opened 2024-04-03
Add workflow trace

sebneira:
## What was changed

Ported over trace to the new cli rewrite. 

Added a couple of fixes to get this to work:
- Moved all the tracer printing logic to the `WorkflowTracer` (see https://github.com/temporalio/cli/pull/507 for the isolated changes). This was the initial fix I was working on. This should remove the issues when state == nil and the weirdness with the `setupUpdateChannels`.
- Removed ExecutionState's `GetDuration` and `GetStartTime` time reference since the events now use `durationpb.Duration` and  `timestamppb.Timestamp` and the conversion to non-pointer values makes more sense. Adjusted the templates to present this data correctly.

## Why?

Ported the command to the cli rewrite and provided fixes for the nil pointer issues reported by Chad.

## Checklist

1. Closes 

2. How was this tested:

All unit tests are running and I've run the tool against a couple different workflow executions with temporalite.

3. Any docs updates needed?


cretz:
This is great! Will review when you mark "ready for review"

temporalcli/commands.gen.go
cretz, line 1: This file is an auto-generated file from `commandsmd/commands.md` but I don't see that changed in this PR. Take a look at `CONTRIBUTING.md`. Basically you update that file following rules specified at the top of that file, and then regen this file.

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"github.com/fatih/color"
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"github.com/temporalio/cli/temporalcli/internal/trace"
+	"go.temporal.io/sdk/client"
+)
+
+func (c *TemporalWorkflowTraceCommand) run(cctx *CommandContext, _ []string) error {
```
cretz, line 11:
```suggestion
func (c *TemporalWorkflowTraceCommand) run(cctx *CommandContext, _ []string) error {
	if cctx.JSONOutput {
		return fmt.Errorf("JSON output not supported for trace command")
	}
```

Let's disable JSON for this command for now which also means you can remove JSON tags from your struct below

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,67 @@
...
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	ctx := cctx.Context
```
cretz, line 18: This is not needed, `cctx` satisfies the context interface by embedding context, so it can be used directly as a context

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,67 @@
...
+	cctx.Printer.Println(color.MagentaString("Execution summary:"))
+
+	_ = cctx.Printer.PrintStructured(workflowSummary{
+		WorkflowId: info.GetExecution().GetWorkflowId(),
+		RunId:      info.GetExecution().GetRunId(),
+		Type:       info.GetType().GetName(),
+		Namespace:  c.Parent.Namespace,
+		TaskQueue:  info.GetTaskQueue(),
+	}, printer.StructuredOptions{})
+	_, _ = fmt.Println()
```
cretz, line 64: While I know we use stdout in the internal/trace stuff, would like to avoid doing so in the command code, can just use the printer to print a line.

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err = c.printWorkflowSummary(cctx, cl, c.WorkflowId, c.RunId); err != nil {
+		return err
+	}
+	_, err = trace.PrintWorkflowTrace(ctx, cl, c.WorkflowId, c.RunId, opts)
+
+	return err
+}
+
+type workflowSummary struct {
```
cretz, line 38:
```suggestion
type workflowTraceSummary struct {
```

This package is shared by many commands, so `workflowSummary` specific to trace is too general of a name

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,67 @@
+package temporalcli
+
+import (
+	"fmt"
+	"github.com/fatih/color"
```
cretz, line 5:
```suggestion
	"fmt"

	"github.com/fatih/color"
```

Hrmm, usually go fmt would separate these

temporalcli/commands.workflow_trace.go
cretz, line 11: Can you have a high-level test or two that invokes the command using the harness (can see other commands for how this is done)

temporalcli/internal/trace/execution_state.go
```diff
@@ -0,0 +1,625 @@
+// The MIT License
```
cretz, line 1: Can probably get rid of these license headers

cretz (commented):
Looks pretty good. Not too concerned with the details of the `internal/trace` package, so didn't review that closely.

temporalcli/commandsmd/commands.md
```diff
@@ -1025,6 +1025,23 @@ Use the options listed below to change the behavior of this command.
...
+### temporal workflow trace: Terminate Workflow Execution by ID or List Filter.
+
+The `temporal workflow trace` command display the progress of a [Workflow Execution](/concepts/what-is-a-workflow-execution) and its child workflows with a trace.
+This view provides a great way to understand the flow of a workflow.
+
+Use the options listed below to change the behavior of this command.
+
+#### Options
+
+* `--fold` (string) - Statuses for which Child Workflows will be folded in (this will reduce the number of information fetched and displayed). Case-insensitive and ignored if no-fold supplied. Available values: running, completed, failed, canceled, terminated, timedout, continueasnew. Default: completed,canceled,terminated.
```
cretz, line 1037: If we supported it, would suggest `string-enum[]` as the type, but in the meantime, might as well just have `string[]` as the type. This allows them to specify `--fold` multiple times instead of using commas. However, default not supported for `string[]`. You can either add support in `commandsmd/code.go` or just have something like "Default is completed, canceled, terminated" which won't trigger the parser and let you set the default at runtime.

sebneira: I'm happy to do this is a later iteration so we can move this (already pretty big thing) forward. Does this work for you?

cretz: Sure, but it will be a backwards-incompatible alteration at that time unfortunately.

temporalcli/commands.workflow_trace.go
cretz: Bump, I do think we want at least one test confirming the call from the CLI POV works

temporalcli/commandsmd/commands.md
josh-berry: Can you pass multiple statuses to a single instance of `--fold`? The help text says "Statuses", which makes me think this is possible, but we should make it explicit in the help text that commas are needed.

sebneira: Could I just not use the flag for now and we can figure out how to work this out in a future PR? I think it's a non-trivial decision that we would benefit from putting some good thinking behind (and not delay this PR).

cretz: Will defer to @josh-berry. I do think however it is fairly trivial to just change this to `string[]` and wait until you see an empty slice at runtime to set the default 3 values.

josh-berry: Not having the flag and adding it later is fine by me; so is accepting `[]string`. The only thing I'm not thrilled about is making a breaking change (i.e. commas now and `[]string` later).

cretz: :+1: @sebneira - can you make this type `string[]` and set the defaults in code on empty slice when no-fold is not set? You can still write "Defaults are completed, canceled, and terminated." and it won't be read as a special option like `Default:` is.

temporalcli/commands.workflow_trace.go
cretz: They're great, thanks!

temporalcli/commandsmd/commands.md
sebneira: Thanks for the suggestion. I updated the PR to include the `string[]` flag for fold which defaults to the completed, canceled and terminated if the flag is empty.

Also, as I was moving the options struct around it kinda dawned on me that the pkg made more sense as `tracer` rather than `trace`. Let me know if this doesn't make sense.

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/temporalio/cli/temporalcli/internal/tracer"
+
+	"go.temporal.io/api/enums/v1"
+
+	"github.com/fatih/color"
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"go.temporal.io/sdk/client"
+)
+
+var foldFlag = map[string]enums.WorkflowExecutionStatus{
```
cretz, line 18:
```suggestion
var workflowTraceFoldFlags = map[string]enums.WorkflowExecutionStatus{
```

This package is shared by many commands, need to qualify global names

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,133 @@
...
+			enums.WORKFLOW_EXECUTION_STATUS_TERMINATED,
+		}, nil
+	}
+
+	// parse flags
+	var values []enums.WorkflowExecutionStatus
+	for _, flag := range foldFlags {
+		status, ok := foldFlag[flag]
+		if !ok {
+			return nil, fmt.Errorf("fold status \"%s\" not recognized", flag)
```
cretz, line 43:
```suggestion
			return nil, fmt.Errorf("fold status %q not recognized", flag)
```

temporalcli/commands.workflow_trace_test.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.temporal.io/api/enums/v1"
+
+	"go.temporal.io/sdk/workflow"
+
+	"go.temporal.io/sdk/client"
```
cretz, line 12: No need for the extra newlines

cretz (approved):
Just pedantic stuff (would like to see fixed since easy though if you don't mine). Marking approved, LGTM overall.

temporalcli/commands.workflow_trace.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"fmt"
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/temporalio/cli/temporalcli/internal/tracer"
+
+	"go.temporal.io/api/enums/v1"
+
```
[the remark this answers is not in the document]
sebneira: Right, I was expecting goimports to take care of this 🤦

PR #525, opened 2024-04-05
Exit zero if help/version is explicitly requested

### LINKED ISSUE #495: [Bug] --version and --help are exiting with failure code
```
$ /Users/dan/src/temporalio/cli/temporal-new --version > /dev/null; echo $?
unknown command
1

$ /Users/dan/src/temporalio/cli/temporal-new --help > /dev/null; echo $?
unknown command
1
```

josh-berry:
We want a non-zero exit when the user invokes `temporal` with an unknown command/subcommand, or when invoked with no arguments.  But if the user explicitly asks for help or version information, we should print it and return success.

Closes #495.

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -322,8 +322,23 @@ func Execute(ctx context.Context, options CommandOptions) {
 	if err != nil {
 		cctx.Options.Fail(err)
 	}
-	// If no command ever actually got run, exit nonzero
+
+	// If no command ever actually got run, exit nonzero with an error.  This is
+	// an ugly hack to make sure that iff the user explicitly asked for help, we
+	// exit with a zero error code.  (The other situation in which help is
+	// printed is when the user invokes an unknown command--we still want a
+	// non-zero exit in that case.)  We should revisit this if/when the
+	// following Cobra issues get fixed:
+	//
+	// - https://github.com/spf13/cobra/issues/1156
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

cretz (approved):
LGTM, but don't want to close #495 without also addressing `--version` or say we don't need to address it

temporalcli/commands.go
```diff
@@ -360,6 +361,25 @@ func (c *TemporalCommand) initCommand(cctx *CommandContext) {
...
+
+	// Ugly hack to make sure that iff the user explicitly asked for help, we
+	// exit with a zero error code.  (The other situation in which help is
+	// printed is when the user invokes an unknown command--we still want a
+	// non-zero exit in that case.)  We should revisit this if/when the
+	// following Cobra issues get fixed:
+	//
+	// - https://github.com/spf13/cobra/issues/1156
+	// - https://github.com/spf13/cobra/issues/706
+	inner_help := c.Command.HelpFunc()
```
cretz, line 373:
```suggestion
	innerHelp := c.Command.HelpFunc()
```

Go does not use snake_case

josh-berry:
> LGTM, but don't want to close #495 without also addressing `--version` or say we don't need to address it

🤦 It even listed `--version` in the ticket, oops. Will fix and re-post.

cretz (approved):

PR #537, opened 2024-04-22
Update to latest Server, UIServer, and Go SDK

josh-berry:
Needed to pick up the fix for CVE-2024-2435.

NOTE: We're only going up to go-sdk 1.26.0 and api 1.29.2, because newer
versions introduce compile errors in the server. (Looks like something
related to versioning in the frontend but I didn't look closely.)


josh-berry:
Turns out there is a backward-incompatible change in the history JSON output, so we can't do this as part of a point release. Instead, we've decided to release the fix for this CVE as part of the upcoming 0.12 release (which is in RC right now).

PR #540, opened 2024-04-29
Resurrect install.sh

josh-berry:
We're apparently serving `https://temporal.download/cli.sh` out of main.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,546 @@
+#!/bin/sh
+# shellcheck shell=dash
+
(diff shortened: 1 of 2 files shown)


cretz (approved):
Can you confirm this shell script still works with latest release?

josh-berry:
> Can you confirm this shell script still works with latest release?

Works on Mac, yes. Also means it's probably good enough for Linux as well (the script does nothing complicated).

PR #551, opened 2024-05-06
Batch rate limit

stephanos:
## What was changed

Added support for passing `--rps` flag to batch operations.

ℹ️ FYI I've only added a single test for TerminateWorkflow, even though there are 3 batch APIs. Reason being that the API is the same and the only difference is the "operation". Let me know if I should add it to the others as well.

❓ Note sure why I'm seeing the CLA request on my PR.

## Checklist

1. Closes 

2. How was this tested: Added new tests.

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.workflow_test.go
+++ b/temporalcli/commands.workflow_test.go
@@ -1,18 +1,20 @@
 package temporalcli_test
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"math/rand"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
 	"go.temporal.io/api/enums/v1"
-	"go.temporal.io/api/history/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/workflow"
+	"google.golang.org/grpc"
 )
 
 func (s *SharedServerSuite) TestWorkflow_Signal_SingleWorkflowSuccess() {
@@ -277,9 +279,9 @@ func (s *SharedServerSuite) TestWorkflow_Terminate_BatchWorkflow_Ratelimit_Missi
 }
 
 func (s *SharedServerSuite) TestWorkflow_Terminate_BatchWorkflowSuccess_Ratelimit() {
-	events, _ := s.testTerminateBatchWorkflow(2, 1, false)
-	delay := events[1].EventTime.AsTime().Sub(events[0].EventTime.AsTime()).Abs()
-	s.InDelta(delay, 1*time.Second, float64(100*time.Millisecond))
+	var rps float32 = 1
+	req, _ := s.testTerminateBatchWorkflow(2, rps, false)
+	s.Equal(rps, req.MaxOperationsPerSecond)
 }
 
 func (s *SharedServerSuite) TestWorkflow_Terminate_BatchWorkflowSuccess_JSON() {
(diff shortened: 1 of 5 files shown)


temporalcli/commands.workflow_test.go
```diff
@@ -261,26 +262,41 @@ func (s *SharedServerSuite) TestWorkflow_Terminate_SingleWorkflowSuccess_WithRea
...
+		"--address", s.Address(),
+		"--rps", "5",
+	)
+	s.Error(res.Err, "must set either workflow ID or query")
+}
+
+func (s *SharedServerSuite) TestWorkflow_Terminate_BatchWorkflowSuccess_Ratelimit() {
+	events, _ := s.testTerminateBatchWorkflow(2, 1, false)
+	delay := events[1].EventTime.AsTime().Sub(events[0].EventTime.AsTime()).Abs()
+	s.InDelta(delay, 1*time.Second, float64(100*time.Millisecond))
```
stephanos, line 282: In my tests it hasn't been exactly 1s, so I added some delta here.

temporalcli/commands.workflow_test.go
```diff
@@ -308,6 +324,7 @@ func (s *SharedServerSuite) testTerminateBatchWorkflow(json bool) *CommandResult
 	// Send batch terminate with a "y" for non-json or "--yes" for json
 	args := []string{
 		"workflow", "terminate",
+		"--rps", fmt.Sprint(rps),
```
stephanos, line 327: This introduces a delay in the test suite of up to 1s. Not sure how to avoid that while still testing the Server end-to-end.

temporalcli/commandsmd/commands.md
```diff
@@ -911,6 +911,7 @@ Use the options listed below to change reset behavior.
 * `--build-id` (string) - Only used if type is BuildId. Reset the first workflow task processed by this build id. Note that by default, this reset is allowed to be to a prior run in a chain of continue-as-new.
 * `--query`, `-q` (string) - Start a batch reset to operate on Workflow Executions with given List Filter.
 * `--yes`, `-y` (bool) - Confirm prompt to perform batch. Only allowed if query is present.
+* `--rps` (float) - Limit batch's requests per second. Only allowed if query is present.
```
cretz, line 914: Can you update the top of this file to include this new data type in the bullet that lists the data types supported? It's our admittedly-informal "spec".

temporalcli/commands.workflow_test.go
cretz: We may have to accept some things cannot reasonably tested in CLI without slowing all tests down.

temporalcli/commands.workflow_test.go
cretz: This is the kind of test that grows flaky with Go runner slowness I've found. Expecting this margin of error is a bit rough. I think for this you may only need to test that it's set on the outbound request (e.g. w/ gRPC interceptor maybe), you don't need to test server behavior (that's for the server to test).

cretz (commented):
LGTM. One minor suggestion and concerns about using time-based expectations in tests

cretz (approved):
Looks great!

tlalfano:
Let's make sure this is in the next CLI release.

antlai-temporal:
@stephanos I think the cli-rewrite branch is dead, main becoming the new target branch for the last two weeks, you may want to merge in main instead... @cretz Can you confirm?
I'm going to use your float changes for one of my flags, thanks for doing that!

PR #564, opened 2024-05-15
get-free-port: prevent duplicate ports on Linux

### LINKED ISSUE #543: SIGSEGV when using `--http-port` option for dev server and port is already in use
### Describe the bug

With CLI version 0.12.0, I started a local development server with the `--http-port` option. Since I had already started one in another terminal window, I expected this to fail with a "port already in use" message. While it did do this, that message was followed by `panic: runtime error: invalid memory address or nil pointer dereference`, `[signal SIGSEGV: segmentation violation code=0x2 addr=0x20 pc=0x1023b3680]`, and a stack trace that obscured the "port already in use" message until I scrolled through 100 lines of output.

### Minimal Reproduction

1. Install Temporal CLI version 0.12.0 and ensure that it's foremost in your executable path by running `temporal --version`
2. In one terminal, run `temporal server start-dev --http-port 8234`
3. In another terminal, run `temporal server start-dev --http-port 8234`
4. Observe that the second command fails immediately, as described above.

### Environment/Versions

- OS and processor: M2 Mac running macOS Ventura 13.6.6 (22G630)
- Temporal Version: CLI = temporal version 0.12.0 (server 1.23.0) (ui 2.26.2)
- Are you using Docker or Kubernetes or building Temporal from source? No

### Additional context

N/A

### LINKED ISSUE #550: [Feature Request] Fix free-port assignment to actually try to connect

We assumed incorrectly that OSs won't give out recently bound ports, but tests reveal it may. Fix https://github.com/temporalio/cli/blob/main/temporalcli/devserver/freeport.go to:

* Just have one `GetFreePort() (int, error)` call and none of that extra port-provider stuff
* Make the `GetFreePort()` listener also accept a socket connection before closing both the listener and socket and returning the port

Probably don't even need a separate file anymore, this can be a simple single-method utility with the rest of the dev server stuff

mjameswh:
## What changed

- When assigning ephemeral ports for use by the Temporal Server, on all platforms except MacOS and Windows, immediately open a connection to that port, and force the connection into `TIME_WAIT` state. This avoids various race conditions that could happen due to the port being reallocated between the moment the port is obtained and the moment the server actually binds to that port. Resolves #550.

- Add checks to assert that ports provided by the user are indeed free; this would previously cause a SIGSEGV. Resolves #543.
  - Note that these checks are subject to race conditions (that's essentially a check-before-use sequence), though such race occurrences are very unlikely for ports numbers that have been specifically requested by the user.

- Added tests to confirm that it is possible to start multiple dev server concurrently.


temporalcli/internal/freeport/freeport.go
cretz, line 1: Do we need a new package instead of just putting this file back where the file that was deleted was (and add the test file next to it)? I don't mind that much though, but if you do move, calling `devserver.CheckPortFree` from the command is ok. Your choice.

temporalcli/internal/freeport/freeport.go
```diff
@@ -0,0 +1,101 @@
+package freeport
+
+import (
+	"fmt"
+	"net"
+	"runtime"
+)
+
+/**
```
cretz, line 9: This is not the common format of Godoc, it's just `// `-prefixed lines

cretz (commented):
LGTM. Only needed comment is fixing Godoc format.

temporalcli/internal/freeport/freeport.go
mjameswh: No, I can move it back. I originally moved it to a different package when I introduced `freeport_windows.go` vs `freeport_darwin.go` vs …, but since I reduced all of that to a single file, there no longer a justification for this.

temporalcli/devserver/freeport.go
```diff
@@ -1,84 +1,95 @@
...
+// within a few seconds.
+//
+// On Unix-based systems, binding to the port returned by this function requires
+// setting the `SO_REUSEADDR` socket option (Go already does that by default,
+// but other languages may not); otherwise, the OS may fail with a message such
+// as "address already in use". Windows default behavior is already appropriate
+// in this regard; on that platform, `SO_REUSEADDR` has a different meaning and
+// should not be set (setting it may have unpredictable consequences).
+func GetFreePort(host string) (int, error) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
```
cretz, line 24:
```suggestion
	l, err := net.Listen("tcp", host+":0")
```

I wonder if this makes more sense. Not that it matters much.

cretz (approved):

PR #565, opened 2024-05-16
Provide reachability info with `describe` task queue for the new versioning API (Backward incompatible)

antlai-temporal:
## What was changed

Provides an enhanced mode (by default) to `temporal task-queue describe` that provides task reachability information 
for Build IDs. The goal is to facilitate retiring workers with old code that use the new worker versioning API.

The old behavior of `describe` is available using the `legacy-mode` flag.


temporalcli/commands.taskqueue.go
```diff
@@ -11,31 +11,237 @@ import (
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/taskqueue/v1"
 	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/server/common/tqid"
 )
 
+const unversioned = "UNVERSIONED"
```
cretz, line 18: This package is shared by many commands. May want to make the top-level constructs clearly specific to task queue work. For some of the functions this can maybe be easier putting them as methods on the command instead of top-level functions, but up to you.

temporalcli/commands.taskqueue.go
```diff
@@ -11,31 +11,237 @@ import (
...
+		if err != nil {
+			return fmt.Errorf("displaying reachability failed: %w", err)
+		}
+
+		cctx.Printer.Println(color.MagentaString("Pollers:"))
+		return cctx.Printer.PrintStructured(descRows.Pollers, printer.StructuredOptions{Table: &printer.TableOptions{}})
+	}
+
+	// json output
+	return cctx.Printer.PrintStructured(descRows, printer.StructuredOptions{})
```
cretz, line 172: I think it may be clearer to just print the proto describe response when in JSON mode. The printer is already prepared to properly print protos as JSON. No need for a bunch of custom JSON objects. And we usually find the protos are stable and easy to point users to if they want to see which fields would be present in their JSON.

temporalcli/commands.taskqueue.go
```diff
@@ -11,31 +11,237 @@ import (
...
+	}
+
+	return taskQueueDescriptionType{
+		Reachability: rRows,
+		Pollers:      pRows,
+	}, nil
+}
+
+func printTaskQueueDescription(cctx *CommandContext, taskQueueDescription client.TaskQueueDescription) error {
+	descRows, err := descriptionToRows(taskQueueDescription)
```
cretz, line 155: Arguably limited value breaking these functions into other package-level functions that are only ever used once, but no biggie

temporalcli/commands.taskqueue.go
cretz, line 18: Where is this constant used?

antlai-temporal: good point

antlai-temporal: The unversioned queue is represented with an empty string, this is just a constant to define a printable name. I'll make it more specific...

temporalcli/commands.taskqueue.go
antlai-temporal: Talking with @ShahabT there are some things that we don't want to include, like `WorkerVersionCapabilities` that are mostly redundant, and sometimes expose Build IDs derived from code hashes that could be confusing. 
Also, I thought that a "row friendly" layout could be easier if they want to add results to a traditional DB, and more compatible with the text view...

temporalcli/commands.taskqueue.go
antlai-temporal: I tend to prefer smallish functions, but clearly they don't need to be package-level. I could add then as methods to the `TemporalTaskQueueDescribeCommand` object, but they are pure functions, ... What is the convention in Go for pure functions that are private to a file?

cretz: There is no exact convention. Commonly they are inlined if they aren't very big and only called in one place.

temporalcli/commands.taskqueue.go
cretz: Our HTTP API is going to return that other JSON. But I don't mind if we want this separate JSON format for just this call, I have no strong opinion.

cretz (approved):
Can you make it clear in the description (and probably summary/title) that is this is a backwards incompatible change so that the person doing the release will know to call it out?

ShahabT:
Reachability could be an expensive operation and many time the callers may only be interested in pollers only. I think we should make it disabled by default and only report it when user sets a flag such as `--report-reachability` or similar.

ShahabT:
Also similarly, one may not want poller info so ideally there should be a way to disable that such as `--report-pollers=false`.

Note that a new section will be added soon (BacklogInfo). That should also have a flag (maybe true by default).

antlai-temporal:
> Reachability could be an expensive operation and many time the callers may only be interested in pollers only. I think we should make it disabled by default and only report it when user sets a flag such as `--report-reachability` or similar.

I was thinking whether to export the flags or not. I thought it was a bit confusing not providing the info by default, and this is the cli, so I'm not expecting lots of calls using it, and create a real overhead problem. I'm assuming that most demanding users will use the SDK directly, and they can apply there the flags. Also, even with report off it provides info in json mode (e.g., similar to no pollers found), something that is confusing.

Maybe we can revisit after we get more experience with the pre-release?

cretz:
:+1: Would like to know how expensive. It is a common use case for people to poll reachability I assume so you know when you can shut a worker down?

ShahabT:
The expense comes from the fact that it needs to make up to 2 Visibility queries per build ID. Although those queries will not be needed for the "default" build ID (when no `--select-*` flag is used). We do caching of the Visibility queries in server, so maybe at the end of the day it is OK. We plan to do some more optimizations to reduce the visibility query cost.

I'm fine with keeping this as is and revisit after prerelease if we are OK adding the flags (and change default behavior) at a later stage if needed.

cretz:
> I'm fine with keeping this as is and revisit after prerelease if we are OK adding the flags (and change default behavior) at a later stage if needed.

Hrmm, I'm not sure we are ok w/ removing this from default and making a flag later. I noticed in the pending Go SDK update we require users to opt-in to this data as well. @antlai-temporal - if this data is not available by default in SDK or API and has an expense, can we put it behind a flag here too?

antlai-temporal:
> Hrmm, I'm not sure we are ok w/ removing this from default and making a flag later. I noticed in the pending Go SDK update >we require users to opt-in to this data as well. @antlai-temporal - if this data is not available by default in SDK or API and has >an expense, can we put it behind a flag here too?

There is caching and rate limiting built-in @ShahabT mentioned, and I'm not expecting cli users to be doing heavy stuff. More advanced users will use the SDK, where there is by default more control.

I really hate it when you issue a command in the cli, and nothing useful comes back, and you need to start digging with the options to get anything sensible... I prefer to get the info back by default, and if we need to save, have a disable flag. But instead of falling for premature optimization, I wanted to wait and see whether the flag was actually needed...

cretz:
> nothing useful comes back

But something useful is coming back by default (poller info), just not this. Many users don't need reachability anyways because they are not using versioning.

antlai-temporal:
> But something useful is coming back by default (poller info), just not this. Many users don't need reachability anyways because they are not using versioning.

In the SDK poller info is also gated by a flag, so if we do the same nothing would be returned by the cli by default. We could just add the `enableReachability` flag and get rid of the `enablePollerInfo` flag in the cli...

cretz:
> We could just add the enableReachability flag and get rid of the enablePollerInfo flag in the cli...

Yes, I vote this. Lets show the cheap valuable-to-everyone poller info by default and the more expensive valuable-to-versioning-users-only reachability info a flag is set.

antlai-temporal:
Adding `report-reachability` flag. Note that there is a bug in the current server version that makes a new unit test fail. This will be fixed upgrading the server hash in the next "NOT FOR MERGE" commit in the versioning-2 branch.

Issue #541, opened 2024-04-29
[Bug] TEMPORAL_ENV may not be working


cretz:
### Describe the bug

Confirm `TEMPORAL_ENV` env var properly sets the equivalent of `--env`. This may require special handling since that flag is used before others (because it governs what the other flags have to be).

josh-berry:
Took a quick look at the code and [preprocessOptions()](https://github.com/temporalio/cli/blob/f806427a06f251f38e74ca34608484d1d08ccc45/temporalcli/commands.go#L110) does not appear to check the environment variable, but looks directly at the command line arguments. So I think this is an implementation oversight.

(We should probably have more testing around `--env` in general I think; I noticed another regression around `temporal env delete` I'll have to either fix or write up properly tomorrow.)

PR #568, opened 2024-05-18
Fix setting env name with a property

antlai-temporal:
## What was changed
Ensure that we read the property `TEMPORAL_ENV`, which sets then env name,
early on, i.e., in `preprocessOptions`.
## Why?
Otherwise it is too late to have an effect...

## Checklist

1. Closes 
#541 
2. How was this tested:

Added a system test
3. Any docs updates needed?
No

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -89,6 +89,8 @@ func NewCommandContext(ctx context.Context, options CommandOptions) (*CommandCon
 	return cctx, stop, nil
 }
 
+const temporalEnv = "TEMPORAL_ENV"
+
 func (c *CommandContext) preprocessOptions() error {
 	if len(c.Options.Args) == 0 {
 		c.Options.Args = os.Args[1:]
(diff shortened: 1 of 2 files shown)


temporalcli/commands.workflow_exec_test.go
```diff
@@ -489,6 +489,21 @@ func (s *SharedServerSuite) TestWorkflow_Execute_EnvConfig() {
 	)
 	s.NoError(res.Err)
 	s.ContainsOnSameLine(res.Stdout.String(), "Result", `"env-conf-input"`)
+
+	// And we can specify `env` with a property
+	s.NoError(os.Setenv("TEMPORAL_ENV", "myenv"))
+	defer os.Unsetenv("TEMPORAL_ENV")
```
cretz, line 495: You should be able to replace `LookupEnv` in the CLI options instead of using actual environment variables that affect everything globally. See `TestWorkflow_Execute_EnvVars` in `commands.workflow_exec_test.go` for an example (I saw we unfortunately have others not doing this in tests that should).

antlai-temporal: I saw that but I though using a system property "for real" may be a better test than faking it... Are there any side-effects of using a real system property in the test?

cretz: You can damage potential parallel tests in the future. It's mocked out at the highest level, so you're not getting much more using the real env var I don't think.

temporalcli/commands.go
```diff
@@ -89,6 +89,8 @@ func NewCommandContext(ctx context.Context, options CommandOptions) (*CommandCon
 	return cctx, stop, nil
 }
 
+const temporalEnv = "TEMPORAL_ENV"
```
cretz, line 92: I don't think there's much value in making this a constant that is used in one place, but meh

cretz (approved):

PR #570, opened 2024-05-20
Update the docker-builds repo when releases are published

tdeebswihart:
## What was changed
I added a GitHub workflow that will keep the CLI submodule in [temporalio/docker-builds](https://github.com/temporalio/docker-builds) up to date with new CLI releases.

As part of this change we're updating our internal release processes to include releasing a new version of `temporalio/admin-tools` whenever a release to the CLI is made.

## Why?
So that we always use the most recent CLI when building containers.

### DIFF
--- a/.github/workflows/trigger-publish.yml
+++ b/.github/workflows/trigger-publish.yml
@@ -1,0 +1,38 @@
+name: 'Trigger Docker image build'
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  trigger:
+    name: 'trigger Docker image build'
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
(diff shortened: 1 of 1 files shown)


alexshtin (approved):

.github/workflows/trigger-publish.yml
cretz, line 1: We have used something like https://github.com/temporalio/sdk-python/blob/11a97d1ab2ebfe8c973bf396b1e14077ec611e52/.github/workflows/ci.yml#L102-L103 to run workflows from other repos before. Would that make sense here?

tdeebswihart: Done

cretz: > Done

I saw this PR was merged with no non-merge commits after this comment

PR #576, opened 2024-05-22
Revert "Use "uses" instead of curl (#575)"

tdeebswihart:
## What was changed
This reverts commit 9581b4ed379bd5b5f83760250b78a9e8a69e4d93.

## Why?

See discussion at https://github.com/temporalio/api/pull/407#discussion_r1610428499

### DIFF
--- a/.github/workflows/trigger-publish.yml
+++ b/.github/workflows/trigger-publish.yml
@@ -19,9 +19,20 @@ jobs:
         run: |
           echo branch="${GITHUB_REF#refs/heads/}" >> "$GITHUB_OUTPUT"
 
-      - name: Dispatch docker builds Github Action
-        uses: temporalio/docker-builds/update-submodules.yml@main
+      - name: Generate a token
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92
         with:
-          repo: "cli"
-          branch: ${{ steps.get_branch.outputs.branch }}
-          commit: ${{ env.GITHUB_SHA }}
+          app_id: ${{ secrets.TEMPORAL_CICD_APP_ID }}
+          private_key: ${{ secrets.TEMPORAL_CICD_PRIVATE_KEY }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-publish.yml
```diff
@@ -19,9 +19,20 @@ jobs:
...
+      - name: Dispatch docker builds Github Action
+        env:
+          PAT: ${{ steps.generate_token.outputs.token }}
+          PARENT_REPO: temporalio/docker-builds
+          PARENT_BRANCH: ${{ toJSON('main') }}
+          WORKFLOW_ID: update-submodules.yml
+          REPO: ${{ toJSON('cli') }}
+          BRANCH: ${{ toJSON(steps.get_branch.outputs.branch) }}
+        run: |
+          curl -fL -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/workflows/$WORKFLOW_ID/dispatches" -d '{"ref":'"$PARENT_BRANCH"', "inputs": { "repo":'"$REPO"', "branch":'"$BRANCH"', "commit": '"$GITHUB_SHA"' }}'
```
cretz, line 38: I wonder if this should take the `gh workflow run` approach of https://github.com/temporalio/api/pull/407. Up to you, so long as it works.

cretz (approved):

.github/workflows/trigger-publish.yml
tdeebswihart: Have we checked to see if that approach works yet? I know this one does as it's what keeps docker-builds in sync with temporal

cretz: Unsure, may have to ask @rodrigozhou. I don't mind either.

tdeebswihart: I'm going to merge this for now so that we have something working. I'd be happy to make the change in the future once we're sure it works

PR #577, opened 2024-05-23
Nexus

bergundy:
## What was changed

- Added Nexus Endpoint CRUD commands
- Added Pending Nexus Operations and Callbacks to workflow describe command.


temporalcli/commandsmd/commands.md
```diff
@@ -378,6 +379,113 @@ However, URI values for archival states cannot be changed after the states are e
...
+temporal operator nexus endpoint update \
+  --name my-endpoint \
+  --description-file DESCRIPTION.md
+```
+
+#### Options
+
+* `--name` (string) - Endpoint name. Required.
+* `--description` (string) - Endpoint description in markdown format (encoded using the configured codec server).
+* `--description-file` (string) - Endpoint description file in markdown format (encoded using the configured codec server).
```
cretz, line 483: This was removed in cloud API at https://github.com/temporalio/api-cloud/pull/23#discussion_r1595946337, I think we should remove in this one too. Or we can have description, but I don't think we should get specific with the format expected at this time (I suspect there is a subset of markdown we will start using as part of https://github.com/temporalio/features/issues/486).

antlai-temporal:
The fix for tqname is already in the versioning-2 branch but it may take a while (blocked on go SDK merge) until we merge it. It is just two lines, and I don't want to block your PR,  do you want to go ahead and add them?

cretz (commented):
Mostly LGTM (waiting for server support before approving). Also, would like to see the `tcld` equivalent to ensure similarity.

temporalcli/commandsmd/commands.md
bergundy: We may need to discuss this more. There's a strict requirement to have this description. We have it in all of the docs.
I've replied in that cloud API thread to request bringing the description back.

cretz: Is there a strict requirement to have the description, or a strict requirement to say descriptions are markdown formatted. My issue is more with the latter and I think it should at least be subject to discussion.

bergundy: Yeah, we want markdown. Maybe it's worth consolidating on summary and details to match workflow metadata, I would be okay changing that now.

cretz: I am not sure we can generally say we want all long-form "description" text accepted by Temporal APIs to be full markdown, and whatever we do want Temporal long-form format to be should be centrally/generally documented. Setting a long-form format specific to Nexus descriptions and not for Temporal as a whole I don't think is the best approach.

cretz (approved):

Issue #547, opened 2024-05-02
[Feature Request] Add `ResultEncoding` when printing workflow results that cannot be converted to shorthand JSON


cretz:
### Describe the solution you'd like

Today users get:

```
Result:
  Status  COMPLETED
  Result  {"metadata":{"encoding":"YmluYXJ5L3NuYXBweQ=="},"data":"NdAKFgoIZW5jb2RpbmcSCmpzb24vcGxhaW4SGyJSZWNlaXZlZCBQbGFpbiB0ZXh0IGlucHV0Ig=="}
```

We need to tell them that the result is encoded and with what encoding if we can. So maybe something like:

```
Result:
  Status          COMPLETED
  Result          {"metadata":{"encoding":"YmluYXJ5L3NuYXBweQ=="},"data":"NdAKFgoIZW5jb2RpbmcSCmpzb24vcGxhaW4SGyJSZWNlaXZlZCBQbGFpbiB0ZXh0IGlucHV0Ig=="}
  ResultEncoding  Encoded with 'binary/snappy` encoding
```

PR #579, opened 2024-05-28
Print encoding of workflow results

Sushisource:
DRAFT because it seems like tests are getting stuck but, nothing I changed should do that. Maybe just my machine.

## What was changed
Print encoding of workflow results 

## Why?
To help the user understand what's going on in the event Result can't be decoded

## Checklist

1. Closes https://github.com/temporalio/cli/issues/547

2. How was this tested:
Added UT

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.workflow_view_test.go
+++ b/temporalcli/commands.workflow_view_test.go
@@ -138,9 +138,7 @@ func (s *SharedServerSuite) TestWorkflow_Describe_NotDecodable() {
 	s.ContainsOnSameLine(out, "Status", "COMPLETED")
 	s.ContainsOnSameLine(out, "ResultEncoding", "some-encoding")
 
-	// TODO: The below fails seemingly due to a marshalling bug in api-go. The conversion of the
-	//   returned raw payload in to JSON bytes is invalid json, and breaks the printing on this
-	//   side.
+	// TODO: Enable once updated to api-go >= 1.33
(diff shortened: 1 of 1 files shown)


temporalcli/commands.workflow_view_test.go
```diff
@@ -107,6 +108,53 @@ func (s *SharedServerSuite) TestWorkflow_Describe_Completed() {
...
+		"--address", s.Address(),
+		"-w", run.GetID(),
+	)
+	s.NoError(res.Err)
+	out := res.Stdout.String()
+	s.ContainsOnSameLine(out, "Status", "COMPLETED")
+	s.ContainsOnSameLine(out, "ResultEncoding", "some-encoding")
+
+	// TODO: The below fails seemingly due to a marshalling bug in api-go. The conversion of the
+	//   returned raw payload in to JSON bytes is invalid json, and breaks the printing on this
```
Sushisource, line 142: I'm going to take a look at fixing this in api-go

Sushisource: NM, apparently this is already fixed, probably by https://github.com/temporalio/api-go/pull/159

Sushisource: We just need to update versions when possible

cretz (approved):

PR #583, opened 2024-06-04
Normalize semantic of --data args

### LINKED ISSUE #552: `temporal operator namespace commands` do not treat `--data` consistently
`temporal operator namespace create -n foo --data key1=value1,key2=value2`

...is not consistent with...

`temporal operator namespace update -n foo --data key1=value1 --data key2=value2` 

We should make `namespace create` accept multiple key/value pairs thru multiple uses of `--data`, to be consistent with `namespace update`.

mjameswh:
## What was changed

* The `--data` option on `temporal operator namespace create` now behave the same way as on `temporal operator namespace update`. That is, multiple key-value pairs can be provided by repeating the `--data` flag, like this:

  ```
  temporal operator namespace create -n foo --data key1=value1 --data key2=value2`
  ```

* 💥 **BREAKING** 💥: The former syntax allowed by the `--data` option, i.e. `--data key1=value1,key2=value2`, is no longer supported. That specific command is not frequent enough to justify soft deprecation.

## Checklist

1. Closes #552

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -622,7 +622,7 @@ type TemporalOperatorNamespaceCreateCommand struct {
 	Command                 cobra.Command
 	ActiveCluster           string
 	Cluster                 []string
-	Data                    string
+	Data                    []string
 	Description             string
 	Email                   string
 	Global                  bool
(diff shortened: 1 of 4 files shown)


cretz (approved):
LGTM. Can you make it clear in the description this is a breaking change so we can note it in the release notes?

PR #584, opened 2024-06-04
Prepare dependencies and tests for CLI 0.13-rc.1  release (server v1.24.1)

josh-berry:


temporalcli/commands.taskqueue.go
```diff
@@ -10,7 +10,7 @@ import (
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/taskqueue/v1"
 	"go.temporal.io/api/workflowservice/v1"
-	"go.temporal.io/server/common/tqname"
+	"go.temporal.io/server/common/tqid"
```
cretz, line 13: IMO we should open a ticket to stop referencing server code outside of devserver (if possible).

cretz (approved):

PR #587, opened 2024-06-06
Upgrade tchannel-go.

robholland:
## What was changed
Upgraded tchannel-go.

## Why?
This brings in a fix which allows it to compile on FreeBSD.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -93,7 +93,7 @@ require (
 	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/temporalio/ringpop-go v0.0.0-20230606200434-b5c079f412d3 // indirect
 	github.com/temporalio/sqlparser v0.0.0-20231115171017-f4060bcfa6cb // indirect
-	github.com/temporalio/tchannel-go v1.22.1-0.20231116015023-bd4fb7678499 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -93,7 +93,7 @@ require (
 	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/temporalio/ringpop-go v0.0.0-20230606200434-b5c079f412d3 // indirect
 	github.com/temporalio/sqlparser v0.0.0-20231115171017-f4060bcfa6cb // indirect
-	github.com/temporalio/tchannel-go v1.22.1-0.20231116015023-bd4fb7678499 // indirect
+	github.com/temporalio/tchannel-go v1.22.1-0.20240528171429-1db37fdea938 // indirect
```
cretz, line 96: This is an indirect dependency that needs to be updated when we update server. We need to keep the CLI indirect dependencies the same with server that have been tested with server's CI. It looks like it will have to wait until 1.25 to get your change from https://github.com/temporalio/temporal/pull/6003. This may happen sooner than we think as we are considering releasing RC versions of CLI w/ RC versions of server.

Also, if we want the server to officially support FreeBSD there should be a CI test for it in server.

PR #594, opened 2024-06-12
Publish only non-RC releases

antlai-temporal:
## What was changed
 Change GitHub action for publishing releases so it only updates docker images with releases that are NOT release candidates.

## Why?

Avoid publishing OSS server docker images with RC cli dependencies.


cretz (approved):
LGTM. We may have to revisit if we ever do a release of of another branch besides main.

Issue #479, opened 2024-02-27
[Feature Request] Add support for the new versioning API


antlai-temporal:
### Is your feature request related to a problem? Please describe.

The versioning API is changing with new abstractions that need to be supported by the CLI.

This work should be done using the new version of the CLI.

PR #559, opened 2024-05-10
Add AssignedBuildId to workflow describe

ShahabT:
## What was changed

`workflow describe` now shows WF's assigned build ID as follows:
```
$temporal --env shahab-test4 workflow describe --workflow-id b72b1a82-7a83-4cc9-9bde-6ad8d411a47c      
Execution Info:
  WorkflowId            b72b1a82-7a83-4cc9-9bde-6ad8d411a47c
  RunId                 c81a42c2-efc1-4c41-9088-a0123c0d1034
  Type                  MyWorkflow
  Namespace             shahab-test4.temporal-dev
  TaskQueue             my-tq
  AssignedBuildId       BUILD_ID_HERE
  StartTime             4 hours ago
  ExecutionTime         4 hours ago
  SearchAttributes      map[BuildIds:metadata:{key:"encoding" value:"json/plain"} metadata:{key:"type" value:"KeywordList"} data:"[\"assigned:C\",\"versioned:C\"]"]
  StateTransitionCount  2
  HistoryLength         2
  HistorySize           263
```

## Why?

This is an important attribute for WFs running on versioned workers.

## Checklist

1. Closes 

2. How was this tested:

Unit test added.

3. Any docs updates needed?

No.

PR #556, opened 2024-05-10
Implement Get/Update methods for versioning rules

antlai-temporal:
## What was changed

Implements `task-queue versioning` methods including:

  `add-redirect-rule`       Adds the rule to the list of redirect rules for this Task Queue.
  `commit-build-id`         Completes the rollout of a Build ID for this Task Queue.
  `delete-assignment-rule`  Deletes the rule at a given index in the list of assignment rules for this Task Queue.
  `delete-redirect-rule`    Deletes the rule with the given build ID for this Task Queue.
  `insert-assignment-rule`  Inserts the rule to the list of assignment rules for this Task Queue.
  `replace-assignment-rule` Replaces the rule at a given index in the list of assignment rules for this Task Queue.
  `replace-redirect-rule`   Replaces the redirect rule with the given source build ID for this Task Queue.
   `get-rules`        Retrieves the worker Build ID assignment and redirect rules on the Task Queue.

PR #596, opened 2024-06-14
Support worker versioning with new API

antlai-temporal:
## What was changed

This PR combines changes from several already (partially) reviewed to provide support for new worker versioning APIs in pre-release.

In particular:
#556 
#559 
#565 

It also updates the cli to the latest go SDK (v1.27.0),  which is required for the new APIs

**Breaking Changes**

`task-queue describe` is now also returning task reachability information in addition to pollers info. It uses an enhanced API that is not supported by a server with version prior to 1.24. To access pollers info in older servers set the `--legacy-mode` flag, and select the  task queue type, if different from `workflow`, using  the `--task-queue-type-legacy` argument.  

## Why?

Enable Pre-release testing of new worker versioning APIs.

## Checklist

1. Closes 
#479


temporalcli/commands.workflow.go
```diff
@@ -200,16 +200,17 @@ func (c *TemporalWorkflowUpdateCommand) run(cctx *CommandContext, args []string)
...
 
-	request := &client.UpdateWorkflowWithOptionsRequest{
+	request := client.UpdateWorkflowOptions{
 		WorkflowID:          c.WorkflowId,
 		RunID:               c.RunId,
 		UpdateName:          c.Name,
 		UpdateID:            c.UpdateId,
 		FirstExecutionRunID: c.FirstExecutionRunId,
 		Args:                input,
+		WaitForStage:        client.WorkflowUpdateStageCompleted,
```
Quinn-With-Two-Ns, line 210: Not sure we should specify a default here for the same reason we have a default in the SDK

antlai-temporal: I was assuming that a different update PR would decide what to expose. I could open an issue so that we don't forget...

temporalcli/commands.taskqueue.go
```diff
@@ -10,31 +10,243 @@ import (
...
 	"go.temporal.io/api/taskqueue/v1"
 	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/server/common/tqid"
 )
 
+const taskQueueUnversioned = "UNVERSIONED"
+
+type reachabilityRowType struct {
+	BuildID      string `json:"buildID"`
```
cretz, line 20:
```suggestion
	BuildID      string `json:"buildId"`
```

Pedantic, but I think the `Id` form is better JSON

temporalcli/commands.taskqueue.go
```diff
@@ -10,31 +10,243 @@ import (
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/taskqueue/v1"
 	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/server/common/tqid"
 )
 
+const taskQueueUnversioned = "UNVERSIONED"
+
+type reachabilityRowType struct {
```
cretz, line 19:
```suggestion
type taskQueueReachabilityRowType struct {
```

Since this package is shared by all, might as well qualify the task queue stuff, same with all the little functions, but not a big deal

temporalcli/commands.taskqueue.go
```diff
@@ -10,31 +10,243 @@ import (
...
+			pRows, err = buildIDToPollerRows(pRows, k, val.TypesInfo)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	return pRows, nil
+}
+
+func descriptionToRows(taskQueueDescription client.TaskQueueDescription, reportReachability bool) (taskQueueDescriptionType, error) {
```
cretz, line 137: This name is definitely too general for a large shared package. Also, if a function is called once, consider just inlining it instead of all of these tiny functions (not very common in Go to have a bunch of tiny single-use functions)

temporalcli/commandsmd/commands.md
```diff
@@ -624,17 +639,25 @@ request.
...
 
 * `--task-queue`, `-t` (string) - Task queue name. Required.
-* `--task-queue-type` (string-enum) - Task Queue type. Options: workflow, activity. Default: workflow.
-* `--partitions` (int) - Query for all partitions up to this number (experimental+temporary feature). Default: 1.
+* `--task-queue-type` (string[]) - Task queue types considered. If not specified, all types are reported. The current valid queue types are workflow, activity, or nexus.
+* `--select-build-id` (string[]) - Task queue filter based on Build ID.
+* `--select-unversioned` (bool) - Include the unversioned queue.
+* `--select-all-active` (bool) - Include all active versions. A version is active if it had new tasks or polls recently.
+* `--report-reachability` (bool) - Display task reachability information.
+* `--legacy-mode` (bool) - Enable a legacy mode for servers that do not support rules-based worker versioning. This mode only provides pollers info.
```
cretz, line 656: Now I am wondering if we should just move the whole `task-queue describe` to `task-queue describe-legacy` or something

temporalcli/commands.taskqueue.go
antlai-temporal: Done

temporalcli/commands.taskqueue.go
antlai-temporal: Yes, keep forgetting is single package...

temporalcli/commands.taskqueue.go
antlai-temporal: renamed to `taskQueueDescriptionToRows`

temporalcli/commandsmd/commands.md
antlai-temporal: I don't want to emphasize too much the legacy mode, the new mode is a superset, and the old one should go away... Also, in the usage stats there are very few current users of `task describe`,  so adding a top level command feels like an overkill... But yes, I agree with you that we should make it very clear the breaking changes in the release notes...

cretz: :+1: Agreed on overkill, no prob

go.mod
```diff
@@ -153,3 +153,7 @@ require (
 	modernc.org/strutil v1.2.0 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
+
+replace go.temporal.io/api v1.34.0 => go.temporal.io/api v1.32.1
```
dandavison, line 157: What's the reason for downgrading `api`? This is the cause of https://github.com/temporalio/cli/issues/607, because it reintroduces a bug fixed in `api v1.34.0`: https://github.com/temporalio/api-go/pull/159

(I don't think `replace` statements should be committed; it makes it easy to look at the top of the file and reach an incorrect conclusion about what version it's on.)

dandavison: I'm fixing this by having `cli` use a cherry-pick branch of `api` for now: https://github.com/temporalio/cli/pull/617

antlai-temporal: I had to downgrade the api with redirect because the go sdk release moved to a version of the api that was incompatible with the server one (the embedded server would not compile with that api), but luckily the sdk did not really need the new api , so I could get it to compile with the old one with the redirect...

Go sdk tends to upgrade to latest api in most releases, but server is slower, and the embedded server shares api with the sdk...

PR #606, opened 2024-06-27
Clarify Versioning CLI Commands

carlydf:
## What was changed
-  Remove redirect rule reference in `commit-build-id`, fix spelling
- Improve `--select-build-id` flag description

## Why?
- `commit-build-id` does not touch redirect rules, only assignment rules
- add optional behavior of `--select-build-id`

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -1344,7 +1344,7 @@ func NewTemporalTaskQueueDescribeCommand(cctx *CommandContext, parent *TemporalT
 	s.Command.Flags().StringVarP(&s.TaskQueue, "task-queue", "t", "", "Task queue name. Required.")
 	_ = cobra.MarkFlagRequired(s.Command.Flags(), "task-queue")
 	s.Command.Flags().StringArrayVar(&s.TaskQueueType, "task-queue-type", nil, "Task queue types considered. If not specified, all types are reported. The current valid queue types are workflow, activity, or nexus.")
-	s.Command.Flags().StringArrayVar(&s.SelectBuildId, "select-build-id", nil, "Task queue filter based on Build ID.")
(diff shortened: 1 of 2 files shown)


temporalcli/commands.gen.go
```diff
@@ -1344,7 +1344,7 @@ func NewTemporalTaskQueueDescribeCommand(cctx *CommandContext, parent *TemporalT
 	s.Command.Flags().StringVarP(&s.TaskQueue, "task-queue", "t", "", "Task queue name. Required.")
 	_ = cobra.MarkFlagRequired(s.Command.Flags(), "task-queue")
 	s.Command.Flags().StringArrayVar(&s.TaskQueueType, "task-queue-type", nil, "Task queue types considered. If not specified, all types are reported. The current valid queue types are workflow, activity, or nexus.")
-	s.Command.Flags().StringArrayVar(&s.SelectBuildId, "select-build-id", nil, "Task queue filter based on Build ID.")
+	s.Command.Flags().StringArrayVar(&s.SelectBuildId, "select-build-id", nil, "Task queue filter based on Build ID. If not specified, the default Build ID will be used.")
```
cretz, line 1347: This file is auto-generated, you have to update the `commands.md` source and regen. See `CONTRIBUTING.md`.

drewhoskins-temporal, line 1347: What's the default Build ID?  Is that something the user specifies in an assignment rule or another command?  If so, maybe add a breadcrumb.  (sometimes it's good to use code names rather describe things indirectly in english)

antlai-temporal: The default Build ID is described in the command description. There is also a catch-all comment about default values on --select-* so there is no need to add one for each. See `commandsmd/commands.md`

josh-berry:
@carlydf Is this still relevant or can it be closed?

PR #614, opened 2024-07-03
Update UI module for versioning pre-release

antlai-temporal:
## What was changed

Update UI dependency.

## Why?

Needed for versioning pre-release.


cretz (approved):
We don't have a great end-to-end testing story for dev server UI. Just want to make sure we have at least confirmed it works manually clicking around the UI.

antlai-temporal:
> We don't have a great end-to-end testing story for dev server UI. Just want to make sure we have at least confirmed it works 
> manually clicking around the UI.
I did start the server, created a workflow, got the UI showing history, Build ID for workers, and so on...

PR #637, opened 2024-08-19
Include additional build-time information in the version string

### LINKED ISSUE #635: [Feature Request] Add information about server build to `start-dev` banner
I'd like to see info about the commit hash of Server that it has been built against, if go tooling makes that possible. (From brief investigation it seemed that `go list -json -m go.temporal.io/server` did not contain that information when a local `replace` is in effect.)

dandavison:
Fixes #635

Support injection of build-time information into the version string printed by the binary on startup and via `--version`. See code comments and docs for further explanation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,10 +31,10 @@ This will expect every non-parent command to have a `run` method, so for new com
 ## Inject additional build-time information
 To add build-time information to the version string printed by the binary, use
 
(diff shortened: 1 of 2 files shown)


cretz (approved):
LGTM. My only concern is that this now effectively becomes part of the public API and we can't change or remove this var. But that's probably ok.

dandavison:
Thanks @cretz. Yeah, so we basically have to be happy with it being `github.com/temporalio/cli/temporalcli.buildInfo`.

PR #638, opened 2024-08-19
Making tq stats visible from cli

Shivs11:
## What was changed
- Adds support to view TQ stats from the CLI.
- Sample input and output for the command:

`temporal task-queue describe --task-queue xxx`

### Task Queue Statistics

| BuildID     | TaskQueueType | ApproximateBacklogCount | ApproximateBacklogAge | BacklogIncreaseRate | TasksAddRate | TasksDispatchRate |
|-------------|---------------|-------------------------|-----------------------|---------------------|--------------|-------------------|
| UNVERSIONED | workflow       | 0                       | 0s                    | 0                   | 0            | 0                 |
| UNVERSIONED | activity       | 0                       | 0s                    | 0                   | 0            | 0                 |

`temporal task-queue describe --task-queue xxx  -o json`

`
"stats": [
  {
    "buildId": "UNVERSIONED",
    "taskQueueType": "workflow",
    "approximateBacklogCount": 0,
    "approximateBacklogAge": 0,
    "backlogIncreaseRate": 0,
    "tasksAddRate": 0,
    "tasksDispatchRate": 0
  },
  {
    "buildId": "UNVERSIONED",
    "taskQueueType": "activity",
    "approximateBacklogCount": 0,
    "approximateBacklogAge": 0,
    "backlogIncreaseRate": 0,
    "tasksAddRate": 0,
    "tasksDispatchRate": 0
  }
]
`

## Why?
- Shall help in worker fleet management.

## Checklist

1. Closes 

2. How was this tested:
- Modified an existing unit test
- Ensured all other unit tests pass

3. Any docs updates needed?

- Updated docs.


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/stretchr/testify v1.9.0
 	github.com/temporalio/ui-server/v2 v2.29.2
 	go.temporal.io/api v1.36.0
-	go.temporal.io/sdk v1.27.0
+	go.temporal.io/sdk v1.28.2-0.20240816033637-2a02c483658a
```
Shivs11, line 18: Upgraded sdk to point to the version where TQ stats were introduced in DescribeTaskQueueEnhanced (#1553)

temporalcli/commands.taskqueue.go
```diff
@@ -134,15 +145,64 @@ func descriptionToPollerRows(taskQueueDescription client.TaskQueueDescription) (
...
+			ApproximateBacklogCount: info.Stats.ApproximateBacklogCount,
+			ApproximateBacklogAge:   info.Stats.ApproximateBacklogAge,
+			BacklogIncreaseRate:     info.Stats.BacklogIncreaseRate,
+			TasksAddRate:            info.Stats.TasksAddRate,
+			TasksDispatchRate:       info.Stats.TasksDispatchRate,
+		})
+	}
+	return statsRows, nil
+}
+func descriptionToStatsRows(taskQueueDescription client.TaskQueueDescription) ([]statsRowType, error) {
```
ShahabT, line 166: Nit: should have empty line between functions

temporalcli/commands.taskqueue_test.go
```diff
@@ -68,11 +79,26 @@ func (s *SharedServerSuite) TestTaskQueue_Describe_Simple() {
...
+	// With task queue statistics info
+	res = s.Execute(
+		"task-queue", "describe",
+		"--address", s.Address(),
+		"--report-stats",
+		"--task-queue", s.Worker().Options.TaskQueue)
+	s.NoError(res.Err)
+
+	s.NotContains(res.Stdout.String(), "reachable")
+	s.ContainsOnSameLine(res.Stdout.String(), "UNVERSIONED", "workflow", "0", "0s", "0", "0", "0")
```
ShahabT, line 91: can we have a test with non-zero results?

temporalcli/commands.gen.go
```diff
@@ -1556,6 +1557,7 @@ func NewTemporalTaskQueueDescribeCommand(cctx *CommandContext, parent *TemporalT
 	s.TaskQueueTypeLegacy = NewStringEnum([]string{"workflow", "activity"}, "workflow")
 	s.Command.Flags().Var(&s.TaskQueueTypeLegacy, "task-queue-type-legacy", "Task Queue type (legacy mode only). Accepted values: workflow, activity.")
 	s.Command.Flags().IntVar(&s.PartitionsLegacy, "partitions-legacy", 1, "Query for all partitions up to this number (experimental+temporary feature) (legacy mode only).")
+	s.Command.Flags().BoolVar(&s.ReportStats, "report-stats", false, "Display task queue statistics.")
```
ShahabT, line 1560: what's the default value? I'm thinking it should be true by default.

Shivs11: hmm, why do you think it should be true? The default value is false and I think it should be kept that way - noticed other flags were also following a similar principle unless we have something specific for this one

cretz (commented):
Can you show what the output of this command w/ stats for a normal task queue (no versioning) looks like in text and JSON form? Can just be as a comment here in the PR. It's hard to get a feel for what the user sees from the code alone.

temporalcli/commands.gen.go
```diff
@@ -1556,6 +1557,7 @@ func NewTemporalTaskQueueDescribeCommand(cctx *CommandContext, parent *TemporalT
 	s.TaskQueueTypeLegacy = NewStringEnum([]string{"workflow", "activity"}, "workflow")
 	s.Command.Flags().Var(&s.TaskQueueTypeLegacy, "task-queue-type-legacy", "Task Queue type (legacy mode only). Accepted values: workflow, activity.")
 	s.Command.Flags().IntVar(&s.PartitionsLegacy, "partitions-legacy", 1, "Query for all partitions up to this number (experimental+temporary feature) (legacy mode only).")
+	s.Command.Flags().BoolVar(&s.DisableStats, "disable-stats", false, "Disable task queue statistics.")
```
Shivs11, line 1560: @ShahabT  - Since this has now been changed to `disable-stats` with a default value of false, it brings us back to the bug we found in matching where a versioned queue may not be loaded leading to null dereferences. Again, not a big deal as you said since versioning happens to be pre-release but just wanted to make it a note here.

temporalcli/commands.taskqueue_test.go
```diff
@@ -10,6 +11,107 @@ import (
...
+	s.NoError(res.Err)
+
+	result := s.Execute(
+		"task-queue", "describe",
+		"--address", s.Address(),
+		"--task-queue", s.Worker().Options.TaskQueue,
+	)
+	s.NoError(result.Err)
+	out := result.Stdout.String()
+	s.NotContainsOnSameLine(out, "UNVERSIONED", "workflow", "0", "0s", "0", "0", "0")
```
ShahabT, line 84: `NotContains` would return true even if nothing is printed. I think we should check the things we want to be shown are shown. (if backlog age is an issue, we should skip that somehow or put a wildcard there).

Also, we should put these checks in EventuallyWithT because the stats may not be updated right away and that'd make the test flaky.

Lastly, I'd check for stats of both wf and activity types, even if activity is expected to have zero stats. This would catch the case if by accident (now or in the future) we print workflow stats for activity as well.

temporalcli/commands.taskqueue.go
```diff
@@ -29,9 +29,20 @@ type pollerRowType struct {
...
+	ApproximateBacklogAge   time.Duration `json:"approximateBacklogAge"`
+	BacklogIncreaseRate     float32       `json:"backlogIncreaseRate"`
+	TasksAddRate            float32       `json:"tasksAddRate"`
+	TasksDispatchRate       float32       `json:"tasksDispatchRate"`
+}
+
 type taskQueueDescriptionType struct {
 	Reachability []taskQueueReachabilityRowType `json:"reachability"`
 	Pollers      []pollerRowType                `json:"pollers"`
+	Stats        []statsRowType                 `json:"stats"`
```
ShahabT, line 45: What's the benefit of separate maps for build ID -> reachability, build ID -> poller, and build ID -> stats? isn't better to have a single map of build ID -> {reachability, pollers, stats}? cc @antlai-temporal @cretz

Shivs11: I too, am not totally opposed to this idea - included the new field cause I thought a potential norm could have been to have separate maps for each but would love to hear your thoughts @cretz

temporalcli/commands.gen.go
Shivs11: Outdated comment since a patch was merged to main last week - [PR](https://github.com/temporalio/temporal/pull/6437)

temporalcli/commands.taskqueue.go
cretz: I have no strong opinions on the structure of the output (nor did I when this was first developed). Anything sane works for me. But just remember, it must be stable.

ShahabT:
@antlai-temporal @Shivs11 I think it's better to change the short and long descriptions of the `describe` command. Right now it seems we're over-focusing on the reachability info. e.g. the short description is "Provides task reachability and pollers information for Workers on this Task Queue.".

I'd suggest we mention poller info and stats as the main output. Reachability can be mentioned as a experimental part of the API with proper warning that it may change in the future. Note that there is a chance that we want to move reachability API under `temporal versioning` command, and in that case it's not clear if we'd still keep it under `temporal task-queue describe` too.

Shivs11:
> @antlai-temporal @Shivs11 I think it's better to change the short and long descriptions of the `describe` command. Right now it seems we're over-focusing on the reachability info. e.g. the short description is "Provides task reachability and pollers information for Workers on this Task Queue.".
> 
> I'd suggest we mention poller info and stats as the main output. Reachability can be mentioned as a experimental part of the API with proper warning that it may change in the future. Note that there is a chance that we want to move reachability API under `temporal versioning` command, and in that case it's not clear if we'd still keep it under `temporal task-queue describe` too.

I was not aware that the reachability side of things could be moved over later and knowing this now, it makes sense to not over-focus on it. I am for the act of keeping pollers and stats as the main output but shall wait for @antlai-temporal to give us the green light 🚥

temporalcli/commands.taskqueue.go
```diff
@@ -29,9 +29,20 @@ type pollerRowType struct {
 	RatePerSecond  float64   `json:"ratePerSecond"`
 }
 
+type statsRowType struct {
+	BuildID                 string        `json:"buildId"`
+	TaskQueueType           string        `json:"taskQueueType"`
+	ApproximateBacklogCount int64         `json:"approximateBacklogCount"`
+	ApproximateBacklogAge   time.Duration `json:"approximateBacklogAge"`
```
ShahabT, line 36: What's the type of age in json? is it a number? if so, we need to qualify the name with something as "approximateBacklogAgeSeconds".

Shivs11: We are returning a Duration (effectively, nanoseconds) which gets printed for JSON. Yes, it is a number. Sample output - 

``` "approximateBacklogAge": 4108891000, ```

Comment (Shivs11) [reply temporalcli/commands.taskqueue.go:36]: As discussed offline, the type for JSON was changed from `time.Duration` to `string`. Thus, I think it's okay to keep the same name.

**Comment by cretz** [comment]
> Can you show what the output of this command w/ stats for a normal task queue (no versioning) looks like in text and JSON form?

I saw the markdown table in PR description but I think we want to see what the raw table looks like in text (and will wait until after y'all's discussion on text/json format to review what latest output looks like and approve).

**Comment by Shivs11** [comment]
> > Can you show what the output of this command w/ stats for a normal task queue (no versioning) looks like in text and JSON form?
> 
> I saw the markdown table in PR description but I think we want to see what the raw table looks like in text (and will wait until after y'all's discussion on text/json format to review what latest output looks like and approve).

Ah, my apologies for the misunderstanding. Pasting the output as seen from the CLI here:

Scenario - Adding a task in the backlog with no pollers:

1. JSON 
image

2. Non-JSON
image

note - the backlog increase rate becomes 0 in the second picture because we are fetching the rate over a window of 30 seconds which were elapsed between the two screenshots.

**Review by cretz — APPROVED** [review]
LGTM, but would like to wait for other approvers because I think they may have stronger opinions on the output structure. And please don't merge until we have a tagged SDK dependency.

File: temporalcli/commands.taskqueue.go (hunk 7 above)
Comment (antlai-temporal) [reply temporalcli/commands.taskqueue.go:45]: It is just for modularity, these are very different top-level concepts let's keep them separate. The overhead of three maps is just minimal...

**Comment by antlai-temporal** [comment]
> I was not aware that the reachability side of things could be moved over later and knowing this now, it makes sense to not 
> over-focus on it. I am for the act of keeping pollers and stats as the main output but shall wait for @antlai-temporal to give us the green light

The main reason to over focus on reachability was the feedback that it was hard to find which command would give us that info. I think that once we move reachaibility to versioning there will be no issue, but until we do, it makes sense to give it priority in the short/long description.

File: temporalcli/commandsmd/commands.md [hunk 9]
```diff
@@ -765,6 +765,7 @@ If there is no default Build ID, the result for the unversioned queue will be re
 * `--legacy-mode` (bool) - Enable a legacy mode for servers that do not support rules-based worker versioning. This mode only provides pollers info.
 * `--task-queue-type-legacy` (string-enum) - Task Queue type (legacy mode only). Options: workflow, activity. Default: workflow.
 * `--partitions-legacy` (int) - Query for all partitions up to this number (experimental+temporary feature) (legacy mode only). Default: 1.
+* `--disable-stats` (bool) - Disable task queue statistics.
```
antlai-temporal, line 769: for reachability we use `report-reachability` as opposed to `disable-reachability` because it is expensive, and we wanted to avoid calls.
Is getting the stats also expensive? In that case we should use also use `report-stats`...

ShahabT: Stats is not the cheapest thing (because it need partition fan-out) but still relatively cheap. I'm comfortable with having it enabled by default. wdyt @dnr ?

We may want to add caching for stats in the root partition in the future (if SDK starts to query it for poller scaling).

temporalcli/commandsmd/commands.md
```diff
@@ -721,10 +721,35 @@ Queue command, run `temporal task-queue [command] [command options]`.
...
 Includes options set for [client](#options-set-for-client).
 
-### temporal task-queue describe: Provides task reachability and pollers information for Workers on this Task Queue.
+### temporal task-queue describe: Provides pollers information, backlog statistics, and task reachability (experimental) for Workers on this Task Queue.
 
-The `temporal task-queue describe` command provides task reachability information for the requested versions and all task types,
-which can be used to safely retire Workers with old code versions, provided that they were assigned a Build ID.
+The `temporal task-queue describe` command provides poller information, backlog statistics and task reachability (experimental) 
+information for the requested versions and all task types, which can be used to safely retire Workers with old code versions, 
+provided that they were assigned a Build ID.
```
ShahabT, line 728: this part of the sentence is only about reachability and it should be removed.

temporalcli/commandsmd/commands.md
```diff
@@ -721,10 +721,35 @@ Queue command, run `temporal task-queue [command] [command options]`.
...
+    - `TasksAddRate`: Approximate tasks per second added to the task queue, averaging over the last 30 seconds.
+Includes tasks whether or not they were added from the backlog or those that were dispatched immediately without
+going to the backlog (sync-matched).
+    - `TasksDispatchRate`: Approximate tasks per second dispatched to the task queue, averaging over the last 30 seconds.
+Includes tasks whether or not they were dispatched from the backlog or those that were dispatched immediately without
+going to the backlog (sync-matched).
+    -  `Backlog Increase Rate`: Approximate net tasks per second added to the backlog, averaging the last 30 seconds.
+This is calculated as `TasksAddRate` - `TasksDispatchRate`.
 
 The reachability states of a Build ID are:
```
ShahabT, line 754:
```suggestion
Task reachability information is returned for the requested versions and all task types, which can be used to safely retire Workers with old code versions, provided that they were assigned a Build ID.

Note that task reachability status is experimental and may significantly change or be removed in a future release.

The reachability states of a Build ID are:
```

temporalcli/commandsmd/commands.md
Shivs11: I agree, however we should have the part : `the requested versions and all task types`. 

All in all, the sentence should look like:
The `temporal task-queue describe` command provides poller information, backlog statistics and task reachability (experimental) information for the requested versions and task types.

ShahabT: > task reachability (experimental) ~information~

"information" seems redundant?

ShahabT (approved):

Issue #597, opened 2024-06-14
[Feature Request] Support `update start` in addition to `update execute`


drewhoskins-temporal:
### Is your feature request related to a problem? Please describe.

Child of https://github.com/temporalio/features/issues/469

josh-berry:
Notes from team discussion:

- `temporal workflow update` really needs sub-commands
- `temporal workflow update start` should be added, and require a wait stage
- `temporal workflow update execute` should follow the existing behavior of `temporal workflow update`
- We can add `temporal workflow update terminate` later for workflow cancellation

PR #642, opened 2024-08-21
Add `workflow update start` and `workflow update execute` subcommands

dandavison:
Closes https://github.com/temporalio/cli/issues/597

This PR changes the CLI API so that `workflow update` is not itself a leaf command, but has two leaf subcommands: `start` and `execute`. Previously, `workflow update` had the behavior of what is now `workflow update`, so this is a breaking change to the not-yet-GA workflow update feature.

`workflow update start` will output the update ID. In order to block and fetch the result of the update, users can do `workflow update execute` passing the started update ID. See https://github.com/temporalio/cli/issues/647 for relevant plans to evolve the CLI update API, creating a new way to fetch the result.

### How this was tested

- Manually against a Python sample
- PR adds integration tests

### DIFF
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -187,7 +187,15 @@ func (c *TemporalWorkflowTerminateCommand) run(cctx *CommandContext, _ []string)
 }
 
 func (c *TemporalWorkflowUpdateStartCommand) run(cctx *CommandContext, args []string) error {
-	return workflowUpdateHelper(cctx, c.Parent.Parent.ClientOptions, c.PayloadInputOptions, c.UpdateOptions, client.WorkflowUpdateStageAccepted)
+	waitForStage := client.WorkflowUpdateStageUnspecified
+	switch c.WaitForStage.Value {
+	case "accepted":
+		waitForStage = client.WorkflowUpdateStageAccepted
+	}
+	if waitForStage != client.WorkflowUpdateStageAccepted {
+		return fmt.Errorf("invalid wait for stage: %v, valid values are: 'accepted'", c.WaitForStage)
+	}
+	return workflowUpdateHelper(cctx, c.Parent.Parent.ClientOptions, c.PayloadInputOptions, c.UpdateOptions, waitForStage)
 }
 
 func (c *TemporalWorkflowUpdateExecuteCommand) run(cctx *CommandContext, args []string) error {
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -1,7 +1,9 @@
 package temporalcli
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os/user"
 
@@ -234,6 +236,14 @@ func workflowUpdateHelper(cctx *CommandContext,
 		return fmt.Errorf("unable to update workflow: %w", err)
 	}
 	if waitForStage == client.WorkflowUpdateStageAccepted {
+		// Use a canceled context to check whether the initial server response
+		// shows that the update has _already_ failed, without issuing a second request.
+		ctx, cancel := context.WithCancel(cctx)
+		cancel()
+		err = updateHandle.Get(ctx, nil)
+		if err != nil && !errors.Is(err, context.Canceled) {
+			return fmt.Errorf("unable to update workflow: %w", err)
+		}
 		return cctx.Printer.PrintStructured(
 			struct {
 				Name     string `json:"name"`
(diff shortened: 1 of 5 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -2521,19 +2521,23 @@ temporal workflow trace \
...
+### temporal workflow update: Start and wait for Updates
+An Update is a synchronous call to a Workflow Execution that can change its
+state, control its flow, and return a result.
 
-Send a message to a Workflow Execution to invoke an update handler. An update
-can change the state of a Workflow Execution and return a response:
+### temporal workflow update execute: Send an Update and wait for it to complete
+Send a message to a Workflow Execution to invoke an Update handler, and wait for
+the update to complete or fail. You can also use this to wait for an existing
+update to complete, by submitting an existing update ID.
```
dandavison, line 2531: This line will be changed when https://github.com/temporalio/cli/issues/647 is implemented.

temporalcli/commands.workflow.go
```diff
@@ -187,33 +186,53 @@ func (c *TemporalWorkflowTerminateCommand) run(cctx *CommandContext, _ []string)
...
+
+func (c *TemporalWorkflowUpdateExecuteCommand) run(cctx *CommandContext, args []string) error {
+	return workflowUpdateHelper(cctx, c.Parent.Parent.ClientOptions, c.PayloadInputOptions, c.UpdateOptions, client.WorkflowUpdateStageCompleted)
+}
+
+func workflowUpdateHelper(cctx *CommandContext,
+	clientOpts ClientOptions,
+	inputOpts PayloadInputOptions,
+	updateOpts UpdateOptions,
+	waitForStage client.WorkflowUpdateStage,
```
cretz, line 201: Didn't we decide for start that we are going to require this be provided by users in hopes that later we can default this to durable admitted?

dandavison: Oh that's right! With the need to add new subcommands I completely forgot about that. Good catch.

temporalcli/commands.workflow_test.go
```diff
@@ -407,51 +427,108 @@ func (s *SharedServerSuite) TestWorkflow_Update() {
...
+	if t.useStart {
+		// update rejected, name not supplied
+		res = t.s.Execute("workflow", "update", "start", "--wait-for-stage", "accepted", "--address", t.s.Address(), "-w", run.GetID(), "-i", strconv.Itoa(input))
+		t.s.ErrorContains(res.Err, "required flag(s) \"name\" not set")
+
+		// update rejected, wrong workflowID
+		res = t.s.Execute("workflow", "update", "start", "--wait-for-stage", "accepted", "--address", t.s.Address(), "-w", "nonexistent-wf-id", "--name", updateName, "-i", strconv.Itoa(input))
+		t.s.ErrorContains(res.Err, "unable to update workflow")
+
+		// TODO: wrong update name is not currently an error when using `update start` (?!)
```
dandavison, line 476: - [ ] TODO (blocking): investigate

temporalcli/commandsmd/commands.md
```diff
@@ -2521,19 +2521,23 @@ temporal workflow trace \
 
 Includes options set for [workflow reference](#options-set-for-workflow-reference).
 
-### temporal workflow update: Synchronously run a Workflow update handler
+### temporal workflow update: Start and wait for Updates
```
cretz, line 2524: IMO we should mark all of these commands as experimental similar to their SDK counterparts

temporalcli/commands.workflow.go
```diff
@@ -187,33 +186,61 @@ func (c *TemporalWorkflowTerminateCommand) run(cctx *CommandContext, _ []string)
...
+		return cctx.Printer.PrintStructured(
+			struct {
+				Name     string `json:"name"`
+				UpdateID string `json:"updateId"`
+			}{Name: updateOpts.Name, UpdateID: updateHandle.UpdateID()},
+			printer.StructuredOptions{})
+	}
 
 	var valuePtr interface{}
 	err = updateHandle.Get(cctx, &valuePtr)
```
cretz, line 246: Note, this update error may contain lots of valuable information beyond just the message appended to "unable to update workflow: ". It may have details, causes, etc. Don't need to solve now (not solved for query either), but we should probably open an issue to provide a better dump of failure upon failure. I think for workflow execute we dump the entire close event.

cretz (approved):

temporalcli/commands.workflow.go
dandavison: https://github.com/temporalio/cli/issues/651

temporalcli/commands.workflow_test.go
dandavison: Fixed in latest commits; thanks @Quinn-With-Two-Ns and @cretz for discussion there.

cretz (approved):

PR #650, opened 2024-08-29
Dump junit.xml

dandavison:
This PR switches the tests in CI to dump junit XML files. To do so, it switches to use https://github.com/gotestyourself/gotestsum as the test runner. (This is what the server repo uses also). An example of a build with test failures is here: https://github.com/temporalio/cli/actions/runs/10631569269/job/29472736222

As in https://github.com/temporalio/sdk-python/pull/617, the motivation for this is that, given appropriate tools, it will make it easier for us to study patterns of flakiness, test timings etc, across matrix builds within a run, and across runs.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,8 @@ jobs:
       - name: Install gotestsum
         run: go install gotest.tools/gotestsum@latest
 
-      - run: mkdir junit-xml
+      - name: Create junit-xml directory
+        run: mkdir junit-xml
 
       - name: Test
         run: gotestsum --junitfile junit-xml/${{matrix.os}}.xml -- ./...
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,6 +35,7 @@ jobs:
       - run: mkdir junit-xml
 
       - name: Test
+        shell: bash
         run: go test ./... 2>&1 | tee >(go-junit-report -set-exit-code > junit-xml/${{matrix.os}}.xml)
 
       - name: 'Upload junit-xml artifacts'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -29,8 +29,21 @@ jobs:
...
           go-version: "1.22"
 
+      - name: Install go-junit-report
+        run: go install github.com/jstemmer/go-junit-report/v2@latest
+
+      - run: mkdir junit-xml
+
       - name: Test
-        run: go test ./...
+        run: go test ./... 2>&1 | tee >(go-junit-report -set-exit-code > junit-xml/${{matrix.os}}.xml)
```
dnr, line 39: In the server we use `https://gotest.tools/gotestsum` as a test running, which has a junit output mode. That seems better than parsing `go test` output?

dandavison: Ah, thanks! I'll switch.

.github/workflows/ci.yaml
```diff
@@ -27,10 +27,23 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.22"
+          go-version: '1.22'
+
+      - name: Install gotestsum
+        run: go install gotest.tools/gotestsum@latest
+
+      - run: mkdir junit-xml
```
cretz, line 35: I think we should name all steps (or just make the step below be a two-command thing with `run: |` and this be the first command). Not all of our CIs name all steps, but in workflows that do we should probably keep doing it.

dandavison: Sure! Done. (I was thinking based on `sdk-python` that you liked not naming trivial shell steps but this seems good.)

cretz: I would support changing that CI to start naming everything too (that things are not named is a product of history I think)

cretz (approved):

PR #653, opened 2024-09-04
Fix register search attributes when starting server

rodrigozhou:
## What was changed

Fix register search attributes when starting server

## Why?

Currently, search attributes are registered after the server is ready, which can cause race condition with other processes waiting for the server to become ready too. This PR fixes this by registering the search attributes together with the namespaces.

Address https://github.com/temporalio/temporal/issues/6195

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


go.mod
```diff
@@ -14,23 +14,23 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
 	github.com/temporalio/ui-server/v2 v2.28.0
-	go.temporal.io/api v1.34.0
+	go.temporal.io/api v1.39.0
 	go.temporal.io/sdk v1.27.0
-	go.temporal.io/server v1.24.2
-	google.golang.org/grpc v1.65.0
+	go.temporal.io/server v1.26.0-120.0.20240904002151-40431ec95546
```
cretz, line 19: We won't be able to merge this until it is in a tagged version. If it only lands in a non-stable tagged version (i.e. 1.26 series and not one of the 1.25 patches), we will need to change the branch this is targeting to `next` (once that branch is created).

cretz (commented):
Thanks! LGTM, but not approving because we cannot merge an untagged server version.

go.mod
josh-berry: The branch exists (it's called `next-server`).

temporalcli/commands.server.go
```diff
@@ -149,11 +150,6 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 	}
 	defer s.Stop()
 
-	// Register search attributes
-	if err := t.registerSearchAttributes(cctx, searchAttrs, opts.Namespaces); err != nil {
```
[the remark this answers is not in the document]
rodrigozhou: Yes, we can. It's only used here.

go.mod
cretz: :+1: So once this is in a tagged non-GA version, this can be updated and the target if this PR can be changed to target that branch (or we can close and recreate the PR if needed, unsure if GH supports changing target branch).

rodrigozhou: Updated the base branch to `next-server`.

temporalcli/commands.server.go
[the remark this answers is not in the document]
rodrigozhou: Removed function.

cretz:
LGTM, but not marking this approved until on tagged server. There seem to be other tests the latest server is breaking.

cretz (approved):
This LGTM. We have now resolved to not make future server tags depending on SDK/API commit hashes. So when `next-server` makes it into `main` we should of course have stable tags for everything.

Issue #616, opened 2024-07-03
[Feature Request] Support query for temporal schedule list


dnr:
Support query for `temporal schedule list` using this new field: https://github.com/temporalio/api/blob/6cd92a2f214a2d14872af81e31b6eee2f8b745ca/temporal/api/workflowservice/v1/request_response.proto#L1157

cretz:
This is actually needed everywhere, not just CLI. Opened https://github.com/temporalio/features/issues/505.

PR #654, opened 2024-09-04
List schedules query param support

justinp-tt:
## What was changed

Support for schedule list query parameter

## Checklist

1. Closes 
https://github.com/temporalio/cli/issues/616

2. How was this tested:

Added new tests

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.schedule_test.go
+++ b/temporalcli/commands.schedule_test.go
@@ -11,8 +11,11 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/temporalio/cli/temporalcli"
 
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/operatorservice/v1"
 	"go.temporal.io/sdk/workflow"
 )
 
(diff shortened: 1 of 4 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -633,6 +633,7 @@ The `temporal schedule list` command lists all Schedules in a namespace.
 
 * `--long`, `-l` (bool) - Include detailed information.
 * `--really-long` (bool) - Include even more detailed information that's not really usable in table form.
+* `--query`, `-q` (string) - Filter results using a SQL-like query.
```
bergundy, line 636: Looks like we sometimes document query as SQL-like query and other times as "list filter".
I would go with the latter since that's the terminology we use in the docs: https://docs.temporal.io/visibility#list-filter.

CC @josh-berry in case you want to make this consistent in other places.

temporalcli/commands.schedule_test.go
```diff
@@ -249,6 +262,69 @@ func (s *SharedServerSuite) TestSchedule_List() {
 		ok = ok || j.ScheduleId == schedId
 	}
 	s.True(ok, "schedule not found in jsonl result")
+
+	// JSON query (match)
+
+	res = s.Execute(
```
bergundy, line 268: I think you should probably use `Eventually` here due to eventual consistency of the visibility store.

bergundy (approved):
Overall LGTM. Left a couple of small comments.

temporalcli/commands.schedule_test.go
justinp-tt: I rearranged the test assertions to ensure eventually visibility for the first case.

temporalcli/commandsmd/commands.md
justinp-tt: Changed; thanks

cretz (approved):

Issue #645, opened 2024-08-23
[Feature Request] Reduce test suite time


cretz:
### Describe the solution you'd like

Currently tests take 2m+ but they should only take a few seconds. I think there is a 90s sleep in there we should remove, but there may be other sleeps we should remove. We should try to be reactive to results instead of 1s+ sleeping (even if that means high-frequency polling). We can redesign tests or even reduce the strictness of our assertions if we must (hopefully not).

PR #656, opened 2024-09-05
Reduce test suite time

yuandrew:
## What was changed

Set cluster cache refresh rate to seconds instead of minutes. Now our test doesn't need to wait 60+ seconds for the cache to refresh

## Why?

Speed up tests

## Checklist

1. Closes 

#645
2. How was this tested:

Ran test locally

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.operator_cluster_test.go
+++ b/temporalcli/commands.operator_cluster_test.go
@@ -159,7 +159,7 @@ func (s *SharedServerSuite) TestOperator_Cluster_Operations() {
 	s.ContainsOnSameLine(out, fmt.Sprintf("\"clusterId\": \"%s\"", standbyCluster2.Options.ClusterID))
 
 	// Need to wait for the cluster cache to be updated
-	time.Sleep(90 * time.Second)
+	time.Sleep(1 * time.Second)
 
 	// Remove the clusters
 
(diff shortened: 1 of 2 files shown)


cretz (approved):
This is great, thanks! Ideally we have no sleeps in tests, but even this is a big win. Any flakiness concern on sleeping exactly the amount of the refresh impl? Should we sleep 2s to be sure or reduce the refresh interval (maybe only for this test)?

yuandrew:
> This is great, thanks! Ideally we have no sleeps in tests, but even this is a big win. Any flakiness concern on sleeping exactly the amount of the refresh impl? Should we sleep 2s to be sure or reduce the refresh interval (maybe only for this test)?

Changing refresh rate to milliseconds takes another second off of the test, having a buffer of 5 millisecond sleep seems to be good, no issues after running test locally 80 times

cretz:
Can you confirm with server any harm in changing this to 100ms just for our tests? Meaning, is this max expiry before refetch or is this some kind of loop interval. If no prob, lets do that.

yuandrew:
YX confirmed this is fine, only side effect is more DB IO

dandavison:
Nice.

PR #662, opened 2024-09-12
Update help text to conform to style

josh-berry:
- Clean up `temporal operator nexus` commands since they wrap beyond 80 columns (and this would be visible to the user).
- Clarify wording on `temporal task-queue describe`, which was rather confusing in some places.

Both of these got scrutinized because they changed in the `nexus` branch, but not in `main`, so they would not have been seen by our docs folks.


bergundy (approved):
Feels like there's some redundant new lines now but if that's inline with the other commands I don't have any objections.

josh-berry:
> Feels like there's some redundant new lines now but if that's inline with the other commands I don't have any objections.

It's in line with the others, yeah. I agree with you it's a little weird, but it was helpful to distinguish things that are semantically meaningful (like `Required.`) vs. things that aren't. This should get cleaned up in general when we move to a YAML format.

Thanks for taking a look!

PR #663, opened 2024-09-12
Merge nexus branch into main

josh-berry:
This pulls in the new `operator nexus ...` suite of commands, as well as updated Server and UI Server versions. Most of what's here has been reviewed already in separate, smaller PRs, hence why this is a merge commit and not a squash.

### DIFF
--- a/temporalcli/commandsmd/commands.md
+++ b/temporalcli/commandsmd/commands.md
@@ -543,7 +543,7 @@ commands.
 ### temporal operator: Manage Temporal deployments
 
 Operator commands manage and fetch information about Namespaces, Search
-Attributes, and Temporal Services:
+Attributes, Nexus Endpoints, and Temporal Services:
 
 ```
 temporal operator [command] [subcommand] [options]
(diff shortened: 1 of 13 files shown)


josh-berry:
This whole branch does NOT need to be re-reviewed since everything in here is either the result of a merge commit, or covered under separate PRs.

bergundy (approved):

cretz (approved):

temporalcli/commandsmd/commands.md
```diff
@@ -1375,6 +1515,26 @@ temporal task-queue describe \
     --task-queue-type "activity"`
 ```
 
+This command provides the following task queue statistics:
+- `ApproximateBacklogCount`: The approximate number of tasks backlogged in this
+  task queue.  May count expired tasks but eventually converges to the right
+  value.
+- `ApproximateBacklogAge`: How far "behind" the task queue is running. This is
+  the difference in age between the oldest and newest tasks in the backlog,
```
ShahabT, line 1523: This is not accurate. The backlog age is the age of the oldest task in the backlog, period. What is written here suggest that backlog age of a backlog with a single task will be zero, as an example.

temporalcli/commandsmd/commands.md
```diff
@@ -1375,6 +1515,26 @@ temporal task-queue describe \
...
+  measured in seconds.
+- `TasksAddRate`: Approximate rate at which tasks are being added to the task
+  queue, measured in tasks per second, averaged over the last 30 seconds.
+  Includes tasks dispatched immediately without going to the backlog
+  (sync-matched tasks), as well as tasks added to the backlog.
+- `TasksDispatchRate`: Approximate rate at which tasks are being dispatched from
```
ShahabT, line 1529: These two metrics can be significantly off from what the user might expect due to eager tasks and sticky queues. I think it's worth to mention something about that and link to the docs page that explains the inaccuracy in more details. Something like: "Not all tasks dispatched to the workers are counted in this metric, for more information please see [docs]."

PR #561, opened 2024-05-14
EDU-2415: Improve CLI Help content

fairlydurable:
This is up to date as of mid-day 23 July. 

Note: This will not compile in its current state. 
Ready for Josh's review.

PR #665, opened 2024-09-13
Remove Makefile

cretz:
## What was changed

To be consistent with other SDK repos, remove `Makefile` that was added in an unrelated PR at #561. May be reintroduced, but should at least get wider consensus and be called in CI to confirm it even works.


dandavison (changes requested):
I find the Makefile helpful, and so I view it as an incremental step in the right direction, not as a violation of a previously existing policy. So that suggests discussion before removal.

josh-berry:
I have not forgotten about this.

If we remove it, I would like to replace it with something platform-independent and standardized (as much as possible in the Go ecosystem anyway) that accomplishes the same goal. Otherwise I think having it (and maybe adding a script for Windows folks) is better than nothing.

cretz:
> If we remove it, I would like to replace it with something platform-independent and standardized (as much as possible in the Go ecosystem anyway) that accomplishes the same goal

This exists and is called `go build` and `go run` and was an approved method of working with the CLI until this Makefile got added in an unrelated PR sans discussion.

josh-berry:
> This exists and is called `go build` and `go run`

Dan and I have both explained why this is not the same thing as the Makefile--you might want to review that Slack discussion. If you need a pointer to it, poke me separately and I'll find it again for you.

cretz:
:+1: Definitely understood it's not the same thing as a Makefile, on purpose, it's better, consistent, more cross-platform, etc. There was a CLI way, this secretly changed, and looks like it's not going back.

PR #666, opened 2024-09-13
Use YAML for CLI command generation

### LINKED ISSUE #620: Switch to a common format for CLI command generation
We should switch to a more commonly-understood format for CLI command generation that is easily consumable by "standard" parsers (like a Markdown or YAML or JSON parser). The custom parser thing we have now is quite easy to misuse—for example, there is a semantic difference between `Required` and `Required.`, and between `Default is foo` and `Default: foo.`

After some internal discussion, YAML is probably the preferred approach, with Markdown for item descriptions. For example:

```yaml
- temporal env set:
    summary: Set environment properties
    description: |
      `temporal env set --env environment -k property -v value`

      Property names match CLI option names, for example '--address' and '--tls-cert-path':

      [...]
    maximum-args: 2
    ignores-missing-env: true
    options:
      - [-k, --key]:
        type: string
        description: The name of the property.
        required: true
      - [-v, --value]:
        type: string
        description: The value to set the property to.
        required: true
```

Another option might be to embed YAML within Markdown, however this option was discarded as it's still more complex to parse:

```markdown
### temporal env set: Set environment properties.

`temporal env set --env environment -k property -v value`

Property names match CLI option names, for example '--address' and '--tls-cert-path':

`temporal env set --env prod -k address -v 127.0.0.1:7233`
`temporal env set --env prod -k tls-cert-path -v /home/my-user/certs/cluster.cert`

If the environment is not specified, the `default` environment is used.

```yaml
maximum-args: 2
ignores-missing-env: true
options:
  - [-k, --key]:
    type: string
    description: The name of the property.
    required: true
  - [-v, --value]:
    type: string
    description: The value to set the property to.
    required: true
```

(Note: Exact structure of the YAML data is still TBD; the above examples show one possibility but there are others, to be decided at implementation time.)

I also considered but discarded XML because it's significantly harder to work with in code, and therefore much less commonly-used nowadays. I also considered but discarded JSON because it's not flexible enough in its plain-text representation; we expect to have a lot of multi-line string literals, and JSON does not do well with those.

YAML does have some caveats, but we still think it's the best choice despite those caveats:

- We still don't get the nice editor formatting/syntax-highlighting for Markdown within the `description` fields
- It's almost too flexible in some ways; its duck-typing can be surprising, and there are subtleties to the indentation that can be hard to get right. However, most of this can be handled with strict validation at parse time (since we are parsing the doc into a more strongly-typed structure internally before emitting CLI commands).

Further comments and opinions are of course welcome. :)

yuandrew:
## What was changed
Moved from Markdown to YAML for CLI command generation.

This switch also fixes a bug where option set aliases weren't being persisted to commands that use them (i.e. `NewTemporalScheduleCreateCommand`)

## Why?
More standardized format, easier to parse and add to

## Checklist

1. Closes #620 

2. How was this tested:

Passes all CI tests

3. Any docs updates needed?


temporalcli/commandsmd/commands.yml
```diff
@@ -0,0 +1,2668 @@
...
+        options:
+          - disabled
+          - enabled
+        default: disabled
+      - name: visibility-uri
+        type: string
+        description: |
+          Archive visibility data to this `URI`.
+          Once enabled, can't be changed.
+  - name: temporal operator namespace delete
```
yuandrew, line 557: I've removed the `[namespace]` part of the title and moved it into a flag. From what I've observed, this positional argument is not enforced (-y flag can be specified after the [namespace] flag, contrary to what the help doc says). 

I've also removed the same `[namespace]` part from `temporal operator namespace describe` below. In the same vein, it seems to me that we can have the same behavior with `namespace` being a flag, and with the removal of these 2, we no longer need to support the ability to specify a `[argument]` in the title, simplifying the template

yuandrew: Josh has confirmed that the positional form is deprecated, safe to remove

yuandrew:
the new format does not allow enum options for a `string[]` type, so all of the options mentioned in #670 will be temporarily changed to honor their current type (i.e. `string[]` will accept any string value), with the description containing the enum options.

temporalcli/commandsmd/commands.yml
```diff
@@ -0,0 +1,2890 @@
...
+    option-sets:
+      - workflow-reference
+    options:
+      - name: reset-points
+        type: bool
+        description: Show auto-reset points only.
+      - name: raw
+        type: bool
+        description: Print properties without changing their format.
+        default: true
```
yuandrew, line 2051: Description for this option used to be 
```
* `--raw` (bool) -
  Print properties without changing their format.
  Defaults to true.
```
Which states defaults to `true`, but the code didn't seem to support that. Should this be removed from the description? A bool that defaults to true feels weird/doesn't make sense. But is this API changable at this point?

temporalcli/commandsmd/commands.yml
```diff
@@ -0,0 +1,3023 @@
...
+        type: string
+        description: Task Queue name.
+        required: true
+        short: t
+  - name: temporal task-queue update-build-ids promote-id-in-set
+    summary: Set Build ID as set default (Deprecated)
+    description: |
+      +-----------------------------------------------------------------------------+
+      | CAUTION: This command is deprecated and will be removed in a later release. |
+      +-----------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
yuandrew: Good suggestion, created #673

temporalcli/commandsmd/commands.yml
```diff
@@ -0,0 +1,3023 @@
...
+        description: Task Queue name.
+        required: true
+        short: t
+  - name: temporal task-queue versioning
+    summary: Manage Task Queue Build ID handling (Experimental)
+    description: |
+      +---------------------------------------------------------------------+
+      | CAUTION: Worker versioning is experimental. Versioning commands are |
+      | subject to change.                                                  |
+      +---------------------------------------------------------------------+
```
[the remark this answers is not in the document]
yuandrew: Created #674

temporalcli/commands.gen.go
```diff
@@ -194,9 +413,9 @@ func NewTemporalBatchDescribeCommand(cctx *CommandContext, parent *TemporalBatch
 	s.Command.Use = "describe [flags]"
 	s.Command.Short = "Show batch job progress"
 	if hasHighlighting {
-		s.Command.Long = "Show the progress of an ongoing batch job. Pass a valid job ID to display its\ninformation:\n\n\x1b[1mtemporal batch describe \\\n    --job-id YourJobId\x1b[0m"
+		s.Command.Long = "Show the progress of an ongoing batch job. Pass a valid job ID to display its\ninformation:\n\n\x1b[1mtemporal batch describe \\\n    --job-id YourJobId\x1b[0m\n"
 	} else {
-		s.Command.Long = "Show the progress of an ongoing batch job. Pass a valid job ID to display its\ninformation:\n\n```\ntemporal batch describe \\\n    --job-id YourJobId\n```"
+		s.Command.Long = "Show the progress of an ongoing batch job. Pass a valid job ID to display its\ninformation:\n\n```\ntemporal batch describe \\\n    --job-id YourJobId\n```\n"
```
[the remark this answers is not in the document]
yuandrew: stripped

temporalcli/commands.gen.go
```diff
@@ -995,15 +1214,15 @@ func NewTemporalOperatorNexusEndpointUpdateCommand(cctx *CommandContext, parent
...
 	}
 	s.Command.Args = cobra.NoArgs
 	s.Command.Flags().StringVar(&s.Name, "name", "", "Endpoint name. Required.")
 	_ = cobra.MarkFlagRequired(s.Command.Flags(), "name")
-	s.Command.Flags().StringVar(&s.Description, "description", "", "Endpoint description in markdown format (encoded using the configured codec server).")
-	s.Command.Flags().StringVar(&s.DescriptionFile, "description-file", "", "Endpoint description file in markdown format (encoded using the configured codec server).")
+	s.Command.Flags().StringVar(&s.Description, "description", "", "Endpoint description in markdown format (encoded using the configured codec server).\n")
+	s.Command.Flags().StringVar(&s.DescriptionFile, "description-file", "", "Endpoint description file in markdown format (encoded using the configured codec server).\n")
```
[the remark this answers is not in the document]
yuandrew: good point, I've removed all `\n` from the options, generated output should match previous markdown formatting

temporalcli/commands.gen.go
```diff
@@ -51,7 +270,7 @@ func NewTemporalCommand(cctx *CommandContext) *TemporalCommand {
 	s.Command.PersistentFlags().StringVar(&s.EnvFile, "env-file", "", "Path to environment settings file. (defaults to `$HOME/.config/temporalio/temporal.yaml`).")
 	s.LogLevel = NewStringEnum([]string{"debug", "info", "warn", "error", "never"}, "info")
 	s.Command.PersistentFlags().Var(&s.LogLevel, "log-level", "Log level. Default is \"info\" for most commands and \"warn\" for `server start-dev`. Accepted values: debug, info, warn, error, never.")
-	s.Command.PersistentFlags().StringVar(&s.LogFormat, "log-format", "", "Log format. Options are: text, json. Defaults to: text.")
+	s.Command.PersistentFlags().StringVar(&s.LogFormat, "log-format", "text", "Log format. Options are: text, json.")
```
josh-berry, line 273: There was some reason we didn't want the default to be programmatically-encoded that I can't remember. Can you double-check this and make sure it's not changing behavior somehow?

josh-berry (approved):
I'll be honest that I didn't read the parser/generator changes super closely; I mostly looked at the generated output, which seems good to me. There's only one small comment about the `log-format` flag that's not a blocker for merging IMO. Otherwise LGTM!

josh-berry:
Actually, one other thought: let's wait for @cretz to take a look before merging in case he spots something I didn't.

temporalcli/commandsmd/commands.yml
cretz, line 1: IMO this entire package/dir should be changed to `commandsyaml` (and contributing guide and such changed as well)

cretz (commented):
This looks great and basically exactly how I would have done it. Only comment worth noting is the request to change the package/dir name, everything else is non-blocking. Don't forget to update `CONTRIBUTING.md`.

temporalcli/commandsmd/commands.yml
```diff
@@ -0,0 +1,3023 @@
...
+        type: string-enum
+        enum-values:
+          - debug 
+          - info
+          - warn
+          - error
+          - never
+        description: |
+          Log level.
+          Default is "info" for most commands and "warn" for `server start-dev`.
```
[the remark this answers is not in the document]
josh-berry: No, I want to get rid of this eventually and use a proper Markdown renderer for both the description and flags. The reason it was needed for flags is because newlines show up in the generated output, so having newlines inserted in (to-the-user) random places really doesn't look good. I think Andrew has since fixed this? In which case I have no opinion.

temporalcli/commandsmd/parse.go
```diff
@@ -7,62 +7,80 @@ import (
...
+	"slices"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 )
 
-//go:embed commands.md
-var CommandsMarkdown []byte
+//go:embed commands.yml
+var CommandsYaml []byte
```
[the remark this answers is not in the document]
josh-berry: (This was my bad; I think I said `Yaml` 'cause I couldn't remember what Go does by default and guessed wrong, heh)

temporalcli/commandsmd/commands.yml
josh-berry: Not to bikeshed too hard on the name, but how about `commands-gen` or something that doesn't reference the format name?

temporalcli/commandsmd/commands.yml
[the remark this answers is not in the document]
cretz: I think we should have a general/consistent rule for Markdown `description`. Right now it seems a bit inconsistent.

temporalcli/commandsmd/commands.yml
cretz: Works for me, though you should not have a `-` in a Go package dir name, but `commandsgen` is fine

temporalcli/commandsmd/commands.yml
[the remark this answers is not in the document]
josh-berry: Agree, eventually; we can make this consistent once we have a proper Markdown formatter IMO.

temporalcli/commandsgen/parse.go
```diff
@@ -0,0 +1,192 @@
+// Package commandsgen is built to read the markdown format described in
+// temporalcli/commands.md and generate code from it.
```
cretz, line 2: This statement is not accurate anymore

Makefile
```diff
@@ -4,7 +4,7 @@ all: gen build
 
 gen: temporalcli/commands.gen.go
 
-temporalcli/commands.gen.go: temporalcli/commandsmd/commands.md
+temporalcli/commands.gen.go: temporalcli/commandsgen/commands.md
```
cretz, line 7: This still references `commands.md`. In a perfect world this would have failed CI with something like:

> No rule to make target 'temporalcli/commandsgen/commands.md', needed by 'temporalcli/commands.gen.go'

But unfortunately this Makefile is an untested/undocumented thing just sitting in the repo

yuandrew: oops, thanks for catching this!

cretz (approved):

PR #671, opened 2024-09-16
Cap UI port at max value

Sushisource:
## What was changed
Cap maximum UI port value

## Why?
Overflow bad

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.server.go
+++ b/temporalcli/commands.server.go
@@ -73,6 +73,9 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 		}
 		if opts.UIPort == 0 {
 			opts.UIPort = t.Port + 1000
+			if opts.UIPort > 65535 {
+				opts.UIPort = 65535
+			}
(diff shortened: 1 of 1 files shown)


Sushisource:
Relates to https://github.com/temporalio/cli/pull/671

cretz (approved):

temporalcli/commands.server.go
```diff
@@ -73,6 +73,9 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 		}
 		if opts.UIPort == 0 {
 			opts.UIPort = t.Port + 1000
+			if opts.UIPort > 65535 {
+				opts.UIPort = 65535
+			}
```
josh-berry, line 78: This increases the probability of 65535 being chosen randomly (vs. other port numbers). Can we instead do something like:

```suggestion
			if opts.UIPort > 65535 {
				opts.UIPort = (t.Port - 1024) % (65536 - 1024) + 1024
			}
```

(why keep the if-statement at all? To preserve backward compatibility with the existing `+1000` behavior)

PR #675, opened 2024-09-23
Batch rate limit

stephanos:
## What was changed

Rebased https://github.com/temporalio/cli/pull/551 that was never actually merged.

### DIFF
--- a/temporalcli/commands.workflow.go
+++ b/temporalcli/commands.workflow.go
@@ -5,8 +5,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"go.temporal.io/sdk/converter"
 	"os/user"
+
+	"go.temporal.io/sdk/converter"
 
 	"github.com/fatih/color"
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


temporalcli/commandsmd/commands.md
```diff
@@ -2530,6 +2530,7 @@ Includes options set for [payload input](#options-set-for-payload-input).
 * `--yes`, `-y` (bool) -
   Don't prompt to confirm signaling.
   Only allowed when --query is present.
+* `--rps` (float) - Limit batch's requests per second. Only allowed if query is present.
```
josh-berry, line 2533: This will need to be rebased again since Andrew just merged an update that converted this file to YAML. :/ Hopefully will be straightforward, but let me know if it gets gnarly and/or you'd like me to do it.

josh-berry (approved):

temporalcli/commandsmd/commands.md
stephanos: @josh-berry okay, rebased! Should be good now. 🤞

temporalcli/commands.workflow.go
```diff
@@ -442,11 +444,17 @@ func (s *SingleWorkflowOrBatchOptions) workflowExecOrBatch(
 		reason = defaultReason()
 	}
 
+	// Check rps is used together with query
+	if s.Rps != 0 && s.Query == "" {
```
cretz, line 448: This condition will never be true down here (`Query` is always non-empty down here, it's checked above)

cretz (approved):

temporalcli/commands.workflow.go
stephanos: Good callout 👍 fixed!

josh-berry (approved):
👍 Thanks for forward-porting and sorry I missed this before!

PR #676, opened 2024-09-23
Add global `context-rpc-timeout` option

### LINKED ISSUE #648: [Feature Request] Heavy remote codecs need longer timeouts

There is a use case where a slow remote codec can timeout on `workflow show`. The Go SDK defaults to 10s for get history events call for each page, and that 10s is for the context for the whole gRPC call and remote codec is implemented as an interceptor within that.

The beta version of the CLI and tctl had a `--context-timeout` that allowed users to increase the timeout (which the Go SDK respects because by using max(1s, user-timeout / 2)). We can add this in back in (though would suggest `--rpc-timeout`) or find other solutions. 

We can also consider an option to concurrently call the codec with all payloads needed instead of serially in a visitor like we do today, though that's not necessary for this issue.

yuandrew:
## What was changed

Added a new global RPC timeout for within a context.

## Why?

A slow remote codec can cause timeouts

## Checklist

1. Closes #648

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -50,10 +50,6 @@ type CommandContext struct {
 	JSONOutput            bool
 	JSONShorthandPayloads bool
 
-	// Cancel function that comes from manually adding a deadline for RPC
-	// timeout within a context call
-	RpcTimeoutCancel context.CancelFunc
-
 	// Is set to true if any command actually started running. This is a hack to workaround the fact
 	// that cobra does not properly exit nonzero if an unknown command/subcommand is given.
 	ActuallyRanCommand bool
@@ -356,11 +352,6 @@ func Execute(ctx context.Context, options CommandOptions) {
 		cctx.Options.Fail(err)
 	}
 
-	// Clean up resources from setting a context RPC timeout (context-rpc-timeout)
-	if cctx.RpcTimeoutCancel != nil {
-		cctx.RpcTimeoutCancel()
-	}
-
 	// If no command ever actually got run, exit nonzero with an error.  This is
 	// an ugly hack to make sure that iff the user explicitly asked for help, we
 	// exit with a zero error code.  (The other situation in which help is
@@ -504,13 +495,8 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
 		}
 	}
 	cctx.JSONShorthandPayloads = !c.NoJsonShorthandPayloads
-
-	if c.RpcTimeout.Duration() != time.Second*10 {
-		// We multiply the timeout by 2 because the client SDK sets the RPC timeout
-		// to half of the Deadline
-		newDeadline := time.Now().Add(c.RpcTimeout.Duration() * 2)
-		cctx.Context, cctx.RpcTimeoutCancel = context.WithDeadline(cctx.Context, newDeadline)
-	}
+	fmt.Println("rpcTimeout", c.CommandTimeout.Duration())
+	cctx.Context, _ = context.WithDeadline(cctx.Context, time.Now().Add(c.CommandTimeout.Duration()))
 
 	return nil
 }
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -50,6 +50,10 @@ type CommandContext struct {
 	JSONOutput            bool
 	JSONShorthandPayloads bool
 
+	// Cancel function that comes from manually adding a deadline for RPC
+	// timeout within a context call
+	ContextRpcTimeoutCancel context.CancelFunc
+
 	// Is set to true if any command actually started running. This is a hack to workaround the fact
 	// that cobra does not properly exit nonzero if an unknown command/subcommand is given.
 	ActuallyRanCommand bool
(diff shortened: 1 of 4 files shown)


cretz:
This needs to be a general context timeout for the whole system IMO (left unset defaults to per-call of 10s). This is not specific to `workflow show`. Not sure we want to call it a `context-timeout`, unsure of a better name though.

cretz:
> That would let folks know it's the rpc timeout from within a context?

The problem is that "context" is a Go thing and people using the CLI don't care what language it is written in. I think we can probably have a "rpc timeout" though and document that it spans all RPCs on the CLI. Would be open to any other naming too that represents Go context timeout for life of the command without expecting people to even know we're using Go (but a bit afraid of generic "timeout" because it's only for things using the context).

temporalcli/commandsgen/commands.yml
```diff
@@ -195,6 +195,10 @@ commands:
       - name: no-json-shorthand-payloads
         type: bool
         description: Raw payload output, even if they are JSON.
+      - name: context-rpc-timeout
+        type: duration
+        description: Timeout for the underlying RPC call for a context call.
+        default: 20s
```
yuandrew, line 201: Since Go SDK does [Deadline() / 2](https://github.com/temporalio/sdk-go/blob/master/internal/internal_utils.go#L129), setting default to 20s to match the Go SDK's default of 10s

yuandrew: Actually, that can be misleading for the user. I'll instead not have a default for this param, and instead include a blurb in the description that the client defaults to 10s if unspecified

yuandrew:
> we need to explain that it's not necessarily per RPC

Can you explain what this means? Or does this no longer apply now that I moved this to be a global option, and it applies to all RPC calls for the CLI?

temporalcli/commands.go
```diff
@@ -495,6 +504,14 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
 		}
 	}
 	cctx.JSONShorthandPayloads = !c.NoJsonShorthandPayloads
+
+	if c.RpcTimeout.Duration() != time.Second*10 {
```
cretz, line 508: I don't think we need this conditional

temporalcli/commands.go
```diff
@@ -495,6 +504,14 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
...
+		// We multiply the timeout by 2 because the client SDK sets the RPC timeout
+		// to half of the Deadline
+		newDeadline := time.Now().Add(c.RpcTimeout.Duration() * 2)
```
cretz, line 511: This is why I was afraid of calling it "RPC timeout". There are many CLI commands that make multiple RPC calls. Doubling the user timeout makes it no longer accurate to what the user supplied. The reason the Go SDK halves it is to retry, but i am not sure we need to affect the entire context that governs timeout for the whole CLI command with that.

Maybe "context timeout" is an acceptable term, or "command timeout" or something.

temporalcli/commands.go
```diff
@@ -495,6 +504,14 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
...
+		cctx.Context, cctx.RpcTimeoutCancel = context.WithDeadline(cctx.Context, newDeadline)
```
cretz, line 512: I think we can count on the parent context cancelling, so we can ignore this cancel callback

temporalcli/commandsgen/commands.yml
```diff
@@ -195,6 +195,10 @@ commands:
       - name: no-json-shorthand-payloads
         type: bool
         description: Raw payload output, even if they are JSON.
+      - name: rpc-timeout
+        type: duration
+        description: Timeout for all underlying RPC calls for the CLI.
+        default: 10s
```
cretz, line 201: This should not have a default IMO. This timeout is for the entire command, not just one RPC call. A command may span many RPC calls.

josh-berry: I also like "command timeout" or just "timeout" better than "rpc-timeout". If I am reading scripts, docs, etc. and I see `--rpc-timeout` with no other context, I will assume it's for a single RPC when it's for the entire command.

temporalcli/commands.go
```diff
@@ -495,6 +495,8 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
 		}
 	}
 	cctx.JSONShorthandPayloads = !c.NoJsonShorthandPayloads
+	cctx.Context, _ = context.WithDeadline(cctx.Context, time.Now().Add(c.CommandTimeout.Duration()))
```
cretz, line 498:
```suggestion
	if c.CommandTimeout.Duration() > 0 {
		cctx.Context, _ = context.WithTimeoutCause(
			cctx.Context,
			c.CommandTimeout.Duration(),
			fmt.Errorf("command timed out"),
		)
	}
```

I think this or something like it makes more sense than using deadline and always setting it even if not set by user.

cretz (approved):

PR #677, opened 2024-09-23
Create new `string-enum[]` type

### LINKED ISSUE #670: [Feature Request] Create new 'string-enum[]' type for the CLI command generation

There are a few command options in the CLI that are currently type `string[]`, but also have a set of options available. 
Some examples:

```
* `--reapply-exclude` (string[]) -
  Exclude these event types from re-application.
  Options: All, Signal, Update.
* `--task-queue-type` (string[]) -
  Task Queue type.
  Options are: workflow, activity, nexus.
```

Other options that fit this criteria:
- `log-format`
- `type` option for `temporal operator search-attribute create`

We should define a new type `string-enum[]` so we can more strictly check these types in our CLI command parser/generator.

yuandrew:
## What was changed

Created a new `string-enum[]` type

## Why?

Command option behavior already exists, there just wasn't a specific type for this yet.

## Checklist

1. Closes #670

2. How was this tested:

3. Any docs updates needed?


temporalcli/strings.go
```diff
@@ -33,6 +34,27 @@ func (s *StringEnum) Set(p string) error {
...
+}
+
+func NewStringEnumArray(allowed []string, values []string) StringEnumArray {
+	return StringEnumArray{Allowed: allowed, Values: values}
+}
+
+func (s *StringEnumArray) String() string { return strings.Join(s.Values, ",") }
+
+func (s *StringEnumArray) Set(p string) error {
+	if !slices.Contains(s.Allowed, p) {
```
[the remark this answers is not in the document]
yuandrew: yes! thank you for the reminder

temporalcli/strings.go
```diff
@@ -33,6 +34,32 @@ func (s *StringEnum) Set(p string) error {
...
+	return StringEnumArray{Allowed: allowed, Values: values}
+}
+
+func (s *StringEnumArray) String() string { return strings.Join(s.Values, ",") }
+
+func (s *StringEnumArray) Set(p string) error {
+	if !slices.Contains(s.Allowed, strings.ToLower(p)) {
+		return fmt.Errorf("invalid value: %s, allowed values are: %s", p, strings.Join(s.Allowed, ", "))
+	}
+	s.Values = append(s.Values, p)
```
dnr, line 57: maybe this should canonicalize what the user typed to the "correct" case? easy way would be make Allowed into a map from lower to correct, and then check by looking up

temporalcli/commandsgen/commands.yml
```diff
@@ -1026,12 +1030,17 @@ commands:
...
+        enum-values:
+          - Text
+          - Keyword
+          - Int
+          - Double
+          - Bool
+          - Datetime
+          - KeywordList
         required: true
-
```
josh-berry, line 1034: Removed by accident?

temporalcli/commandsgen/commands.yml
```diff
@@ -161,10 +161,14 @@ commands:
...
-        type: string
+        type: string-enum
         description: |
           Log format.
-          Options are: text, json.
+          "pretty" is an alias for "text". This is for compatibility reasons, please use "text".
+        enum-values:
+          - text
+          - json
+          - pretty
```
josh-berry, line 171: @cretz Was there some reason we wanted this not to be a string-enum? I remember you avoided it specifically for some reason but I think maybe that didn't get documented, so I'm not sure if it's relevant anymore.

josh-berry (approved):
LGTM other than please double-check with Chad on the log-format flag.

temporalcli/strings.go
```diff
@@ -33,6 +33,40 @@ func (s *StringEnum) Set(p string) error {
...
+
+func (s *StringEnumArray) String() string { return strings.Join(s.Values, ",") }
+
+func (s *StringEnumArray) Set(p string) error {
+	val, ok := s.Allowed[strings.ToLower(p)]
+	if !ok {
+		values := make([]string, 0, len(s.Allowed))
+		for _, v := range s.Allowed {
+			values = append(values, v)
+		}
```
dnr, line 62: just a suggestion if you like
```suggestion
		values := maps.Values(s.Allowed)
```
(from `"golang.org/x/exp/maps"`)

temporalcli/commandsgen/commands.yml
cretz: I think it is because we wanted "pretty" to be an _undocumented_ enumerate alias. I think #678 could help here.

temporalcli/strings.go
cretz: > from "golang.org/x/exp/maps"

I think it's in the standard library now w/ the iter stuff

dnr: They didn't include Keys and Values from x/exp/maps in the stdlib package. I'm not sure why, they're pretty useful :shrug:

PR #680, opened 2024-09-26
Make `--reapply-type All` reapply updates in addition to signals

### LINKED ISSUE #679: [Bug] reset reapply-type All should mean signals and updates but it is just requesting signals
This should be `reapplyType := enums.RESET_REAPPLY_TYPE_ALL_ELIGIBLE`

https://github.com/temporalio/cli/blob/e39881167c74600d8d874c88c60c41370e2d3dd3/temporalcli/commands.workflow_reset.go#L98-L108

dandavison:
- Fixes #679: we should have been reapplying updates by default but we were not
- Refactor and clean-up test
- Add missing test coverage

### DIFF
--- a/temporalcli/commands.workflow_reset_test.go
+++ b/temporalcli/commands.workflow_reset_test.go
@@ -248,67 +248,81 @@ func (s *SharedServerSuite) TestBatchResetByBuildId() {
 	sut.stopWorkerFor("v3")
 }
 
-func (s *SharedServerSuite) TestWorkflow_Reset_ReapplyExcludeSignal() {
-	var wfExecutions, timesSignalSeen int
+type WorkflowResetTest struct {
+	s                      *SharedServerSuite
+	reapplyType            string
+	reapplyExclude         []string
+	expectUpdatesReapplied bool
+	expectSignalsReapplied bool
+}
+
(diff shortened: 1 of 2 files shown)


temporalcli/commands.workflow_reset_test.go
```diff
@@ -248,27 +248,112 @@ func (s *SharedServerSuite) TestBatchResetByBuildId() {
...
+				signalHandlerExecutions++
 			}
 		})
-		err := workflow.Sleep(ctx, 1*time.Second)
+		// Handle updates
+		workflow.SetUpdateHandler(ctx, "myUpdate", func(ctx workflow.Context) error {
+			updateHandlerExecutions++
+			return nil
+		})
+		err := workflow.Sleep(ctx, 100*time.Millisecond)
```
dandavison, line 325: This test could be written without sleep. In fact, it would be sufficient to just check that the CLI is mapped correctly to gRPC requests -- reset itself is tested in the SDK and server. But that's all orthogonal to this PR.

cretz: So long as not flaky, no concerns here

cretz (approved):

Issue #681, opened 2024-09-27
[Bug] Sqlite pragmas are not applied correctly


vbehar:
### What are you really trying to do?

I'm trying to start a dev-server with specific sqlite settings (pragma), for example to enable WAL journal

### Describe the bug

- The parameters set on the CLI are not applied to the sqlite database
- Temporal CLI doesn't support all pragma statements supported by sqlite

### Minimal Reproduction

- `temporal server start-dev --db-filename temporal.sqlite --sqlite-pragma journal_mode=WAL`

this should create 3 files:
- `temporal.sqlite`
- `temporal.sqlite-shm`
- `temporal.sqlite-wal`

but only 1 is created: `temporal.sqlite` - which means that the `journal_mode=WAL` pragma statement has not been applied to the sqlite DB

also, not all pragma statements supported by sqlite in https://www.sqlite.org/pragma.html are accepted by temporal CLI. the CLI only support 2 statements: `journal_mode` and `synchronous`.

### Environment/Versions

- Temporal CLI Version: v1.1.0

### Additional context

- this code: https://github.com/temporalio/cli/blob/v1.1.0/temporalcli/devserver/server.go#L324-L329 accepts only 2 statements, but it's not clear why. I don't see any reason to restrict only to these 2...
- this line https://github.com/temporalio/cli/blob/v1.1.0/temporalcli/devserver/server.go#L330 adds an underscore `_` to the pragma key, but the sqlite plugin doesn't remove it before building the DSN https://github.com/temporalio/temporal/blob/release/v1.23.x/common/persistence/sql/sqlplugin/sqlite/plugin.go#L205 so we end up with an sqlite URL similar to `temporal.sqlite?_pragma=_journal_mode=WAL` instead of `temporal.sqlite?_pragma=journal_mode=WAL`

cretz:
Thanks! Looking at PR.

PR #682, opened 2024-09-27
fix(sqlite): support all sqlite pragma statement

vbehar:
## What was changed

- I removed the extra underscore in the key
- I removed the pragma key check

## Why?

- the extra underscore was breaking the pragma support, because sqlite couldn't recognise the pragma key
- the pragma key check was too restricting, users should be able to configure any pragma statement from https://www.sqlite.org/pragma.html - removing the check all to use any pragma, without adding more work to maintain an hardcoded list

## Checklist

1. Closes https://github.com/temporalio/cli/issues/681

2. How was this tested:

manually:
- by running `./temporal server start-dev --db-filename temporal.sqlite --sqlite-pragma journal_mode=WAL --sqlite-pragma synchronous=NORMAL --sqlite-pragma busy_timeout=5000`
- and ensuring that 3 files have been created: `temporal.sqlite`, `temporal.sqlite-shm` and `temporal.sqlite-wal`

3. Any docs updates needed?

no, because it's just a bug fix. the docs are already there.

### DIFF
--- a/temporalcli/devserver/server.go
+++ b/temporalcli/devserver/server.go
@@ -321,13 +321,7 @@ func (s *StartOptions) buildSQLConfig() (*config.SQL, error) {
 		conf.DatabaseName = s.DatabaseFile
 	}
 	for k, v := range s.SqlitePragmas {
-		// Only some pragmas allowed
-		switch k {
-		case "journal_mode", "synchronous":
-		default:
-			return nil, fmt.Errorf("unrecognized pragma %q, only 'journal_mode' and 'synchronous' allowed", k)
-		}
-		conf.ConnectAttributes["_"+k] = v
+		conf.ConnectAttributes[k] = v
 	}
 
 	// Apply migrations to sqlite if using file but it does not exist
--- a/temporalcli/commands.server_test.go
+++ b/temporalcli/commands.server_test.go
@@ -3,6 +3,8 @@ package temporalcli_test
 import (
 	"context"
 	"net"
+	"os"
+	"path/filepath"
 	"strconv"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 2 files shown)


temporalcli/devserver/server.go
```diff
@@ -321,13 +321,7 @@ func (s *StartOptions) buildSQLConfig() (*config.SQL, error) {
...
 	}
 	for k, v := range s.SqlitePragmas {
-		// Only some pragmas allowed
-		switch k {
-		case "journal_mode", "synchronous":
-		default:
-			return nil, fmt.Errorf("unrecognized pragma %q, only 'journal_mode' and 'synchronous' allowed", k)
-		}
-		conf.ConnectAttributes["_"+k] = v
+		conf.ConnectAttributes[k] = v
```
cretz, line 324: Interesting, I suspect we just didn't have test cases. Would you be willing to add a test (or adjust an existing one) that just adds any innocuous pragma just to ensure it works?

cretz (commented):
Thanks! Will confer with team just to make sure there wasn't a reason we were limiting the pragmas. I don't see anything obvious at https://github.com/temporalio/temporalite-archived/pull/30 at first glance.

temporalcli/devserver/server.go
vbehar: Sure, I'll add one

cretz (approved):
LGTM, let me confirm with team we have no issues with this, then we'll merge

cretz:
@vbehar - We are curious, for our own tuning reasons, what are your use cases and what different pragmas might you want to set? We are also wanting to tune some defaults, do you think there are opportunities here for us to improve defaults?

vbehar:
> @vbehar - We are curious, for our own tuning reasons, what are your use cases and what different pragmas might you want to set? Any reason we shouldn't continue to limit the pragmas we accept? We are also wanting to tune some defaults, do you think there are opportunities here for us to improve defaults?

yes, our use-case is that we're still running temporal in dev-server mode with sqlite until we properly deploy it - in a more "production-ready" setup. but as it's a shared setup (not on our own laptop), we want to make backups - but we had conflicts between our backup script and temporal.

we quickly found out that enabling the WAL journal is a good start, as well as setting the busy_timeout - cf https://litestream.io/tips/ for example which explains these different settings.
and that's when we discovered that the sqlite-pragma flag doesn't work, and is too restrictive.

so we want to run temporal with the following sqlite pragma:
- journal_mode=WAL
- synchronous=NORMAL
- busy_timeout=5000

cretz:
Thanks!

PR #672, opened 2024-09-16
making the describe task queue command more descriptive

Shivs11:
## What was changed

- addressed comments by adding more description to the `task-queue describe` command

## Why?

- better readability

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

PR #684, opened 2024-10-01
Improving `Describe-TQ` Command Description

Shivs11:
## What was changed

- addressed minor comments which were requested in #672 

## Why?

- opened a new PR since the repo went through some changes with respect to generating code (now uses a .yml file)
- attempted at resolving conflicts on the previous open PR but was faced with unknown issues 
- thought it was best, in the interest of time and given the relatively small size of the PR, to open a new branch and close the issue.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalcli/commandsgen/commands.yml
```diff
@@ -1512,8 +1512,10 @@ commands:
       NOTE: The `TasksAddRate` and `TasksDispatchRate` metrics may differ from the
       actual rate of add/dispatch, because tasks may be dispatched eagerly to an
       available worker, or may apply only to specific workers (they are "sticky").
-      Such tasks are not counted by these metrics.
-
+      Such tasks are not counted by these metrics. Despite the inaccuracy of 
+      these two metrics, the derived metric of `BacklogIncreaseRate` is accurate
+      for backlogs older than a few seconds.
```
josh-berry, line 1517: This seems to contradict the statement above re. the Backlog Increase Rate:

```
This is equivalent to:
`TasksAddRate` - `TasksDispatchRate`.
```

(Also, should it be `BacklogIncreaseRate`, not `Backlog Increase Rate`?)

Shivs11: I think what that statement means to say is that: Despite the inaccuracies put forward by the two metrics (wrt not counting eagerly dispatched tasks, the `BacklogIncreaseRate` metric is an accurate estimate for the number of tasks that are actually present in the backlog.

josh-berry: This may be pedantic but let's change that text to then read: "This is roughly equivalent to"—if we say "This is equivalent to", someone will probably expect to be able to do math and get the same result, and it sounds like that's not the case.

Also, still same question about the spaces in `Backlog Increase Rate`—can you also change that while you're here if it's supposed to be `BacklogIncreaseRate`?

Shivs11: Sorry, I forgot to address that in my response but had made the change locally. Yes, I shall be changing Backlog Increase Rate to the required output.

Made the changes required.

PR #685, opened 2024-10-02
Auto generate CLI docs for documentation repo

yuandrew:
## What was changed
Auto generate all of the *.mdx files present in the current documentation repo. index.mdx is not auto-generated

IMPORTANT: The generated files are not yet being consumed anywhere. A future PR will auto-publish a PR into the docs repo whenever there is a docs related change.

Added a field in the YML to indicate doc-specific information, like SEO related keywords and descriptions.

Removed `cmd-options.cdx` and print each command option description inline in each file instead.
  https://github.com/temporalio/documentation/pull/3122 removes all references to this file from the documentation side.

## Why?
This is in a larger effort to have the CLI be the source of truth for docs, and have the documentation auto-ingest any CLI changes, keeping all information in sync.

This is the first step to this goal, there are a number of follow up tasks that I wanted to split up from this PR:
1. Diff the current information on the docs with the CLI, and unify into a single master-description for each command.
2. Auto-publish PR to docs repo whenever there is a CLI doc change
  1. Auto update sidebars.js and vercel.js whenever a new temporal command is added.
3. Clean up the SEO keywords/tags - being handled by the documentation team
4. Support markdown in the CLI help output in terminal #687

## Checklist
7. Any docs updates needed?

For now, docs will not be updated. There will be a future PR that auto-creates a PR into the documentation repo with any updates.


temporalcli/commandsgen/parse.go
```diff
@@ -44,6 +44,14 @@ type (
...
 		Options                []Option `yaml:"options"`
 		OptionSets             []string `yaml:"option-sets"`
+		Docs                   Docs     `yaml:"docs"`
+	}
+
+	// Docs represents docs-only information that is not used in CLI generation.
+	Docs struct {
+		// Tags are the same as Keywords, but with `-` instead of ` `.
+		Keywords    []string `yaml:"keywords-tags"`
+		Description string   `yaml:"description"`
```
cretz, line 54:
```suggestion
		DescriptionHeader string   `yaml:"description-header"`
```

I wonder if this makes sense just to differentiate from command description and clarify this is a single-line thing

temporalcli/commandsgen/commands.yml
```diff
@@ -213,6 +213,20 @@ commands:
       ```
     option-sets:
       - client
+    docs:
+      description: |
```
cretz, line 217:
```suggestion
      description: >-
```

This will do the proper thing of making it a single line

temporalcli/commandsgen/parse.go
```diff
@@ -44,6 +44,14 @@ type (
 		IgnoreMissingEnv       bool     `yaml:"ignores-missing-env"`
 		Options                []Option `yaml:"options"`
 		OptionSets             []string `yaml:"option-sets"`
+		Docs                   Docs     `yaml:"docs"`
+	}
+
+	// Docs represents docs-only information that is not used in CLI generation.
+	Docs struct {
+		// Tags are the same as Keywords, but with `-` instead of ` `.
+		Keywords    []string `yaml:"keywords-tags"`
```
cretz, line 53:
```suggestion
		KeywordsTags    []string `yaml:"keywords-tags"`
```

Ideally keep the field names the same as the YAML names (can probably just call these "keywords", that you happen to also use them as tags is a generator detail probably)

temporalcli/commandsgen/docs.go
```diff
@@ -0,0 +1,263 @@
...
+	for _, set := range c.OptionSets {
+		// map into optionSet map
+		optionSetOptions := w.optionSetMap[set].Options
+		options = append(options, optionSetOptions...)
+	}
+
+	w.optionsStack = append(w.optionsStack, options)
+}
+
+// TODO: Remove this page and throw inline with each command
```
cretz, line 132: We need to do this in this PR IMO. There are options duplicates, so now `--name` is defined as "Cluster/Service name." which makes no sense for `workflow signal --name`. Ideally we can just inline the options for each command in their page.

temporalcli/internal/cmd/gen-docs/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Generate docs
+	b, err := commandsgen.GenerateDocsFiles(cmds)
+	if err != nil {
+		return err
+	}
+
+	// Write
+	for filename, content := range b {
+		filePath := filepath.Join(commandsDir, filename+".mdx")
```
cretz, line 38: Is there anything `.mdx` specific about these files, or can they just be `.md` for use by everyone viewing this repo and not just our documentation engine?

Alternatively, we should consider having the docs repo generate this and put the files in their repo instead of having docs files outside of docs repo (I wouldn't expect any other SDK project to have the docs.temporal.io docs inside their repo).

Alternatively, like other SDKs, we could consider a place for a more "API docs" type home. In the same way we have typescript.temporal.io, python.temporal.io, dotnet.temporal.io, we could have cli.temporal.io. We can have CI push to this automatically like we do in those other repos.

temporalcli/internal/cmd/gen-docs/main.go
cretz, line 1: Can we make sure we add to CONTRIBUTING.md on how to run this and add this step to the "Regen code, confirm unchanged" step in CI just before the `git diff` part to confirm that any PRs have the latest generated content?

temporalcli/internal/cmd/gen-docs/main.go
yuandrew: I don't think so, we can name it `.md` here, and in the future PR where I automate pushing out a PR to the docs repo, I'll add logic to rename it to the `.mdx` that the doc engine expects. 

Good idea!

yuandrew: Actually, GH renders .mdx in the same way it renders .md, so anyone viewing this repo will see the .mdx files as if they were .md files. Is there any other reason we'd want to use .md here?

fairlydurable: I'm checking with team to see if there's any issues delivery md and not mdx for these very flat files

fairlydurable: There are no known issues but no compelling reason to pull one section out of alignment with the rest of the docs site. Recommend keep as mdx.

cretz: > Is there any other reason we'd want to use .md here?

GH is not the only renderer/viewer of these files. I think unless it uses mdx-specific features, we might as well make it just md. Can rename for docs engine specifically of course.

Or maybe even better, don't commit the generated files in this repo if they are only used for the docs repo. Rather, have the generator here and have the docs engine/generator/CI/whatever run the generator and point to a dir they want to put the mdx files instead of assuming they should be in a `docs/` folder here.

yuandrew: good idea, that will make contributions in this space easier, allowing the CI to handle everything and generate a docs PR without requiring the user to run an additional command to build the docs in this repo

CONTRIBUTING.md
```diff
@@ -28,7 +28,15 @@ First, update [commands.yml](temporalcli/commandsgen/commands.yml) following the
 This will expect every non-parent command to have a `run` method, so for new commands developers will have to implement
 `run` on the new command in a separate file before it will compile.
 
+Once a command is updated, the CI will automatically generate new docs
+and create a PR in the Documentation repo with the corresponding updates. To generate these docs locally, run:
+
+    go run ./temporalcli/internal/cmd/gen-docs
+
+This will auto-generate a new set of docs to `temporalcli/docs/`. If a new root command is added, a new file will be automatically generated, like `temporal activity` and `activity.mdx`.
```
cretz, line 36: You could also accept the dir to write to as a CLI arg to this tool, but up to you

cretz (approved):

PR #686, opened 2024-10-02
[Do not merge] Adds proof-of-concept for man page

fairlydurable:
- Requires help2man
- Available from homebrew
- See https://github.com/Distrotech/help2man

Follow-up to 1 October CLI meeting


cretz:
> Available from homebrew

This needs to be available on all developer platforms IMO. Is there a better tool than this really old one? Would you consider generating markdown and then using pandoc? This will give you more control too I suspect.

fairlydurable:
Josh said he wasn't actually serious about adding the man page material, and the source for the utility is linked if you ever decide that's the direction you want to go in.

This was a fun tiny way to take a break, but I'm going to close this as a non-starter. 

I do appreciate you taking a peek though. :)

PR #691, opened 2024-10-15
Improved docs gen

prasek:
## What was changed
- Improved `Command` model for simpler docs rendering
- Rendered inline options under each command, without cluttering up the right side nav.
- Acorn rendering fix for JSON examples in `.mdx` 
- Add rendering of experimental options.

## Why?
Nexus docs need to be created and want to use docs gen.

## Checklist
1. How was this tested:
- `go run ./temporalcli/internal/cmd/gen-docs `
- verified via http://localhost:3000/

2. Any docs updates needed?
- will feed into overall docs cleanup


yuandrew (changes requested):
Left some small nit comments, but I'm not sure the split between this PR and #691 is what we want. I think the split we want is:
1. The `cmd-options.mdx` improvements and the enrichments that are required for that.
2. All other enrichments and improvements to the current docs generator.

That way we can merge 2. in first, and 1. can be left open as a tool for when someone cleans up the CLI documentation.

Right now it feels like a lot is getting added in, and it's not clear to me what's actually beneficial/needed.

temporalcli/internal/cmd/gen-docs/main.go
```diff
@@ -32,6 +32,11 @@ func run() error {
 		return fmt.Errorf("failed parsing markdown: %w", err)
 	}
 
+	cmds, err = commandsgen.EnrichCommands(cmds)
```
cretz, line 35: Can the logic for this just become part of docs gen or parsing or whatever? Why a separate call?

prasek: It could for sure.

Considerations, IMO:
- `parse.go` is more about reading from`commands.yml` content and doing field validation
- `enrich.go` is about adding fields that aren't read from `commands.yml` to make it easier for `docs.go`
- also this https://github.com/temporalio/tcld/pull/398
  - port docs gen to `tcld`
  - skipped `ParseCommands()` since [we do `ConvertCommands()` -> `EnrichCommands()`](https://github.com/temporalio/tcld/pull/398/files#diff-de51fa3d61c20d869a43ce7264592eb7bef72af035915ba01b80277f00f1e043R31)
  - we do [emit a tcld.yml now](https://github.com/temporalio/tcld/pull/398/files#diff-4e3ba5b9cad41fc0a5def1b128823a7ed408f7fd3a0e392e25cbf2e190d0715c)
   - maybe do want to add front matter to `tcld.yml` and round trip in a way compatible with `ConvertCommands` so maybe we will need a `ParseCommands` over there, for example
      - `ParseCommands()` -> `ConvertCommands()` -> save `tcld.yml`-> `EnrichCommands()`
        - convert would overlay from cli.Command content onto `Commands`
        - round trip writes back to `tcld.yml` in a way the doesn't wipe out docs front matter
        - don't want enriched fields being written to `tcld.yml`
  - so wanted something more consistent across both, in the event we could extract a common package.

temporalcli/commandsgen/enrich.go
```diff
@@ -0,0 +1,252 @@
...
+// temporalcli/commandsgen/commands.yml and generate code from it.
+package commandsgen
+
+import (
+	_ "embed"
+	"sort"
+	"strings"
+)
+
+func EnrichCommands(m Commands) (Commands, error) {
```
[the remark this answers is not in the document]
prasek: added

temporalcli/internal/cmd/gen-docs/main.go
cretz: > enrich.go is about adding fields that aren't read from commands.yml to make it easier for docs.go

If code is only used for docs.go then just put it as part of docs.go. However, at parse time, it's totally fine to populate things that are not necessarily used by everyone. This is a codegen step, feel free to add extra things to parse.

> port docs gen to tcld

`tcld` is beta software that may (will likely) be completely replaced to fit in this CLI model, so I wouldn't spend too much effort on it nor would I affect anything in this repo with anything done over there

prasek: > However, at parse time, it's totally fine to populate things that are not necessarily used by everyone. Feel free to add extra things to parse and/or extra fields to the command struct that only docs side will use.

Sounds good. Moved the enrichments to `parse.go`.

prasek:
> Left some small nit comments, but I'm not sure the split between this PR and #692 is what we want. I think the split we want is:
> 
> 1. The `cmd-options.mdx` improvements and the enrichments that are required for that.
> 2. All other enrichments and improvements to the current docs generator.
> 
> That way we can merge 2. in first, and 1. can be left open as a tool for when someone cleans up the CLI documentation.
> 
> Right now it feels like a lot is getting added in, and it's not clear to me what's actually beneficial/needed.

Makes sense. If we move away from the `cmd-options.mdx` and render command options inline under each command, think it will scale better and be more clear -- and then we don't need the `cmd-options.mdx` improvements. `tcld` docs [does nested options](https://github.com/temporalio/tcld/pull/398/files#diff-119cb3ae9908bc291809b31d6ea1a292670f3c864de416a6ff822102eb847e9aR79) and it's much cleaner IMO.

prasek:
Addressed all feedback and think we're good to go.

temporalcli/commandsgen/parse.go
```diff
@@ -19,15 +20,16 @@ var CommandsYAML []byte
...
+		Name         string   `yaml:"name"`
+		Type         string   `yaml:"type"`
+		Description  string   `yaml:"description"`
+		Short        string   `yaml:"short,omitempty"`
+		Default      string   `yaml:"default,omitempty"`
+		Env          string   `yaml:"env,omitempty"`
+		Required     bool     `yaml:"required,omitempty"`
+		Aliases      []string `yaml:"aliases,omitempty"`
+		EnumValues   []string `yaml:"enum-values,omitempty"`
+		Experimental bool     `yaml:"experimental,omitempty"`
```
cretz, line 32: Is this used by any commands today? Want to set it on the YAML for the ones needed and/or wait until needed to add it?

temporalcli/commandsgen/parse.go
```diff
@@ -45,6 +47,16 @@ type (
 		Options                []Option `yaml:"options"`
 		OptionSets             []string `yaml:"option-sets"`
 		Docs                   Docs     `yaml:"docs"`
+		Index                  int
```
cretz, line 50: How many of these fields are required for docs gen? I don't see `Index` or `SubCommandName` or any of that used there. Are all of these fields needed to generate docs from the existing structure?

temporalcli/commandsgen/parse.go
```diff
@@ -45,6 +47,16 @@ type (
...
+		Base                   *Command
+		Parent                 *Command
+		Children               []*Command
```
cretz, line 53: A lot of times these can be determined dynamically at runtime instead of pre-storing references. See how the code generator handles subcommands and such. I think it's more valuable keeping this field set to the bare minimum than precalculating this information (no perf concerns). So `FileName` may still make sense, but many other things can be determined during generation (e.g. subcommands) or passed as state during generation (e.g. parent).

temporalcli/internal/cmd/gen-docs/main.go
cretz: Made a couple of comments. IMO we should only add fields to the parsed field list that are not otherwise obtainable and are needed. Otherwise, obtain or pass around at runtime.

temporalcli/commandsgen/parse.go
prasek: `Index` was needed to [set other derivative fields](https://github.com/temporalio/cli/pull/691/files/68f575f383e943323d11c0c71f6762eec38aebea#diff-cc2ae8272d97aad62d713b302b77891db4f3e1c9cbf504e4aabcc2e49c46badaR247) in `parse.go`.

`SubCommandName` was used for auto-generated keywords to match the existing docs, but if we're not going to auto-generate keywords we can probably remove this one.

temporalcli/commandsgen/parse.go
prasek: For me having these pre-calculated made `docs.go` rendering cleaner and more focused on the rendering vs. calculating fields in `docs.go`.

The single pass `docs.go` rendering we have today , doesn't provide much latitude for 2nd-order sorting of subcommands e.g.
image

That said, we can always leave this out for now, and bring it back in later if needed.

temporalcli/commandsgen/parse.go
cretz: :+1: I see, so we had it in YAML but not in the parsed structure. I think this shows a gap in the strictness of our parser that you can add any YAML key. We should probably fix that, but doesn't need to be done here.

temporalcli/commandsgen/parse.go
cretz: > across multiple passes so passing it around would be cumbersome.

I think it is best to use contextual state when recursing over updating the external state of the object as temporary storage. Or maybe even better, derive it live as needed during docs gen.

temporalcli/commandsgen/parse.go
cretz: > For me having these pre-calculated made docs.go rendering cleaner and more focused on the rendering vs. calculating fields in docs.go.

Made rendering cleaner for docs only at a cost of expanded/unnecessary/easily-derived state on the model objects. The reason I suggested bringing the "enrich" in the main part is because it looked like it was general purpose things that were missing.

Obviously this isn't a big deal, but ideally the docs code like the code-gen code can derive what it needs when it needs. It'd be nicer to keep the model small, but I won't hold up review.

temporalcli/commandsgen/parse.go
prasek: > How many of these fields are required for docs gen?

Unused fields have been removed.

> Or maybe even better, derive it live as needed during docs gen.

What would you propose for [deriving `isLeafCommand()`](https://github.com/temporalio/cli/pull/691/files#diff-204e1216d4ad67aaaa69283810195e4b426bb9ba834c3ec336a8c8f2fdb1553dR84) during the single pass `docs.go` gen? Scanning the full list of commands for children for each subcommand with `slices.IndexFunc` or similar, since performance isn't a concern?

> I think it is best to use contextual state when recursing over updating the external state of the object as temporary storage.

So use a lookup map for the index or scan with `slices.IndexFunc`?

prasek: I see, the `code.go` is just [scanning for subcommands](https://github.com/temporalio/cli/blob/main/temporalcli/commandsgen/code.go#L182) for each command.

If that's the approach we're using, happy to make it consistent.

temporalcli/commandsgen/parse.go
prasek: > though will still wait on removing the fields that are never used

Removed the unused fields.

> ideally the docs-gen code like the code-gen code can derive what it needs when it needs.

just saw that [`code.go` is scanning for subcommands for each command](https://github.com/temporalio/cli/pull/691#discussion_r1823258517). Can make it consistent if that's the approach we're using in this package, esp. since performance isn't an issue.

temporalcli/commandsgen/parse.go
cretz: I am a fan of it, but I won't push too hard on this code since it's build-time tooling.

> Scanning the full list of commands for children for each subcommand with slices.IndexFunc or similar, since performance isn't a concern?

Definitely not a concern for this use case

> So use a lookup map for the index or scan with slices.IndexFunc?

Probably only need `slices.Index` if you have a pointer, but sure.

cretz (approved):
LGTM (any concerns about approach of pre-populating derivable values are non-blocking). May want to have another reviewer like @josh-berry also look before merging.

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +76,38 @@ func (w *docWriter) writeCommand(c *Command) {
...
-	sort.Slice(allOptions, func(i, j int) bool {
-		return allOptions[i].Name < allOptions[j].Name
-	})
+	if isLeafCommand(c) {
+		w.fileMap[fileName].WriteString("Use the following options to change the behavior of this command.\n\n")
 
-	for _, option := range allOptions {
-		w.fileMap[fileName].WriteString(fmt.Sprintf("## %s\n\n", option.Name))
-		w.fileMap[fileName].WriteString(option.Description + "\n\n")
+		// gather options from command and all options aviailable from parent commands
```
josh-berry, line 87:
```suggestion
		// gather options from command and all options available from parent commands
```

Also, an optional bit of feedback that can be improved upon later if you don't have time now:

I think I would try to keep the grouping similar to what we have in the actual CLI help output—global options like `--env` probably should be separate from command-specific options so there's less for users to sort through.

(I would ideally like to break this down even further along the lines of option sets, even in the `--help` output itself, but that's for another time.)

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +76,38 @@ func (w *docWriter) writeCommand(c *Command) {
...
+			allOptions = append(allOptions, options...)
+		}
 
+		// alphabetize options
+		sort.Slice(allOptions, func(i, j int) bool {
+			return allOptions[i].Name < allOptions[j].Name
+		})
+
+		for _, option := range allOptions {
+			w.fileMap[c.FileName].WriteString(fmt.Sprintf("**--%s**\n\n", option.Name))
```
josh-berry, line 99: For flags that take an argument (as opposed to just being a boolean), we should probably show that here like we do in the `--help` output:

```
  -n, --namespace string    Temporal Service Namespace. (default "default")
```

So printing something like `**--namespace** _string_` would be nice.

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +76,38 @@ func (w *docWriter) writeCommand(c *Command) {
...
+			w.fileMap[c.FileName].WriteString(encodeJSONExample(option.Description) + "\n\n")
+			if len(option.Short) > 0 {
+				w.fileMap[c.FileName].WriteString("Alias: `" + option.Short + "`\n\n")
```
josh-berry, line 102: Similarly, if we have an alias, I think it would be a little more compact and readable to print it on the same line as the full-name flag, rather than in its own paragraph. (e.g. my `--namespace` example above)

So the full output, including my previous comment, would be something like:

```md
**--namespace** _string_, **-n** _string_
```

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +76,38 @@ func (w *docWriter) writeCommand(c *Command) {
...
+			}
```
josh-berry, line 103: Can we also print the default option value, if we have one? We probably also need to print the allowable choices for enum-types (e.g. string-enums).

josh-berry (commented):
I just looked at the output, not the code, and left comments based on what I saw in the generated docs. Thanks again @prasek for working on this!

Also, apologies for the delay in getting to this review. It's been in my list since you first opened the PR, and I usually try to get to them faster than this but I've been a bit scatterbrained.

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +76,38 @@ func (w *docWriter) writeCommand(c *Command) {
 }
 
 func (w *docWriter) writeSubcommand(c *Command) {
-	fileName := strings.Split(c.FullName, " ")[1]
-	subCommand := strings.Join(strings.Split(c.FullName, " ")[2:], "")
-	w.fileMap[fileName].WriteString("## " + subCommand + "\n\n")
+	fileName := c.FileName
+	prefix := strings.Repeat("#", c.Depth)
+	w.fileMap[fileName].WriteString(prefix + " " + c.LeafName + "\n\n")
```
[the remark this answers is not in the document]
prasek: The docs right side nav is rendered using this heading as well and the full name doesn't fit, so using the shorter name matches what the docs currently do.

temporalcli/commandsgen/docs.go
prasek: Agree, would do that as a follow on.

temporalcli/commandsgen/docs.go
prasek: Agree, that would be a nice addition.

temporalcli/commandsgen/docs.go
prasek: Added.

temporalcli/commandsgen/docs.go
prasek: Added.

prasek:
Thanks @josh-berry, incorporated all your suggestions except using the full command name for the heading, [see above](https://github.com/temporalio/cli/pull/691#discussion_r1823635323).

temporalcli/commandsgen/parse.go
prasek: > I am a fan of it, but I won't push too hard on this code since it's build-time tooling.

No worries, updated to use derived state to keep the model small.

prasek:
Addressed feedback on options rendering and using derived state to keep the model small. Think this is good to go.

temporalcli/commandsgen/parse.go
prasek: Moved to derived state.

temporalcli/internal/cmd/gen-docs/main.go
prasek: Fixed.

temporalcli/commandsgen/docs.go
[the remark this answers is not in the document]
josh-berry: Gotcha, that makes sense. There are examples for each command, so this is probably okay even if not ideal. Marking comment resolved.

temporalcli/commandsgen/docs.go
```diff
@@ -75,27 +79,75 @@ func (w *docWriter) writeCommand(c *Command) {
...
 
-	for _, option := range allOptions {
-		w.fileMap[fileName].WriteString(fmt.Sprintf("## %s\n\n", option.Name))
-		w.fileMap[fileName].WriteString(option.Description + "\n\n")
+	w.fileMap[fileName].WriteString(fmt.Sprintf("**%s:**\n\n", prefix))
 
+	for _, o := range options {
+		// option name and alias
+		w.fileMap[fileName].WriteString(fmt.Sprintf("**--%s** _%s_", o.Name, o.Type))
+		if len(o.Short) > 0 {
```
yuandrew, line 128: Do we need to print the type for each form of the option?
i.e. for `**--output** _string-enum_, **-o** _string-enum_`

`**--output**, **-o** _string-enum_` would be more succinct. The type should always be the same, so I don't see why we need to print it twice

yuandrew (approved):
sorry, I've been out sick the last two days, changes look good! One minor comment, but looks great otherwise! Thanks for the improvement here Phil :)

temporalcli/commandsgen/docs.go
prasek: Agree will update.

PR #692, opened 2024-10-15
Enhanced docs gen to more closely match existing docs (legacy)

prasek:
> [!NOTE]  
> Not for merging, only for reference for generating legacy docs during the transition.

## What was changed
- Use enrichments from #691
- Emit `cmd-options.mdx` with collated on option description
- Surface different option descriptions and command usages to make it easier to combine into existing `cmd-options.mdx`

See this branch for [the combined approach](https://github.com/prasek/temporal-cli/tree/cli-docs-gen-all) with follow-on PRs merged in.

Note: longer term would recommend moving away from `cmd-options.mdx` and to an inline options model [similar to how `tcld` docs works](https://docs.temporal.io/cloud/tcld/namespace/), as some option names like `name`, `build-id`, and `yes` have lots of overlapping definitions and some are specific to command usage.

## Why?
Needed updated `operator.mdx` and `cmd-options.mdx` output for Nexus docs and wanted to use docs gen.

## Checklist
1. How was this tested:
- `go run ./temporalcli/internal/cmd/gen-docs `
- also tested with [the combined appraoch](https://github.com/prasek/temporal-cli/tree/cli-docs-gen-all)
   - copied generated docs (or subset) to temporalio/documentation
   - `yarn start`
   - verified via http://localhost:3000/

2. Any docs updates needed?
- follow on PR will generate the Nexus docs


temporalcli/commandsgen/docs.go
```diff
@@ -75,43 +97,119 @@ func (w *docWriter) writeCommand(c *Command) {
...
+	w.fileMap[fileName].WriteString(prefix + " " + c.LeafName + "\n\n")
 	w.fileMap[fileName].WriteString(c.Description + "\n\n")
-	w.fileMap[fileName].WriteString("Use the following options to change the behavior of this command.\n\n")
 
-	// gather options from command and all options aviailable from parent commands
-	var allOptions = make([]Option, 0)
-	for _, options := range w.optionsStack {
-		allOptions = append(allOptions, options...)
-	}
+	if len(c.Children) == 0 {
```
yuandrew, line 105: Great improvement here! The old code fails to handle a few edge cases with more complex subcommands.

How are `Parent` and `Children` used? I only see `Children` being used to check `if len(c.Children) == 0 {`, and I don't see the `Parent` information used anywhere. If that's the case, I feel like a simpler approach would be to keep a `IsLeafCommand` from `enrich.go` and use that for this condition here

temporalcli/commandsgen/enrich.go
cretz, line 1: I don't think this separate concept needs to exist. Just put inside parse or put inside docs.

temporalcli/commandsgen/docs.go
prasek: @yuandrew Thanks! `Parent` is used in `EnrichCommand()` for various things including setting `Children` on the `Parent`.

`Children` is used for the more [hierarchical docs gen style used in `tcld`](https://github.com/temporalio/tcld/pull/398/files#diff-119cb3ae9908bc291809b31d6ea1a292670f3c864de416a6ff822102eb847e9aR84) which we probably want to move to for `temporal` at some point.

temporalcli/commandsgen/enrich.go
prasek: sounds good, combined

prasek:
Incorporated feedback into #691. Think we can close this PR as legacy docs / will not merge - since we're moving away from `cmd-options.mdx` to inline option rendering under each command.

This PR may be useful for generating legacy docs during the transition.

josh-berry:
Closing per @prasek's last comment.

PR #694, opened 2024-10-17
update --command-timeout description

prasek:
## What was changed
update `--command-timeout` description per docs feedback

## Why?
make it more clear

## Checklist
1. How was this tested:
locally with `go run ./cmd/temporal operator nexus endpoint create --help`

2. Any docs updates needed?
yes, see related:
- https://github.com/temporalio/documentation/pull/3162

### DIFF
--- a/temporalcli/commandsgen/commands.yml
+++ b/temporalcli/commandsgen/commands.yml
@@ -197,7 +197,8 @@ commands:
         description: Raw payload output, even if the JSON option was used.
       - name: command-timeout
         type: duration
-        description: Timeout for the span of a command.
+        description: |
+          The command execution timeout in seconds. 0 means no timeout.
 
   - name: temporal activity
     summary: Complete or fail an Activity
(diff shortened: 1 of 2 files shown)


josh-berry (approved):

temporalcli/commandsgen/commands.yml
```diff
@@ -197,7 +197,8 @@ commands:
         description: Raw payload output, even if the JSON option was used.
       - name: command-timeout
         type: duration
-        description: Timeout for the span of a command.
+        description: |
+          The command execution timeout in seconds. 0 means no timeout.
```
cretz, line 201: This is not "in seconds", you have to provide the unit. You can provide `10ms` for instance. I don't think `0` is an acceptable value unless I'm misreading the code.

josh-berry: Amusingly enough, `0` does work, but `5` doesn't. Probably because it doesn't matter what the units are if you specify `0`; it's still no time. But yes, you're right, probably we should show the units and remove "in seconds".

cretz: Yup, 0 works: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/time/format.go;l=1631-1634

prasek: fixed

josh-berry (approved):

cretz (approved):

PR #704, opened 2024-11-13
Standardize CLI StringEnum format

### LINKED ISSUE #678: [Feature Request] Standardize Command Option string-enum value format

Some use camel case (`type` has values like `FirstWorkflowTask`), others use snake case (`reject-condition` has `not_open`), some capitalize the first letter (`reapply-type` has `All`), others are all lowercase (`wait-for-stage` has `accepted`). This is likely a product of history.

We should land on a standardized format for this. But we also need to make sure we support aliasing the old options for backwards compat

yuandrew:
## What was changed

Standardized CLI StringEnum arguments to follow `kebob-case`

Also addressed https://github.com/temporalio/cli/pull/677#discussion_r1773519018 regarding "pretty" as a legacy alias for "text"

## Why?

Should standardize our format across the CLI, and `kebob-case` seems to be somewhat common, like AWS CLI and others

## Checklist

1. Closes #678

2. How was this tested:
Added tests

3. Any docs updates needed?

Probably, still a TODO I need to complete


temporalcli/commandsgen/commands.yml
```diff
@@ -1071,12 +1074,14 @@ commands:
...
-          - Double
-          - Bool
-          - Datetime
+          - text
+          - keyword
+          - int
+          - double
+          - bool
+          - datetime
+          - keyword-list
```
cretz, line 1083: These values represent actual, documented constants at https://docs.temporal.io/visibility#supported-types. Not sure the CLI's desire for enum consistency should override Temporal's desire for consistency for all uses of these constants in all environments.

yuandrew: gotcha, yeah while going through this exercise I was wondering if there were certain constants that were beyond the CLI and take precedence over this idea for consistency

cretz: I do not have a strong opinion, it is just worth noting for this enumerate specifically we document certain string terms globally

PR #707, opened 2024-11-18
Upgrade the macos runner in ci workflow

rodrigozhou:
## What was changed

Upgrade the `macos` runner in the CI workflow.

## Why?

`macos-12` is in the process of deprecation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, macos-12, macos-14, windows-latest, ubuntu-arm]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, macos-12, macos-14, windows-latest, ubuntu-arm]
+        os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-arm]
```
josh-berry, line 13: Thanks! Let's keep `macos-14` and just swap `macos-12` for `macos-latest` (which should be 15 nowadays).

josh-berry (approved):

PR #708, opened 2024-11-18
Upgrade the macos runner in ci workflow (#707)

rodrigozhou:
## What was changed

Upgrade the `macos` runner in the CI workflow.

## Why?
 `macos-12` is in the process of deprecation

## Checklist

1. Closes 

2. How was this tested: 

3. Any docs updates needed?


josh-berry:
No need, I just fast-forwarded `next-server` to be the tip of `main`. Thanks though!

PR #725, opened 2024-12-09
ci: Directly check for needed secrets before testing against cloud

josh-berry:
There are other situations besides running in a fork in which the necessary secrets are not available.  For example, PRs opened by dependabot do not seem to have access to the secrets.

We now directly check if secrets are present, rather than inferring their presence from other environmental things.


cretz (approved):

josh-berry:
I did confirm that cloud tests are still run. Can't really tell if the inverse is true just in this PR itself, but I'm confident. I'll confirm on #703 after I merge this PR.

PR #726, opened 2024-12-09
disable tq info caching in task queue test

carlydf:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

go.mod
```diff
@@ -11,25 +11,25 @@ require (
...
-	go.temporal.io/api v1.38.0
-	go.temporal.io/sdk v1.29.1
-	go.temporal.io/server v1.25.2
-	google.golang.org/grpc v1.65.0
-	google.golang.org/protobuf v1.34.2
+	go.temporal.io/api v1.43.0
+	go.temporal.io/sdk v1.30.1
+	go.temporal.io/server v1.26.2-125.1
+	google.golang.org/grpc v1.67.1
+	google.golang.org/protobuf v1.35.1
```
josh-berry, line 24: Do we need all these updates for this PR?

josh-berry (commented):
@carlydf Do you plan to merge this soon? Would like to take it in the next release if it's ready.

go.mod
carlydf: Nope, just for testing with the SDK that Antonio was using when he ran into this issue last month. I removed the uneccessary updates and will merge!

temporalcli/commands_test.go
```diff
@@ -216,6 +216,9 @@ func (s *SharedServerSuite) SetupSuite() {
 				// Allow a high rate of change to namespaces, particularly
 				// for the task-queue command tests.
 				"frontend.namespaceRPS.visibility": 10000,
+				// this is overridden since we don't want caching to be enabled while testing DescribeTaskQueue
+				// behaviour related to versioning
+				"matching.TaskQueueInfoByBuildIdTTL": 0 * time.Second,
```
cretz, line 221: So we disable search attribute caching for all users of the dev server, not just our test suite. I wonder if we should do that here too. Though it may give the dev server users unreasonable expectations about eventual consistency for task queue info.

carlydf: I think it's reasonable for the dev server to have the caching for now -- can always change it later, but I think this caching mechanism may change in the future as well

PR #729, opened 2024-12-13
CLI comamnd for UpdateActivityOptions API

ychebotarev:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalcli/commands.activity.go
```diff
@@ -70,3 +92,109 @@ func (c *TemporalActivityFailCommand) run(cctx *CommandContext, args []string) e
...
+	_ = cctx.Printer.PrintStructured(updatedOptions, printer.StructuredOptions{})
+
+	return nil
+}
+
+// Converts the duration to Go's native time.Duration.
+// Returns the zero time.Duration value for nil duration.
+func toDuration(duration *durationpb.Duration) (d time.Duration) {
+	if duration != nil {
+		d = duration.AsDuration()
```
cretz, line 197: I think this call is already `nil` safe, not sure a new helper is needed

temporalcli/commands.activity_test.go
```diff
@@ -106,6 +106,81 @@ func (s *SharedServerSuite) TestActivity_Fail_InvalidDetail() {
...
+	// updated
+	s.ContainsOnSameLine(out, "ScheduleToCloseTimeout", "41s")
+	s.ContainsOnSameLine(out, "ScheduleToStartTimeout", "11s")
+	s.ContainsOnSameLine(out, "StartToCloseTimeout", "10s")
+	s.ContainsOnSameLine(out, "InitialInterval", "4s")
+	s.ContainsOnSameLine(out, "MaximumAttempts", "10")
+
+	// old value
+	// note - this is a snapshot of current values
+	//if this test fails, check the default values of activity options
```
cretz, line 177:
```suggestion
	// if this test fails, check the default values of activity options
```

Usually go fmt catches these IIRC

cretz (approved):
LGTM, nothing blocking, though would like @josh-berry to look or get another opinion at least

temporalcli/commands.activity.go
ychebotarev: yeah, looks both GetSeconds() and GetNanos() are nil sage.
Thus making  AsDuration nil safe as well.
I will remove it

ychebotarev:
I want to move forward, I waited a day, if there is something I will fix in the follow up PR.

PR #730, opened 2024-12-16
CLI for pause/unpause activity API

ychebotarev:
## What was changed

Add pause/unpause activity API.

## Why?

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalcli/commandsgen/commands.yml
```diff
@@ -374,6 +375,99 @@ commands:
...
+    description: |
+      Pausing an activity means:
+        - If the activity is currently waiting for a retry or is running 
+      and subsequently fails, it will not be rescheduled until it is unpaused.
+      - If the activity is paused, calling this method will have no effect.
+      - If the activity is running and finishes successfully, 
+        the activity will be completed.
+      - If the activity is running and finishes with failure:
+        * if there is no retry left - the activity will be completed.
+        * if there are more retries left - the activity will be paused.
```
cretz, line 389: This formatting is a bit strange. Some `-` are indented, some aren't, some parts are "If ..., ..." and some are "if ... - ...", etc.

temporalcli/commandsgen/commands.yml
```diff
@@ -374,6 +375,99 @@ commands:
...
+        description: Also reset the activity.
+      - name: no-wait
+        type: bool
+        description: |
+          If reset flag is provided, the activity will be scheduled immediately.
+          If reset flag is not provided, the activity will be scheduled if 
+          it is waiting for retry.
+      - name: reset-heartbeats
+        type: bool
+        description: I reset flag is provided - reset the activity heartbeats.
```
cretz, line 466:
```suggestion
        description: If reset flag is provided - reset the activity heartbeats.
```

Also, after merge, may want to see if docs team wants to do a pass on these docs (they have cleaned up other CLI docs for us)

temporalcli/commands.activity.go
```diff
@@ -189,3 +189,62 @@ func (c *TemporalActivityUpdateOptionsCommand) run(cctx *CommandContext, args []
...
+		return err
+	}
+	defer cl.Close()
+
+	_, err = cl.WorkflowService().PauseActivityById(cctx, &workflowservice.PauseActivityByIdRequest{
+		Namespace:  c.Parent.Namespace,
+		WorkflowId: c.WorkflowId,
+		RunId:      c.RunId,
+		ActivityId: c.ActivityId,
+
```
cretz, line 205:
```suggestion
```

Blank line here separating the identity field but not on the unpause equivalent

temporalcli/commands.activity.go
```diff
@@ -189,3 +189,62 @@ func (c *TemporalActivityUpdateOptionsCommand) run(cctx *CommandContext, args []
...
 }
+
+func (c *TemporalActivityPauseCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	_, err = cl.WorkflowService().PauseActivityById(cctx, &workflowservice.PauseActivityByIdRequest{
```
cretz, line 200: Since you are using direct gRPC instead of a high-level Go SDK API, you should set the request ID on calls yourself (same for update options from previous merge)

temporalcli/commands.activity.go
```diff
@@ -189,3 +189,62 @@ func (c *TemporalActivityUpdateOptionsCommand) run(cctx *CommandContext, args []
...
+	defer cl.Close()
+
+	request := &workflowservice.UnpauseActivityByIdRequest{
+		Namespace:  c.Parent.Namespace,
+		WorkflowId: c.WorkflowId,
+		RunId:      c.RunId,
+		ActivityId: c.ActivityId,
+		Identity:   c.Identity,
+	}
+	if c.Reset {
```
cretz, line 229: For me to understand not related to CLI specifically, what would happen server-side if I didn't set either of the `oneof`s for operation? I couldn't tell from proto docs.

temporalcli/commands.activity_test.go
```diff
@@ -142,43 +142,34 @@ func (s *SharedServerSuite) TestActivityOptionsUpdate_Accept() {
 	s.ContainsOnSameLine(out, "MaximumAttempts", "5")
 }
 
-func (s *SharedServerSuite) TestActivityOptionsUpdate_Partial() {
```
cretz, line 145: Did the partial update options test get removed?

temporalcli/commands.activity.go
```diff
@@ -189,3 +189,62 @@ func (c *TemporalActivityUpdateOptionsCommand) run(cctx *CommandContext, args []
...
+		request.Operation = &workflowservice.UnpauseActivityByIdRequest_Resume{
+			Resume: &workflowservice.UnpauseActivityByIdRequest_ResumeOperation{
+				NoWait: c.NoWait,
+			},
+		}
+	}
+
+	_, err = cl.WorkflowService().UnpauseActivityById(cctx, request)
+	if err != nil {
+		return fmt.Errorf("unable to update Activity options: %w", err)
```
cretz, line 246: Error message improperly says "update Activity options"

cretz (commented):
May also want @josh-berry to look

temporalcli/commandsgen/commands.yml
```diff
@@ -374,6 +375,99 @@ commands:
...
+          the number of attempts.
+        * if 'no_wait' flag is set, the activity will be scheduled immediately.
+        * if 'reset_heartbeats' flag is set, the activity heartbeat timer 
+          and heartbeats will be reset.
+      If the activity is in waiting for retry and past it retry timeout, 
+      it will be scheduled immediately.
+      Once the activity is unpaused, all timeout timers will be regenerated.
+
+      Specify the
+      Activity and Workflow IDs:
```
josh-berry, line 436: Same as above: this is going to be quite hard to read, and we also want to up-level docs to focus more on the effects/outcomes a command has, rather than what it does or how it works.

```suggestion
      Re-schedule a previously-paused Activity for execution.

      If the Activity is not running and is past its retry timeout, it will be
      scheduled immediately. Otherwise, it will be scheduled after its retry
      timeout expires. The Activity can be retried immediately with `--no-wait`.

      Use `--reset` to reset the number of previous run attempts to zero. For
      example, if an Activity is near the maximum number of attempts N specified
      in its retry policy, `--reset` will allow the Activity to be retried
      another N times after unpausing.

      Specify the Activity and Workflow IDs:
```

temporalcli/commandsgen/commands.yml
```diff
@@ -374,6 +375,99 @@ commands:
...
+        description: Identity of the user submitting this request.
+      - name: reset
+        type: bool
+        description: Also reset the activity.
+      - name: no-wait
+        type: bool
+        description: |
+          If reset flag is provided, the activity will be scheduled immediately.
+          If reset flag is not provided, the activity will be scheduled if 
+          it is waiting for retry.
```
josh-berry, line 463:
```suggestion
          Schedule the Activity immediately, even if its retry timeout has not expired.
```

temporalcli/commandsgen/commands.yml
josh-berry:
```suggestion
        description: Reset the Activity's heartbeat timeout. Only works with --reset.
```

temporalcli/commands.activity.go
```diff
@@ -189,3 +189,62 @@ func (c *TemporalActivityUpdateOptionsCommand) run(cctx *CommandContext, args []
...
+				ResetHeartbeat: c.ResetHeartbeats,
+			},
+		}
+	} else {
+		request.Operation = &workflowservice.UnpauseActivityByIdRequest_Resume{
+			Resume: &workflowservice.UnpauseActivityByIdRequest_ResumeOperation{
+				NoWait: c.NoWait,
+			},
+		}
+	}
```
josh-berry, line 242: Check and fail if `--reset-heartbeat` is specified, but not `--reset`.

josh-berry (changes requested):

temporalcli/commandsgen/commands.yml
```diff
@@ -374,6 +375,99 @@ commands:
...
+        type: string
+        description: Activity ID to pause.
+        required: true
+      - name: identity
+        type: string
+        description: Identity of the user submitting this request.
+    option-sets:
+      - workflow reference
+
+  - name: temporal activity unpause
```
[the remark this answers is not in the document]
ychebotarev: I will start a discussion with Drew about this (I also prefer "resume")

temporalcli/commands.activity.go
ychebotarev: Server side:
it neither is provided - activity will be "unpaused', but will not start if it in retry.

temporalcli/commands.activity_test.go
ychebotarev: ah...
I was working on this while Update for on PR, and it looks like the merge did a bad job.
Yeah, I will add that test back, there is no reason to remove it.

temporalcli/commands.activity.go
ychebotarev: fixed

temporalcli/commands.activity.go
cretz: To confirm, that is the equivalent of providing an empty resume in the oneof? I just want to make sure there isn't a state not represented here?

ychebotarev: Not sure what you mean by "empty resume" - the one that doesn't have noWait set?
If so - yes, resume without noWait set and no flags at all will lead to the same result - activity will be unpaused, but will not be forcibly immediately started (it still can if it is past retry time).

temporalcli/commands.activity.go
ychebotarev: add test that will explicitly fail if --reset-heartbeat is specified, but not --reset.

temporalcli/commands.activity.go
ychebotarev: add explicit "requestId" parameter.

That said - to be clear multiple "pause" calls will have the same effect, with same or different request ids.

temporalcli/commands.activity.go
cretz: > Not sure what you mean by "empty resume" - the one that doesn't have noWait set?

Yeah, a proto object with all of its default fields (so yes no_wait unset which defaults to false). So the absent oneof and a no-wait-false resume are the same thing in the API. If you do make any new commits to the API repo, we should add a comment clarifying that the default is resume.

josh-berry (approved):
lgtm from a docs/design perspective, would like @cretz approval as well from a code perspective.

cretz:
(as discussed offline, looked good to me btw)

PR #731, opened 2024-12-18
Expose the underlying slog logger to the dev server

bergundy:
Also upgrade server to `v1.26.2-rc.0`.

The server depends on the Nexus SDK, which expects an `slog.Logger`, the server will do best effort conversion from `common/log.Log` to `slog.Logger` but providing access to the underlying `slog.Logger` gives better control that the default shim provided by the server.

### DIFF
--- a/temporalcli/devserver/log.go
+++ b/temporalcli/devserver/log.go
@@ -14,6 +14,7 @@ type slogLogger struct {
 }
 
 var _ log.Logger = slogLogger{}
+var _ log.SLogWrapper = slogLogger{}
 
 func (s slogLogger) Debug(msg string, tags ...tag.Tag) { s.Log(slog.LevelDebug, msg, tags) }
 func (s slogLogger) Info(msg string, tags ...tag.Tag)  { s.Log(slog.LevelInfo, msg, tags) }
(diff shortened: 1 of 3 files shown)


cretz (approved):

Sushisource (approved):

temporalcli/devserver/log.go
```diff
@@ -25,6 +26,10 @@ func (s slogLogger) DPanic(msg string, tags ...tag.Tag) { s.Log(slog.LevelError,
 func (s slogLogger) Panic(msg string, tags ...tag.Tag)  { s.Log(slog.LevelError, msg, tags) }
 func (s slogLogger) Fatal(msg string, tags ...tag.Tag)  { s.Log(slog.LevelError, msg, tags) }
 
+func (s slogLogger) SLog() *slog.Logger {
+	return s.log
+}
```
josh-berry, line 31: Forgive the probably-dumb Go question, but I don't understand how this is being used? Is there some magic somewhere that's calling this function?

cretz: IIUC, it's being used in the server code. So we pass a `log.Logger` to the server, and server-side they see if it's also a `log.SLogWrapper` they call this to extract the slog.

bergundy: https://github.com/temporalio/temporal/blob/4a71548e03219f52203b059369d66286d30e85f5/common/log/slog.go#L55

PR #732, opened 2024-12-19
Add CLI command for Reset activity API

ychebotarev:
## What was changed

Add CLI support for Reset activity API

## Why?

### DIFF
--- a/temporalcli/commands.activity_test.go
+++ b/temporalcli/commands.activity_test.go
@@ -231,6 +231,37 @@ func (s *SharedServerSuite) TestActivityUnPause_Failed() {
 	s.Error(res.Err)
 }
 
+func (s *SharedServerSuite) TestActivityReset() {
+	run := s.waitActivityStarted()
+	wid := run.GetID()
+	aid := "dev-activity-id"
+	identity := "MyIdentity"
+
+	res := s.Execute(
+		"activity", "reset",
+		"--activity-id", aid,
+		"--workflow-id", wid,
+		"--run-id", run.GetRunID(),
+		"--identity", identity,
+		"--address", s.Address(),
+	)
+
+	s.NoError(res.Err)
+
+	// reset should fail because activity is not found
+
+	res = s.Execute(
+		"activity", "reset",
+		"--activity-id", "fake_id",
+		"--workflow-id", wid,
+		"--run-id", run.GetRunID(),
+		"--identity", identity,
+		"--address", s.Address(),
+	)
+
+	s.Error(res.Err)
+}
+
 // Test helpers
 
 func (s *SharedServerSuite) waitActivityStarted() client.WorkflowRun {
--- a/temporalcli/commands.activity_test.go
+++ b/temporalcli/commands.activity_test.go
@@ -6,6 +6,7 @@ import (
 
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/history/v1"
+	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/sdk/client"
 )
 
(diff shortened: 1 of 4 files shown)


temporalcli/commands.activity_test.go
```diff
@@ -231,6 +231,37 @@ func (s *SharedServerSuite) TestActivityUnPause_Failed() {
...
+	res := s.Execute(
+		"activity", "reset",
+		"--activity-id", aid,
+		"--workflow-id", wid,
+		"--run-id", run.GetRunID(),
+		"--identity", identity,
+		"--address", s.Address(),
+	)
+
+	s.NoError(res.Err)
```
cretz, line 249: Didn't mention it on the last review but I should have. Ideally these tests should confirm the invoked action actually occurs.

josh-berry:
@ychebotarev Do you think you'll be able to get this merged soon? Would like to cut the next CLI release soonish (once a couple other things land) and would love to take this too if I can.

temporalcli/commands.activity_test.go
ychebotarev: We don't return any values for that. No error indicates that the command was processed successfully. I don't want to write more complicated test - it will be a duplicate of existing functional tests.

cretz: How do you know this command did anything? Sometimes we do replicate functional tests in every place we do them. When these tests are written in each SDK (and the CLI is basically an SDK), we make sure the intended effect occurs, yes even if that technically is duplicated in each language.

ychebotarev: I know that this command reach the server and return no errors. Which means it was a valid call, and it was processed. 
How do I know it was correctly processed? Functional tests. 
I don't want to replicate those functional tests here,  don't want to test server functionality in CLI tests.

ychebotarev: > Not all of them, just confirm it did something

>>I know that this command reach the server and return no errors. 
That is enough.

josh-berry: I think Chad's point is that this test isn't actually testing anything. Any implementation of `activity reset` within the CLI, including no implementation at all, would allow this test to pass. We need to check some observable side-effect that occurs as a result of this call.

That doesn't mean we have to replicate all the functional tests the server does. But it does mean checking that the user would see some expected side-effect (any one will do) as a result of running this command.

Hope that clarifies what's expected here.

ychebotarev: Unlike for example for updating activity there is no direct expected side effect - the response is empty. 
If there will be any kind of error on server side - like "Not Found" - it will return an error.

Even If the response was not empty - one may apply the same argument  "what if I just return expected result directly without calling OSS".
So the proper test will be:
* make sure that OSS call happens
* independently make sure that expected effect take place

In this case (just like in my functional test) for the proper testing I need to make activity fail few times, then make DescribeWorkflow call and verify that the number of attempts is greater then 1.
Then block it (otherwise test becomes non-deterministic), then make CLI reset activity call, then make DescribeWorkflow call and verify that the number of attempts is 1. This is literally functional test, just much harder.

cretz: Yes, it'd basically be a functional test that resets after attempt > 1 and confirms it goes back to 1. You don't have to provide full coverage like you do where the implementation is, but at least one end-to-end assertion is worth it. We will surely be writing this same functional test for every SDK we provide this high-level activity reset call on.

ychebotarev: >Yes, it'd basically be a functional test
And that is why  there is no need to do that.

cretz: There is a need and we do these kinds of end-to-end tests in other tests in all of our SDKs (which includes the CLI). Right now if the CLI code did nothing, this assertion would pass. Or if it did something but never actually made the server invocation, you'd never know. Even with the next test testing for failure, you don't even know what failure it might hit.

cretz (approved):

PR #735, opened 2025-01-06
Add versioning-3 support including Deployment API, versioning overrides, and enhanced `workflow describe`

antlai-temporal:
## What was changed

This PR adds support for the new deployment API, extends `workflow describe` to 
provide versioning information, and adds batch and non-batch operators to modify
workflow execution properties, such as, versioning overrides.

```
temporal worker-deployment 
    temporal worker-deployment describe (--report-reachability)
    temporal worker-deployment get-current
    temporal worker-deployment set-current
    temporal worker-deployment list
temporal workflow update-options
temporal workflow describe
 ``` 

## Why?

Support new Worker Versioning approach based on Deployments.

## Checklist

1. Closes 
#720 #719 #718 
2. How was this tested:

System and unit tests added

### CODE REVIEW

File: temporalcli/commandsgen/commands.yml [hunk 1]
```diff
@@ -394,6 +394,160 @@ commands:
...
+      ```
+
+      For example:
+
+      ```
+      temporal deployment get-current \
+          --deployment-series-name YourDeploymentSeriesName
+      ```
+    options:
+      - name: deployment-series-name
```
cretz, line 485:
```suggestion
      - name: series-name
```

Do you need to prefix many of these options with `deployment-`?

temporalcli/commandsgen/commands.yml
```diff
@@ -2498,6 +2653,95 @@ commands:
         type: int
         description: Maximum number of Workflow Executions to display.
 
+  - name: temporal workflow modify-options
```
cretz, line 2656:
```suggestion
  - name: temporal workflow update-options
```

Like https://github.com/temporalio/cli/pull/729

temporalcli/commands.deployment.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/api/common/v1"
+
+)
+
+type taskQueuesInfosRowType struct {
+		Name string `json:"name"`
+		Type string `json:"type"`
+		FirstPollerTime time.Time `json:"firstPollerTime"`
```
cretz, line 17: Is this file `go fmt`'d?

temporalcli/commandsgen/commands.yml
cretz, line 2656: Arguably, like https://github.com/temporalio/cli/pull/729, `update-options` should be the command and let the args determine which values are set on the `UpdateWorkflowExecutionOptions` call. That way, later when we have more options, users can set multiple options at once and it makes it more like the API we expose over gRPC/HTTP. Having separate commands to make the same API call in this case I think is confusing and limiting.

antlai-temporal: This was intentional. Otherwise it gets listed under `update`, and it gets very confusing because people will assume it is a global option for updates.  Activities don't take updates, so they don't have that issue...

temporalcli/commands.deployment.go
antlai-temporal: Let me check, it should do it before every save...

temporalcli/commandsgen/commands.yml
antlai-temporal: Using `update` as a verb should be banned in our APIs. I know CRUD, but we are making so much investment on "update" (as a noun or adjective) that it gets really confusing. For example:
```
temporal workflow
Available commands:
...
 terminate        Forcefully end a Workflow Execution
  trace            Workflow Execution live progress
  update           Updates (Experimental)
  update-options Change Workflow Execution Options
 ```
is update-options a global option setting for update?

File: temporalcli/commandsgen/commands.yml (hunk 1 above)
Comment (antlai-temporal) [reply temporalcli/commandsgen/commands.yml:485]: Maybe not, I'm on the fence... It is a bit verbose, but `series-name` is always part of a deployment, but the previous command scopes for `deployment` anyway...

File: temporalcli/commandsgen/commands.yml (hunk 2 above)
Comment (cretz) [reply temporalcli/commandsgen/commands.yml:2656]: Activities may take updates in the future, but I think we should call this "update options" in both the API and the CLI. And we do for the latter. We are ok saying "update" and "update options" are different I think.

Comment (cretz) [reply temporalcli/commandsgen/commands.yml:2656]: > Using update as a verb should be banned in our APIs.

But it's not banned, it's in use, including the API call this command is making

File: temporalcli/commands.deployment.go (hunk 3 above)
Comment (antlai-temporal) [reply temporalcli/commands.deployment.go]: Done

File: temporalcli/commandsgen/commands.yml (hunk 2 above)
Comment (antlai-temporal) [reply temporalcli/commandsgen/commands.yml:2656]: About whether to split in sub-commands or not. The trade-off is to make it easier to use, with a small number of relevant arguments, vs doing atomic changes to several fields. Not clear to me that you will need to, e.g., change the memo and also the versioning override atomically, they feel really different. 
BTW, the PR for update activity options ended up with 11 arguments, 10 of them optional... 
I guess we could always add a bulk update later if we need it, but not convinced...

Comment (carlydf) [reply temporalcli/commandsgen/commands.yml:2656]: I also feel that it makes more sense to call this `workflow update-options`. We've discussed the modify vs update name choice considering the context of `workflow update` and `activity update-options`, and I think that `update` and `update-options` can be separate things.

Comment (carlydf) [reply temporalcli/commandsgen/commands.yml:2656]: We definitely want to be able to change things like timeouts and priority atomically when we change the pinned version, that's part of why the gRPC API was designed the way that it is.

Comment (carlydf) [reply temporalcli/commandsgen/commands.yml:2656]: At least that's what I intended

File: temporalcli/commandsgen/commands.yml [hunk 4]
```diff
@@ -394,6 +394,160 @@ commands:
...
+  - name: temporal deployment describe
+    summary: Show properties of a Deployment
+    description: |
+      +---------------------------------------------------------------------+
+      | CAUTION: Worker Deployment is experimental. Deployment commands are |
+      | subject to change.                                                  |
+      +---------------------------------------------------------------------+
+
+      Describes properties of a Deployment, such as whether it is current,
+      some custom metadata, or a list of its task queues.
```
carlydf, line 439: If we always return all three of these things, the word "or" is misleading.
```suggestion
      Describes properties of a Deployment, such as whether it is current,
      the list of its task queues, custom metadata if present, and reachability
      status if requested.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+    summary: Show the current Deployment
+    description: |
+      +---------------------------------------------------------------------+
+      | CAUTION: Worker Deployment is experimental. Deployment commands are |
+      | subject to change.                                                  |
+      +---------------------------------------------------------------------+
+
+      Gets the current Deployment for a Deployment Series Name. Whether a
+      Deployment is current or not can affect which Workers will execute
+      Tasks of an existing or new Workflow.
```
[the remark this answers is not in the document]
carlydf: I'm not picky about how this is worded, but I think this is an opportunity to be specific with users about what "current" means for task routing.

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+      | CAUTION: Worker deployment is experimental. Deployment commands are |
+      | subject to change.                                                  |
+      +---------------------------------------------------------------------+
+
+      Sets the current Deployment for a given Deployment Series. Whether a
+      Deployment is current or not can affect which Workers will execute
+      Tasks of an existing or new Workflow.
+
+      ```
+      temporal deployment update-current [options]
```
carlydf, line 530: Same as above, `set-current` seems more straightforward to me and matches the API

temporalcli/commandsgen/commands.yml
carlydf: I think it's helpful to have the "override" word in the flag name, so people know what they're setting. 
Like:
```
# Pin a workflow to a specific Deployment
temporal workflow update-options --workflow-id MY_WF [--run-id="..."] \
     --versioning-behavior-override=pinned \
     --deployment-series-name-override=MY_SERIES \  # required for pinned override
     --deployment-build-id-override=MY_BLD                 # required for pinned override
```
or 
```
# Let this workflow be executed by the current Deployment and automatically move to future current Deployments.
temporal workflow update-options --workflow-id MY_WF [--run-id="..."] \
     --versioning-behavior-override=auto_upgrade
```
or
```
# Unset any versioning override that may be set, and let this workflow be handled per its base versioning behavior.
temporal workflow update-options --workflow-id MY_WF [--run-id="..."] \
     --versioning-behavior-override=unspecified
```

antlai-temporal: The goal of the subcommand "versioning-override" was to simplify the arguments, but if there is a legitimate goal for atomic updates of other properties, we will need to add the prefix to the options, and get rid of the subcommand. @drewhoskins-temporal @ShahabT  any thoughts?

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+      description-header: >-
+        Temporal Deployment commands enable operations on Worker Deployments,
+        such as describe, list, update-current, and get-current, simplifying
+        versioning and management of workers.
+      keywords:
+        - deployment
+        - deployment describe
+        - deployment list
+        - deployment get-current
+        - deployment update-current
```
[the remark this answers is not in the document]
antlai-temporal: Could do "set-current" to be consistent with API.  "update" or "modify" feels less implementation-oriented (setters/getters), but we already use "get" instead of "read"...

temporalcli/commandsgen/commands.yml
antlai-temporal: We don't print empty things, e.g., an empty list of task queues, but I guess the meaning is implicit. I prefer your wording though...

temporalcli/commandsgen/commands.yml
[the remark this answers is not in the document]
antlai-temporal: I'm in two minds about how much context to give in cli docs. Clearly, you cannot expect people to learn Deployments concepts from cli commands, but  a good reminder is sometimes useful, when it is not too repetitive...

temporalcli/commandsgen/commands.yml
antlai-temporal: Sure, they can be separate things, the issue is whether people will get confused with 'options of an update".
 I think using UpdateOptions in the API was a mistake, but the train may have already left the station, particularly, if activities in the CLI are already using it... Any thoughts @drewhoskins-temporal ?

temporalcli/commandsgen/commands.yml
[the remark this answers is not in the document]
carlydf: I don't feel strongly about this, but "set" feels like it describes the goal and impact of this call more directly than "update" which is more vague.
In general it seems less confusing to go with the same names as the gRPC API, and if there's a strong reason to use a different name, then the gRPC/HTTP APIs should probably also be renamed.

temporalcli/commandsgen/commands.yml
drewhoskins-temporal: +1 to @carlydf's take.  This should be consistent with `activity update-options` as well as the general expectation that such methods are called 'update'.

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
         description: Reason for terminating the batch job.
         required: true
 
+  - name: temporal deployment
```
drewhoskins-temporal, line 397: I think this should be called worker-deployment in this context where we're not in a worker.  Deployment is too ambiguous.

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+
+      ```
+      temporal deployment describe [options]
+      ```
+
+      For example, to also include reachability information:
+
+      ```
+      temporal deployment describe \
+          --deployment-series-name YourDeploymentSeriesName \
```
drewhoskins-temporal, line 449: why are we reiterating "deployment" when it's already at the top level?

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+          --deployment-build-id YourDeploymentBuildId \
```
drewhoskins-temporal, line 450: It doesn't have to be for the pre-release but I'd like to sure we're moving on/aligned on the details of https://www.notion.so/temporalio/Safe-Deploys-Deployment-IDs-Proposal-15b8fc567738809386fad0523f775d84 that we want to carry forward.

temporalcli/commandsgen/commands.yml
drewhoskins-temporal: I'm with chad.

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,160 @@ commands:
...
+      temporal deployment get-current \
+          --deployment-series-name YourDeploymentSeriesName
+      ```
+    options:
+      - name: deployment-series-name
+        type: string
+        description: Series Name for the current Deployment.
+        required: true
+
+  - name: temporal deployment list
```
drewhoskins-temporal, line 490: Not sure if you've been aware, but I've been talking to @ShahabT and Candace about this and am pushing against this design.  I think users either want a list of series, or a list of deployments within a specific series.  An API that intermingles them in a (presumably time-sorted) list doesn't serve a real use case AFAICT and is pretty weird.

antlai-temporal: I was also tempted to make the `series-name` attribute required, but the discussions on limits on number of Deployments, and lack of scavenger, may make it useful to enumerate them all, if anything just to count them...
Will revisit once we have a `list series-name` api...

temporalcli/commandsgen/commands.yml
antlai-temporal: See discussion above

temporalcli/commandsgen/commands.yml
antlai-temporal: Makes sense, renaming...

temporalcli/commandsgen/commands.yml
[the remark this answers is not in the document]
antlai-temporal: Changing to set-current

temporalcli/commandsgen/commands.yml
antlai-temporal: Yes, if we change the task queue we probably want to pin a new code atomically. We will still have the issue of tens of optional arguments though. I'm getting rid of the sub command for now, use prefix `versioning-override` for the options, and we will revisit when the argument list gets unwieldy.

antlai-temporal: OK, majority rules, changing to `update-options`...

temporalcli/commandsgen/commands.yml
antlai-temporal: OK,  removing `deployment-`...

temporalcli/commandsgen/commands.yml
cretz: It may be worth a discussion on whether we want this as `temporal worker deployment` instead of `temporal worker-deployment`. There is discussion of more server-side worker management features in the future, so I wonder if `worker` should be a category of commands instead of a prefix.

temporalcli/commandsgen/commands.yml
```diff
@@ -394,6 +394,162 @@ commands:
         description: Reason for terminating the batch job.
         required: true
 
+  - name: temporal worker-deployment
```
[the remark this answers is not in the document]
antlai-temporal: @drewhoskins-temporal you suggested `worker-deployment`. Are you aware of any other worker functionality coming that needs CLI support?
It may not be obvious to people that `deployment` is under worker, so I'm hesitant to hide it unless something is coming soon.
In any case `worker-deployment` commands will be experimental for a while, so we will have time to restructure if needed...

cretz, line 397: @drewhoskins-temporal - I noticed in our gRPC and HTTP API we chose to just call these "deployments" without the term worker at all. I like the "worker" qualifying term, but the inconsistency of not doing so for our API is unfortunate.

antlai-temporal:
Split `worker-deployment` into `worker deployment`
Merging without review into versioning-3. The real review will be a PR for merging versioning-3 into main.

Issue #720, opened 2024-12-03
[Feature Request] versioning override and move API


antlai-temporal:
The new Worker versioning API provides methods to change versioning configuration for a particular workflow or move it to a new Build ID. The CLI is the natural tool to do that.

Issue #719, opened 2024-12-03
[Feature Request] DescribeWorkflow should show versioning info


antlai-temporal:
The new Worker versioning API provides extra versioning information for running workflows that should be visible in the CLI.

Issue #718, opened 2024-12-03
[Feature Request] Support the deployment API


antlai-temporal:
Worker versioning-3 provides a new experimental API to manage deployments and access reachability information.

The goal is to add support in the CLI after Go SDK supports it.

PR #740, opened 2025-01-15
Versioning 3

antlai-temporal:
## What was changed

This is a merge of the branch `versioning-3` into main.

It adds support for the new Deployment API, extends Describe Workflow with versioning info, and 
provides a versioning override API.

See the  PR #735 for further details.

## Checklist

1. Closes 
#720 #719 #718


temporalcli/commands.taskqueue_test.go
```diff
@@ -261,6 +262,9 @@ func (s *SharedServerSuite) TestTaskQueue_Describe_Simple() {
 	)
 	s.NoError(res.Err)
 
+	// TODO(antlai-temporal): Delete when a server caching bug in 1.26.2 is fixed
```
cretz, line 265: Can you link to the issue (ideally in the code comment, but here is fine too)? And/or can you just make a CLI issue that we try not to lose that removes these sleeps once whatever upstream issue is fixed?

temporalcli/commands.workflow.go
```diff
@@ -88,6 +92,117 @@ func (c *TemporalWorkflowDeleteCommand) run(cctx *CommandContext, args []string)
...
+		})
+		if err != nil {
+			return fmt.Errorf("failed to update workflow options: %w", err)
+		}
+		cctx.Printer.Println("Update workflow options succeeded")
+	} else { // Run batch
+		var workflowExecutionOptions *workflowpb.WorkflowExecutionOptions
+		protoMask, err := fieldmaskpb.New(workflowExecutionOptions, "versioning_override")
+		if err != nil {
+			panic("invalid field mask")
```
cretz, line 161: While I know this thing shouldn't happen, I think we can still use regular errors here

cretz (approved):
LGTM, nothing blocking. I admit I didn't review the models too in depth, so long as we're reasonably comfortable with the JSON that is emitted will be stable (no problem with pre-GA incompatible changes, I just mean that we're happy in general with the shape/approach).

temporalcli/commands.deployment.go
```diff
@@ -0,0 +1,377 @@
+package temporalcli
```
josh-berry, line 1: I have a nit about the file name: if this is `temporal worker deployment ...`, we should probably name this file `commands.worker.deployment.go`. I know it's a mouthful but it's at least consistent.

temporalcli/commandsgen/commands.yml
```diff
@@ -605,6 +605,179 @@ commands:
...
 
+  - name: temporal worker
+    summary: Read or update Worker state
+    description: |
+      Modify or read state associated with a Worker, for example,
+      using Worker Deployments commands:
+
+      ```
+      temporal worker deployment
+      ```
```
josh-berry, line 616: Let's also mark this as experimental; no need to stabilize something unnecessarily. :)

temporalcli/commandsgen/commands.yml
```diff
@@ -605,6 +605,179 @@ commands:
...
+      - client
+    docs:
+      description-header: >-
+         Learn how to read or modify state associated with a Worker,
+         such as Worker Deployments.
+      keywords:
+        - worker
+        - worker deployment
+
+  - name: temporal worker deployment
```
josh-berry, line 627: Should we also deprecate the old `task-queue versioning` etc. commands? Or are those still relevant?

temporalcli/commandsgen/commands.yml
```diff
@@ -605,6 +605,179 @@ commands:
...
+          --build-id YourDeploymentBuildId \
+          --report-reachability
+      ```
+    option-sets:
+      - deployment-reference
+    options:
+      - name: report-reachability
+        type: bool
+        description: |
+          Flag to include reachability information of a Worker Deployment.
```
josh-berry, line 688: You don't need to say it's a flag. That's obvious in context:

```suggestion
          Include reachability information of a Worker Deployment.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -2707,6 +2881,67 @@ commands:
         type: int
         description: Maximum number of Workflow Executions to display.
 
+  - name: temporal workflow update-options
```
josh-berry, line 2884: [non-blocking/feel free to ignore] Would it be better to call this `set-options` so as not to confuse it with Workflow Update? Or maybe even `options set` if we think we'll add `options list`, `options show` etc. later?

temporalcli/commandsgen/commands.yml
```diff
@@ -2707,6 +2881,67 @@ commands:
...
+    summary: Change Workflow Execution Options
+    description: |
+      Modify properties of Workflow Executions:
+
+      ```
+      temporal workflow update-options [options]
+      ```
```
josh-berry, line 2891: Should this also be experimental?

temporalcli/commandsgen/commands.yml
```diff
@@ -2707,6 +2881,67 @@ commands:
...
+
+      To see the current override use `temporal workflow describe`
+
+    option-sets:
+      - single-workflow-or-batch
+    options:
+      - name: versioning-override-behavior
+        type: string-enum
+        description: |
+          Flag to override the versioning behavior of a Workflow.
```
josh-berry, line 2932:
```suggestion
          Override the versioning behavior of a Workflow.
```

Also: I'm a little unclear on what "override" means here—it might be good to explain what each of the enum values mean.

josh-berry (commented):
Part one of my review, just the API/YAML. Dropping quickly so you have more time to consider/respond. Will look at the actual code next.

josh-berry (commented):
Took a quick pass at the actual code and the accompanying tests and I have no more comments, the implementation LGTM. 👍

temporalcli/commands.deployment.go
antlai-temporal: Renaming...

temporalcli/commandsgen/commands.yml
antlai-temporal: Not yet, we have not deprecated in the go sdk, the consensus was to wait for public preview...

temporalcli/commandsgen/commands.yml
antlai-temporal: Thanks, much better

temporalcli/commandsgen/commands.yml
antlai-temporal: I don't want to open the can of worms again... It took hours...If we decide to change things it should start with the API and align with activity options...

temporalcli/commandsgen/commands.yml
antlai-temporal: Good catch, thanks

temporalcli/commandsgen/commands.yml
antlai-temporal: Thanks, I modified the description to give examples of the enum values.

temporalcli/commands.taskqueue_test.go
antlai-temporal: Added link, and opened issue https://github.com/temporalio/cli/issues/741

PR #742, opened 2025-01-17
User Metadata set on start / desc

Sushisource:
## What was changed
* Added `--static-summary` and `--static-details` to all workflow-start like operations
* Added summary & details to workflow describe

## Why?
Useful feature

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.workflow_view.go
+++ b/temporalcli/commands.workflow_view.go
@@ -134,6 +134,20 @@ func (c *TemporalWorkflowDescribeCommand) run(cctx *CommandContext, args []strin
 		HistorySize:          info.HistorySizeBytes,
 	}, printer.StructuredOptions{})
 
+	staticSummary := resp.GetExecutionConfig().GetUserMetadata().GetSummary()
+	staticDetails := resp.GetExecutionConfig().GetUserMetadata().GetDetails()
+	if len(staticSummary.GetData()) > 0 || len(staticDetails.GetData()) > 0 {
+		cctx.Printer.Println()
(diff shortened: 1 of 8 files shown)


Sushisource:
Actually, looks like we need to follow up on https://github.com/temporalio/temporal/issues/6412 to make this work properly with schedule creation too

temporalcli/commands.workflow_view.go
```diff
@@ -134,6 +134,20 @@ func (c *TemporalWorkflowDescribeCommand) run(cctx *CommandContext, args []strin
...
+	staticDetails := resp.GetExecutionConfig().GetUserMetadata().GetDetails()
+	if len(staticSummary.GetData()) > 0 || len(staticDetails.GetData()) > 0 {
+		cctx.Printer.Println()
+		cctx.Printer.Println(color.MagentaString("Metadata:"))
+		_ = cctx.Printer.PrintStructured(struct {
+			StaticSummary *common.Payload
+			StaticDetails *common.Payload
+		}{
+			StaticSummary: staticSummary,
+			StaticDetails: staticDetails,
```
cretz, line 147: I wonder if there is value putting this information in the above structured print struct, though I don't mind it being separate

cretz (approved):

Sushisource:
Going to wait on https://github.com/temporalio/sdk-go/pull/1782 to be released (or can depend on unreleased version) so this can include schedules

PR #743, opened 2025-01-22
Add `workflow metadata` command

Sushisource:
## What was changed
* Added new `workflow metadata` command to run the built-in metadata query and show results

## Why?
Useful to run this from the CLI

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commandsgen/commands.yml
+++ b/temporalcli/commandsgen/commands.yml
@@ -206,7 +206,7 @@ commands:
   - name: temporal activity
     summary: Complete, update, pause, unpause, reset or fail an Activity
     description: |
-      Update an Activity's options, manage activity lifecycle or update 
+      Update an Activity's options, manage activity lifecycle or update
       an Activity's state to completed or failed.
 
       Updating activity state marks an Activity as successfully finished or as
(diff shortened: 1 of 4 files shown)


temporalcli/commandsgen/commands.yml
```diff
@@ -2886,6 +2887,20 @@ commands:
...
+      Issue a Query for and display user-set metadata like summary and
+      details for a specific Workflow Execution:
+
+      ```
+      temporal workflow metadata \
+          --workflow-id YourWorkflowId
+      ```
+    option-sets:
+      - workflow-reference
+      - query-modifiers
```
cretz, line 2902: I'd be ok if we didn't even accept query reject condition, but meh

cretz (approved):

dandavison:
>  I dunno if we want to recursively headerize this or not

I think that would make sense and be what most users would expect. At least, having one line per handler

Sushisource:
Looks like this now:

```
         Metadata:
          Query Definitions:
            __open_sessions                 Open sessions on the workflow
            __stack_trace                   Current stack trace          
            __temporal_workflow_metadata    Metadata about the workflow  
            my-query                        q-desc                       
          Signal Definitions:
            my-signal    sig-desc   
          Update Definitions:
            my-update    upd-desc   
          Current Details:
            current-deets

```

dandavison:
> Looks like this now:

WDYT about the alignment -- would be nice to have the descriptions aligned, or perhaps just don't indent them at all if we're not going to enforce an alignment that works for all rows.

dandavison (approved):

PR #747, opened 2025-02-04
Add batch activity unpause.

ychebotarev:
## What was changed

Update "activity unpause" API to support batch operation.

## Why?

## Checklist

1. Closes 
https://temporalio.atlassian.net/browse/OSS-3789

2. How was this tested:

By adding test.

3. Any docs updates needed?

Not yet.

### DIFF
--- a/temporalcli/commands.activity_test.go
+++ b/temporalcli/commands.activity_test.go
@@ -323,6 +323,25 @@ func (s *SharedServerSuite) getActivityEvents(workflowID, activityID string) (
 	return started, completed, failed
 }
 
+func checkActivitiesRunning(s *SharedServerSuite, run client.WorkflowRun) {
+	s.Eventually(func() bool {
+		resp, err := s.Client.DescribeWorkflowExecution(s.Context, run.GetID(), run.GetRunID())
+		s.NoError(err)
+		return len(resp.GetPendingActivities()) > 0
+	}, 5*time.Second, 200*time.Millisecond)
+}
+
+func checkActivitiesPaused(s *SharedServerSuite, run client.WorkflowRun) {
+	s.Eventually(func() bool {
+		resp, err := s.Client.DescribeWorkflowExecution(s.Context, run.GetID(), run.GetRunID())
+		s.NoError(err)
+		if resp.GetPendingActivities() == nil {
+			return false
+		}
+		return len(resp.GetPendingActivities()) > 0 && resp.GetPendingActivities()[0].Paused
+	}, 5*time.Second, 200*time.Millisecond)
+}
+
 func (s *SharedServerSuite) TestUnpauseActivity_BatchSuccess() {
 	var failActivity atomic.Bool
 	failActivity.Store(true)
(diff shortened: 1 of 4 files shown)


temporalcli/commandsgen/commands.yml
```diff
@@ -449,10 +458,21 @@ commands:
           --reset-heartbeats
       ```
     options:
```
cretz, line 460: Can this just use the `single-workflow-or-batch` option set and only add the options needed specifically for this?

temporalcli/commands.activity_test.go
```diff
@@ -338,3 +322,98 @@ func (s *SharedServerSuite) getActivityEvents(workflowID, activityID string) (
...
+	// Send batch activity unpause
+	cmdRes := s.Execute("activity", "unpause",
+		"--rps", "1",
+		"--address", s.Address(),
+		"--query", query,
+		"--reason", "unpause-test",
+		"--yes", "--match-all",
+	)
+	s.NoError(cmdRes.Err)
+	s.NotEmpty(startBatchRequest.JobId)
```
bergundy, line 418: Should the job ID be printed as part of the output? What do we do for other commands?

temporalcli/commands.activity_test.go
```diff
@@ -338,3 +322,98 @@ func (s *SharedServerSuite) getActivityEvents(workflowID, activityID string) (
...
+			}
+			lastRequestLock.Unlock()
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}),
+	)
+
+	failActivity.Store(false)
+
+	// Send batch activity unpause
+	cmdRes := s.Execute("activity", "unpause",
```
bergundy, line 410: You don't really test that the activities were unpaused. Would be nice to have that coverage.
You'll also probably want to have an `Eventually` block that verifies all of the workflows show up in visibility before issuing the batch unpause command.

temporalcli/commandsgen/commands.yml
```diff
@@ -467,8 +487,35 @@ commands:
...
+        description: |
+          Content for an SQL-like `QUERY` List Filter.
+      - name: reason
+        type: string
+        description: |
+          Reason for unpause. Can only be used with --query.
+      - name: match-all
+        type: bool
+        description: |
+          Every paused activity should be unpaused.
```
bergundy, line 502: Say that this is mutually exclusive with `--activity-type`.

bergundy (approved):
Approved with comments.

temporalcli/commandsgen/commands.yml
ychebotarev: I think about this, but then I check "temporal workflow terminate", and for some reason it was not used, so I don't use it either. But it see it is used in other batch operations like "cancel" etc. 
Hm, not sure why it was not used in "terminate". I will add it.

cretz: I too am unsure why it wasn't used in terminate. Digging, I see this was brought up at https://github.com/temporalio/cli/pull/436#discussion_r1483665633 and just the fact that the "reason" docs were slightly different is why they did it. My comment was too late, they had already merged.

temporalcli/commands.activity_test.go
ychebotarev: It doesn't. 
In the other commands we check that jobId is printed for ->json<- output format:
```
	s.NoError(json.Unmarshal(res.Stdout.Bytes(), &jsonRes))
	s.NotEmpty(jsonRes["batchJobId"])
```

temporalcli/commands.activity_test.go
ychebotarev: This is CLI test, not func test. I need to check that command was accepted by the server. 
The test that checks that unpause is working in exactly this scenario exist in the server functional tests. And there I do check that functionality is working e2e.

temporalcli/commandsgen/commands.yml
ychebotarev: will add

temporalcli/commands.activity_test.go
josh-berry: [We have talked about this previously.](https://github.com/temporalio/cli/pull/732#discussion_r1891988616) Testing that the command has the desired effect is the standard we expect from CLI tests. I'm glad that you have tests for this in the server, but that's irrelevant for the CLI.

ychebotarev: Ok added. I also already have "block that verifies all of the workflows show up in visibility".

cretz (approved):
Just a curious question about `--match-all`, nothing blocking

temporalcli/commandsgen/commands.yml
```diff
@@ -467,8 +477,20 @@ commands:
         type: bool
         description: |
           Reset the Activity's heartbeats. Only works with --reset-attempts.
+      - name: match-all
+        type: bool
+        description: |
+          Every paused activity should be unpaused. This flag is ignored if 
+          activity-type is provided. Can only be used with --query.
```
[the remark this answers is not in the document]
ychebotarev: This:
```
	case BatchTypeUnpauseActivities:
		if params.UnpauseActivitiesParams.ActivityType == "" && !params.UnpauseActivitiesParams.MatchAll {
			return fmt.Errorf("must provide ActivityType or MatchAll flag")
		}

```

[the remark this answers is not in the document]
cretz: Can you make that clear in the CLI docs that one of these two is required. It was a bit hard to tell.

PR #748, opened 2025-02-07
Specify pagesize for workflow list command

yiminc:
## What was changed

Add optional pagesize parameter to workflow list command

## Why?

Be able to specify page size is useful when default page is too large for cases where it leads to timeout due to large amount of data. 

## Checklist

1. Closes 

2. How was this tested:

Updated unit test to cover it.

3. Any docs updates needed?


temporalcli/commandsgen/commands.yml
```diff
@@ -2886,6 +2886,9 @@ commands:
       - name: limit
         type: int
         description: Maximum number of Workflow Executions to display.
+      - name: pagesize
```
josh-berry, line 2889: To be consistent with how we name other flags, this should be:

```suggestion
      - name: page-size
```

temporalcli/commandsgen/commands.yml
```diff
@@ -2886,6 +2886,9 @@ commands:
...
+        type: int
+        description: Maximum number of Workflow Executions to return from server per request. Used for pagination.
```
josh-berry, line 2891: Optional; rewords in terms of behavior, not implementation.

```suggestion
        description: Maximum number of Workflow Executions to fetch at a time from the server.
```

josh-berry (approved):
LGTM (assuming we decide to stay with the naming etc. per offline discussion)

PR #752, opened 2025-02-12
Update dependencies

rodrigozhou:
## What was changed

Update dependencies

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

rodrigozhou:
User had reported some security issue in the `golang.org/x/crypto` package that needs to be the latest, but as I'm digging into this, I think the issues reported by the user were fixed already in older version. I think it's already impacting the `tctl` tool, which is included in the `admin-tools` image, which is what the user saw.
So at this moment, I think it's just updating for the sake of updating...

bergundy (approved):
@josh-berry this will make the merge of the `next-server` branch a bit complicated FYI.

PR #755, opened 2025-02-13
Add codec-header config (closes #753)

### LINKED ISSUE #753: [Feature Request] Allow setting custom header on codec server requests

We have created a codec server and used Cloudflare Access to secure the endpoints. To get through Cloudflare Access you need to set the [cf-access-token header](https://developers.cloudflare.com/cloudflare-one/identity/service-tokens/#subsequent-requests). It looks like this is not currently possible as the codec-auth config parameter sets the Authorization header only.

Add in a more flexible codec-header config parameter (in the same form as the grpc-meta parameter) so the header name can be set and even multiple headers be provided if needs be. I made a branch to test out the idea. Happy to raise a PR if you like.

https://github.com/EddieWhi/cli/tree/add-codec-headers-param

EddieWhi:
To set any header other than Authorization, the codec-header config parameter can be used.

## What was changed

Added the code-header property using the same pattern as grpc-meta.
Grammatical change to description of grpc-meta (it's got punctuation when it gets formatted to a single line).
Refracted the conversion from config to `stringMapHeadersProvider` to a factory function. (Slight change to existing error message here).

## Why?

I needed to set the cf-access-token header on codec server requests (see #753)

## Checklist

1. Closes #753

2. How was this tested:

- queried workflow history with codec server behind Cloudflare Access using the cf-access-token header.
- checked grpc-meta still works as before.
- checked validation of grpc-meta still works and codec-header validation works too.

Have NOT tested the codec-auth parameter, although it should still work (just because I haven't got a simple way to test and hope someone else does without me having to set something specific up). 

3. Any docs updates needed?

Could update list of props in[ this section](https://docs.temporal.io/cli#customize-your-environment-variables), although this section look out of date anyway (i.e. `temporal env set -h` does not return a full list of available properties).


cretz (approved):
This looks great, let me get another team member approval and we'll merge (could be middle of next week due to holidays).

josh-berry (approved):

PR #756, opened 2025-02-15
Rebase from main, update deps, test fixes

antlai-temporal:
## What was changed

This brings `main` changes into `next-server`, upgrades server and other deps, and fixes some tests for the new server/api.


temporalcli/commands.taskqueue_test.go
```diff
@@ -261,6 +262,10 @@ func (s *SharedServerSuite) TestTaskQueue_Describe_Simple() {
 	)
 	s.NoError(res.Err)
 
+	// TODO(antlai-temporal): Delete when a server caching bug in 1.26.2 is fixed,
+	// see https://github.com/temporalio/temporal/pull/6978
+	time.Sleep(1 * time.Second)
```
cretz, line 267: It looks like this is already fixed in the server version in use? Can this sleep and the later one be removed?

temporalcli/commands.taskqueue_test.go
```diff
@@ -2,11 +2,12 @@ package temporalcli_test
...
 import (
 	"encoding/json"
-	"github.com/stretchr/testify/assert"
-	"go.temporal.io/sdk/workflow"
 	"strings"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"go.temporal.io/sdk/workflow"
+
```
cretz, line 10: Some of these changes are confusing to me showing up in the PR. So this change is already in `next-server`, why is it showing up in the PR?

cretz (approved):
LGTM

temporalcli/commands.taskqueue_test.go
antlai-temporal: Yes, this coming with my versioning-3 PR right after. This one just enables a draft release with the new server for testing the go sdk.

PR #758, opened 2025-02-20
Add `temporal workflow signal-with-start` command

### LINKED ISSUE #664: UpdateWithStart SDK API
Expose update-with-start operations via the SDK

THardy98:
Added `temporal workflow signal-with-start` command.

Description and sample usage:

      Send an asynchronous notification (Signal) to a Workflow Execution.
      If the Workflow Execution is not running or is not found, it starts the 
      workflow then sends the signal.

      ```
      temporal workflow signal-with-start \
        --signal-name YourSignal \
        --signal-input '{"some-key": "some-value"}' \
        --type YourWorkflowType \
        --task-queue YourTaskQueue \
        --input '{"some-key": "some-value"}'
      ```

1. Closes #664

2. How was this tested:
Added unit tests

3. Any docs updates needed?
Yes


THardy98:
**Note**: the command doesn't configure `VersioningOverride` in the `signal-with-start` request. It would basically need replicating the logic in Go SDK to convert it to proto. I opted to omit the duplication since it was marked `experimental` anyway and added a warning.

Let me know if we usually opt for a different approach here.

temporalcli/commands.workflow.go
```diff
@@ -262,6 +266,96 @@ func (c *TemporalWorkflowSignalCommand) run(cctx *CommandContext, args []string)
...
+		}
+	}
+	var memo *common.Memo
+	if wfStartOpts.Memo != nil {
+		fields, err := encodeMapToPayloads(wfStartOpts.Memo)
+		if err != nil {
+			return err
+		}
+		memo = &common.Memo{Fields: fields}
+	}
```
cretz, line 317: I think this kind of logic needs to match `temporal workflow start` (which uses `stringKeysJSONValues` for memo). In fact, looking at this in general, is there a reason not to use the Go SDK's high-level `SignalWithStartWorkflow` call instead of the raw RPC? Then you can just reuse the `buildStartOptions` helper that exists. Also I would consider moving this into `commands.workflow_exec.go` with its other start workflow friends.

THardy98: There's a comment above the workflow service call below - `SignalWithStartWorkflow` only accepts a single arg for signal. The existing signal command on CLI uses the workflow service call because of this. I think it makes sense to provide the flexibility for multiple signal args and stay consistent with the existing CLI signal command

cretz: I see, that is unfortunate that Go SDK is that way but I agree this needs to support multiple signal arguments. I fear that people that implement a new start option may not know to come update this. But I cannot think of a better way short of having Go SDK support a signal-with-start that takes multiple signal parameters. @Quinn-With-Two-Ns - thoughts here?

I do think it still may make more sense to move that `encodeMapToPayloads` utility that used to be for schedules only into this file for more generic use, and maybe move this command impl to `commands.workflow_exec.go` next to the other commands that work with the workflow start options.

temporalcli/commands.workflow.go
```diff
@@ -262,6 +266,96 @@ func (c *TemporalWorkflowSignalCommand) run(cctx *CommandContext, args []string)
...
+			Identity:                 clientIdentity(),
+			RetryPolicy:              retryPolicy,
+			CronSchedule:             wfStartOpts.CronSchedule,
+			Memo:                     memo,
+			SearchAttributes:         searchAttr,
+			WorkflowIdReusePolicy:    wfStartOpts.WorkflowIDReusePolicy,
+			WorkflowIdConflictPolicy: wfStartOpts.WorkflowIDConflictPolicy,
+		},
+	)
+	return err
```
cretz, line 356: The `startWorkflow` helper when called from `workflow start` prints a structured output like:
```go
		cctx.Printer.Println(color.MagentaString("Running execution:"))
		err := cctx.Printer.PrintStructured(struct {
			WorkflowId string `json:"workflowId"`
			RunId      string `json:"runId"`
			Type       string `json:"type"`
			Namespace  string `json:"namespace"`
			TaskQueue  string `json:"taskQueue"`
		}{
			WorkflowId: run.GetID(),
			RunId:      run.GetRunID(),
			Type:       sharedWorkflowOpts.Type,
			Namespace:  c.Namespace,
			TaskQueue:  sharedWorkflowOpts.TaskQueue,
		}, printer.StructuredOptions{})
```

I think we should match that output here too

cretz (commented):
Only minor suggestions regarding moving code a bit and providing output.

I do wonder if we should do update-with-start at the same time in case we learn something about the approach we take there that can affect signal-with-start. But maybe not since the result types/approaches are so different.

temporalcli/commands.workflow.go
THardy98: `signal-with-start` moved to `commands.workflow_exec.go` and `encodeMapToPayloads` to `commands.go`

temporalcli/commands.workflow.go
THardy98: Added (along with a check in test for expected output)

temporalcli/commands.workflow_exec.go
```diff
@@ -92,6 +98,116 @@ func (c *TemporalWorkflowExecuteCommand) run(cctx *CommandContext, args []string
...
+		WorkflowId: c.WorkflowId,
+		RunId:      resp.RunId,
+		Type:       c.Type,
+		Namespace:  c.Parent.Namespace,
+		TaskQueue:  c.TaskQueue,
+	}, printer.StructuredOptions{})
+	if err != nil {
+		return fmt.Errorf("failed printing: %w", err)
+	}
+	return nil
```
cretz, line 208:
```suggestion
	return cctx.Printer.PrintStructured(struct {
		WorkflowId string `json:"workflowId"`
		RunId      string `json:"runId"`
		Type       string `json:"type"`
		Namespace  string `json:"namespace"`
		TaskQueue  string `json:"taskQueue"`
	}{
		WorkflowId: c.WorkflowId,
		RunId:      resp.RunId,
		Type:       c.Type,
		Namespace:  c.Parent.Namespace,
		TaskQueue:  c.TaskQueue,
	}, printer.StructuredOptions{})
```

Pedantic, but don't really need to wrap printing failures I don't think (even though we did in some other places)

cretz (commented):
This LGTM. Only reason I hesitate to mark approved is because I'd like to see at least a high-level discussed/approved design of update with start in CLI just to make sure we're not missing anything.

THardy98:
> This LGTM. Only reason I hesitate to mark approved is because I'd like to see at least a high-level discussed/approved design of update with start in CLI just to make sure we're not missing anything.

Ok, I can put up a separate `update-with-start` PR before this gets merged to get a better picture.

temporalcli/commands.workflow_exec.go
THardy98: Done.

cretz (approved):
After internal discussion, this is like what we'll do with update-with-start, LGTM

PR #762, opened 2025-02-25
added `workflow start-update-with-start`  and `workflow execute-update-with-start`  commands

### LINKED ISSUE #664: UpdateWithStart SDK API

THardy98:
added `temporal workflow start-update-with-start`  and `temporal workflow execute-update-with-start`  commands

`temporal workflow start-update-with-start` usage: 
```
    temporal workflow start-update-with-start \
      --update-name YourUpdate \
      --update-input '{"update-key": "update-value"}' \
      --type YourWorkflowType \
      --task-queue YourTaskQueue \
      --input '{"wf-key": "wf-value"}'
    ```

`temporal workflow execute-update-with-start` usage:
```
    temporal workflow execute-update-with-start \
      --update-name YourUpdate \
      --update-input '{"update-key": "update-value"}' \
      --type YourWorkflowType \
      --task-queue YourTaskQueue \
      --input '{"wf-key": "wf-value"}'
```

1. Closes #664

2. How was this tested:

3. Any docs updates needed?
Yes

### DIFF
--- a/temporalcli/commands.workflow_exec.go
+++ b/temporalcli/commands.workflow_exec.go
@@ -121,7 +121,7 @@ func (c *TemporalWorkflowSignalWithStartCommand) run(cctx *CommandContext, _ []s
 	signalPayloadInputOpts := PayloadInputOptions{
 		Input:       c.SignalInput,
 		InputFile:   c.SignalInputFile,
-		InputMeta:   c.InputMeta,
+		InputMeta:   c.SignalInputMeta,
 		InputBase64: c.SignalInputBase64,
 	}
 	signalInput, err := signalPayloadInputOpts.buildRawInputPayloads()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by THardy98** [comment]
Need to add tests, but feel free to take a look at the command interface and behaviour

File: temporalcli/commands.workflow_exec.go [hunk 1]
```diff
@@ -204,6 +204,168 @@ func (c *TemporalWorkflowSignalWithStartCommand) run(cctx *CommandContext, _ []s
...
+	)
+	if err != nil {
+		return err
+	}
+
+	var valuePtr interface{}
+	err = handle.Get(cctx, &valuePtr)
+	if err != nil {
+		return fmt.Errorf("unable to update workflow: %w", err)
+	}
```
cretz, line 305: There's a bit of a problem with this approach and the one taken by `temporal workflow update execute`. This presumes that the result can be JSON decoded, but many people use protos or other formats. On user-defined results we need to support all payloads (gracefully using JSON where we can). I believe this is the first time we have needed to get a raw payload from a high-level client call to display to a user.

I think we need to add support for `rawValuePayloadConverter.FromPayload` in `client.go`, then change the code in this PR to be something like:

```go
var raw RawValue
if err = handle.Get(cctx, raw); err != nil {
	return fmt.Errorf("unable to update workflow: %w", err)
}
result, err := cctx.MarshalFriendlyJSONPayloads(raw.Payload)
```

And have `Result` in the structure be a `json.RawMessage`. Same for `temporal workflow update`. And confirm it looks right even with `-o json`. Thoughts @THardy98 and @dandavison?

temporalcli/commandsgen/commands.yml
```diff
@@ -3439,6 +3439,160 @@ commands:
...
+        --update-input '{"update-key": "update-value"}' \
+        --workflow-id YourWorkflowId \
+        --type YourWorkflowType \
+        --task-queue YourTaskQueue \
+        --input '{"wf-key": "wf-value"}'
+      ```
+    option-sets:
+      # workflow-id is "required" (runtime check)
+      - shared-workflow-start
+      - workflow-start
```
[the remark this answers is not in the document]
THardy98: Yes, is there a reason why we wouldn't want to supply workflow start options?

temporalcli/commandsgen/commands.yml
```diff
@@ -3439,6 +3439,160 @@ commands:
...
+        --update-name YourUpdate \
+        --update-input '{"update-key": "update-value"}' \
+        --workflow-id YourWorkflowId \
+        --type YourWorkflowType \
+        --task-queue YourTaskQueue \
+        --input '{"wf-key": "wf-value"}'
+      ```
+    option-sets:
+      # workflow-id is "required" (runtime check)
+      - shared-workflow-start
```
dandavison, line 3462: We want `id-conflict-policy` to be required for update-with-start -- all the SDKs do this. (I wonder how exciting of an implementation challenge that is in the CLI options.)

temporalcli/commandsgen/commands.yml
```diff
@@ -3439,6 +3439,160 @@ commands:
...
+      Experimental.
+
+      ```
+      temporal workflow start-update-with-start \
+        --update-name YourUpdate \
+        --update-input '{"update-key": "update-value"}' \
+        --workflow-id YourWorkflowId \
+        --type YourWorkflowType \
+        --task-queue YourTaskQueue \
+        --input '{"wf-key": "wf-value"}'
```
dandavison, line 3458: This example doesn't show `--update-wait-for-stage`, but it is a required option.
(`--id-conflict-policy` is also required)

temporalcli/commandsgen/commands.yml
```diff
@@ -3439,6 +3439,160 @@ commands:
...
+      Experimental.
+
+      ```
+      temporal workflow execute-update-with-start \
+        --update-name YourUpdate \
+        --update-input '{"update-key": "update-value"}' \
+        --workflow-id YourWorkflowId \
+        --type YourWorkflowType \
+        --task-queue YourTaskQueue \
+        --input '{"wf-key": "wf-value"}'
```
dandavison, line 3539: Will need `--id-conflict-policy` here.

temporalcli/commandsgen/commands.yml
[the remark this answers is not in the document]
cretz: Oh, my mistake, I though `shared-workflow-start` and `workflow-start` were mutually exclusive, can ignore/resolve this

temporalcli/commandsgen/commands.yml
cretz: Also let's add some tests

THardy98:
I removed `TaskQueue` from stdout ouput for the `-with-start` commands because we cannot guarantee the supplied `TaskQueue` is the one the signal/update ran on (i.e. in the scenario that an existing workflow with the same ID ran on a different task queue).

temporalcli/commands.workflow_exec.go
```diff
@@ -204,6 +204,168 @@ func (c *TemporalWorkflowSignalWithStartCommand) run(cctx *CommandContext, _ []s
...
+	cctx.Printer.Println(color.MagentaString("Running execution:"))
+	return cctx.Printer.PrintStructured(struct {
+		WorkflowId string      `json:"workflowId"`
+		RunId      string      `json:"runId"`
+		Type       string      `json:"type"`
+		Namespace  string      `json:"namespace"`
+		TaskQueue  string      `json:"taskQueue"`
+		UpdateName string      `json:"updateName"`
+		UpdateID   string      `json:"updateId"`
+		Result     interface{} `json:"result"`
```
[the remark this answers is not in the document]
THardy98: Done.

temporalcli/commandsgen/commands.yml
THardy98: Added a runtime check making `id-conflict-policy` required

temporalcli/commandsgen/commands.yml
THardy98: Added both to the example.
Added `--wait-for-stage` to the `temporal workflow update start` example as well (also required).

temporalcli/commandsgen/commands.yml
THardy98: Done.

temporalcli/commands.workflow_exec.go
```diff
@@ -194,16 +195,191 @@ func (c *TemporalWorkflowSignalWithStartCommand) run(cctx *CommandContext, _ []s
...
+		// Use a canceled context to check whether the initial server response
+		// shows that the update has _already_ failed, without issuing a second request.
+		ctx, cancel := context.WithCancel(cctx)
+		cancel()
+		err = handle.Get(ctx, nil)
+		var timeoutOrCanceledErr *client.WorkflowUpdateServiceTimeoutOrCanceledError
+		if err != nil && !errors.As(err, &timeoutOrCanceledErr) {
+			return fmt.Errorf("unable to update workflow: %w", err)
+		}
+	}
```
cretz, line 260: Hrmm, I don't think `UpdateWithStartWorkflow` will return until the wait stage is reached, and if there is some error, it will be reflected in the error to that call. @dandavison - is this correct?

dandavison: An example is sending a request for an update name that does not exist. `err` is `nil`, but the update is not accepted, and `handle` contains within it an error, hence this code (which is also present in CLI update impl) to extract that error without making a network request.

cretz: Ah, I see, there is "accepted but failed". We should consider adding a `Done() bool` to `WorkflowUpdateHandle` for this, but in the meantime, all good here.

THardy98:
TYFR :)

PR #764, opened 2025-02-25
💥 Support environment config [BREAKING CHANGES]

### LINKED ISSUE #751: [Feature Request] External client configuration

See See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md. Implement once https://github.com/temporalio/sdk-go/issues/1815 is done.

cretz:
## What was changed

* 💥 BREAKING CHANGE - `--api-key` now assumes TLS (i.e. `--tls`) by default, so users using API keys on non-TLS servers will have to explicitly disable TLS (i.e. `--tls=false`)
* Added experimental `temporal config` commands
* Changed internals to have env vars use envconfig approach
* No longer log when env var overrides config-file-set values as was in #293

## Checklist

1. Closes #751


temporalcli/client.go
```diff
@@ -12,47 +10,167 @@ import (
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed loading client config: %w", err)
+		}
+	}
+
+	// To support legacy TLS environment variables, if they are present, we will
+	// have them force-override anything loaded from existing file or env
+	if !cctx.RootCommand.DisableConfigEnv {
+		oldEnvTLSCert, _ := cctx.Options.EnvLookup.LookupEnv("TEMPORAL_TLS_CERT")
```
Sushisource, line 44: Mentioned in the other PR, but this sure seems like a good reason to export the constants from the SDK

temporalcli/commandsgen/commands.yml
```diff
@@ -143,11 +145,44 @@ commands:
...
+        description: |
+          File path to read TOML config from, defaults to
+          `$CONFIG_PATH/temporal/temporal.toml` where `$CONFIG_PATH` is defined
+          as `$HOME/.config` on Unix, "$HOME/Library/Application Support" on
+          macOS, and %AppData% on Windows.
+        experimental: true
+        implied-env: TEMPORAL_CONFIG_FILE
+      - name: profile
+        type: string
+        description: Profile to use for config file.
```
Sushisource, line 172: Should this use `default` as the default?

temporalcli/commandsgen/commands.yml
```diff
@@ -647,6 +682,119 @@ commands:
...
+      Config files are TOML files that contain profiles, with each profile
+      containing configuration for connecting to Temporal. 
+
+      ```
+      temporal config set \
+          --prop address \
+          --value us-west-2.aws.api.temporal.io:7233
+      ```
+
+      The default config file is at `$CONFIG_PATH/temporal/temporal.toml` where
```
Sushisource, line 697:
```suggestion
      The default config file path is `$CONFIG_PATH/temporal/temporal.toml` where
```

Since that file may not exist until a user makes it (unless we auto-write an empty one or something on first go, but I didn't see that)

temporalcli/commandsgen/commands.yml
```diff
@@ -647,6 +682,119 @@ commands:
...
+    summary: |
+      Delete a config file property or an entire config profile (EXPERIMENTAL)
+    description: |
+      Remove a full profile entirely _or_ remove a property within a profile.
+      When deleting an entire profile, the `--profile` must be set explicitly.
+
+      ```
+      temporal env delete \
+          --profile my-profile
+      ```
```
Sushisource, line 731: IMO adding an explicit `--delete-whole-profile` or something might be good in that instance, but, not hugely concerned.

temporalcli/commandsgen/commands.yml
cretz: No because the default is actually determined by the Go SDK not by the CLI. A CLI flag (whether set via "default" or explicitly) would override where it may be otherwise set, e.g. env var or the actual default.

temporalcli/commandsgen/commands.yml
cretz: We definitely don't auto-write any files. Will update text.

temporalcli/commandsgen/commands.yml
cretz: :+1: Yeah this just matches what we do for `env` today, but I do agree. Actually this should be a top-level command called `delete-profile` so `delete` can always require the `--prop`.

cretz: Moved to its own command. Leaving conversation open for others that may want to see that this changed.

temporalcli/client.go
```diff
@@ -12,47 +10,167 @@ import (
...
+	}
+	// In the past, the presence of API key CLI arg did not imply TLS like it
+	// does with envconfig. Therefore if there is a user-provided API key and
+	// TLS is not present, explicitly disable it so API key presence doesn't
+	// enable it in ToClientOptions below.
+	// TODO(cretz): Or do we want to break compatibility to have TLS defaulted
+	// for all API keys?
+	if c.ApiKey != "" && clientProfile.TLS == nil {
+		clientProfile.TLS = &envconfig.ClientConfigTLS{Disabled: true}
+	}
```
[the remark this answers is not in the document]
Sushisource: I think it makes sense

Sushisource (approved):

temporalcli/client.go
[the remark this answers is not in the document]
cretz: Ok, I have made this breaking change

temporalcli/commandsgen/commands.yml
```diff
@@ -143,11 +145,44 @@ commands:
...
           Defaults to `$HOME/.config/temporalio/temporal.yaml`.
+        # TODO(cretz): Deprecate when `config` GA
+        # deprecated: |
+        #   Use `config-file` instead. If an env file is present, it will take
+        #   precedent over `config` file or config environment variables.
+      - name: config-file
+        type: string
+        description: |
+          File path to read TOML config from, defaults to
+          `$CONFIG_PATH/temporal/temporal.toml` where `$CONFIG_PATH` is defined
```
cretz, line 165:
```suggestion
          `$CONFIG_PATH/temporalio/temporal.toml` where `$CONFIG_PATH` is defined
```

temporalcli/client.go
```diff
@@ -12,52 +10,175 @@ import (
...
+	if cctx.CurrentCommand != nil {
+		addressExplicitlySet = cctx.CurrentCommand.Flags().Changed("address")
+		namespaceExplicitlySet = cctx.CurrentCommand.Flags().Changed("namespace")
+	}
+	if addressExplicitlySet {
+		clientProfile.Address = c.Address
+	}
+	if namespaceExplicitlySet {
+		clientProfile.Namespace = c.Namespace
+	}
```
cretz, line 90:
```suggestion
	if namespaceExplicitlySet {
		clientProfile.Namespace = c.Namespace
	} else {
		c.Namespace = clientProfile.Namespace
	}
```

Some code is using `c.Namespace` flag in its output which is obviously not valid if it's overwritten by env config stuff. Need to either update `c.Namespace` with the valid value or have some other way to relay what is the now-current namespace value.

temporalcli/commandsgen/commands.yml
```diff
@@ -101,6 +101,8 @@
 #       required: Whether the option is required. (bool)
 #       short: The single letter short version of name (i.e. a for address). (string)
 #       env: Binds the environment variable to this flag. (string)
+#       implied-env: Documents the environment variable as bound to the flag,
+#         but does actually bind it.
```
yuandrew, line 105: Should this be `doesn't`? Also missing the type `(string)`

yuandrew: Also, I'm not seeing `ImpliedEnv` used anywhere, do we want it printed into the generated docs?

yuandrew (commented):
Primarily looked at the commandsgen stuff, just one clarity comment

temporalcli/commandsgen/commands.yml
cretz: Yes, I meant `doesn't`, will fix. Yes, I do think it'd make sense to be in the docs just like `env`. I just had to make a separate field because it's effectively doc only, it has no runtime use.

temporalcli/client.go
cretz: Fixed

temporalcli/commandsgen/commands.yml
cretz: Fixed

Issue #744, opened 2025-01-28
[Feature Request] Update Worker deployment commands to versioning-3.1


antlai-temporal:
## Description
The commands associated with the Worker Deployment API are changing to reflect versioning-3.1.

Issue #745, opened 2025-01-28
[Feature Request] Add Ramp to worker deployment


antlai-temporal:
### Is your feature request related to a problem? Please describe.

The new versioning-3.1 adds support for gradually ramping deployments that need to be supported in the CLI.

PR #765, opened 2025-02-26
Deployment API (versioning 3.1)

antlai-temporal:
## What was changed

This PR provide support for the Deployment API with the following commands:
```
temporal worker deployment delete
temporal worker deployment delete-version
temporal worker deployment describe
temporal worker deployment describe-version
temporal worker deployment list
temporal worker deployment set-current-version
temporal worker deployment set-ramping-version
temporal worker deployment update-metadata-version
```
It also updates the `workflow update-options` API to the new versioning override API, and `workflow describe`.

NOTE: the server hash in go.mod will be changed to v1.27.1 before merging.

## Checklist

1. Closes 
#744 #745


temporalcli/commandsgen/commands.yml
```diff
@@ -690,17 +690,40 @@ commands:
...
         - worker deployment describe
         - worker deployment list
-        - worker deployment get-current
-        - worker deployment set-current
+        - worker deployment delete
+        - worker deployment describe-version
+        - worker deployment set-current-version
+        - worker deployment set-ramping-version
+        - worker deployment delete-version
+        - worker deployment update-metadata-version
```
ShahabT, line 726: It should be `update-version-metadata`, right?

cretz (approved):
Approving with the understanding that everything added here is not required to be stable and can change in subsequent versions.

PR #771, opened 2025-02-27
Add fields from extended info to DescribeWorkflow output

ychebotarev:
## What was changed

Add extended info to the DescribeWorkflow output.
## Why?

Customer request.

PR #773, opened 2025-03-07
Fix for DescribeWorkflow

ychebotarev:
## What was changed

Check if extended info is not nil.

## Why?

https://github.com/temporalio/cli/pull/771
Code assumes that some extended info exists in proto. Which is not true for older server versions..

### DIFF
--- a/temporalcli/commands.workflow_view.go
+++ b/temporalcli/commands.workflow_view.go
@@ -102,48 +102,55 @@ func (c *TemporalWorkflowDescribeCommand) run(cctx *CommandContext, args []strin
 
 	cctx.Printer.Println(color.MagentaString("Execution Info:"))
 	info := resp.WorkflowExecutionInfo
-	extendedInfo := resp.WorkflowExtendedInfo
 	_ = cctx.Printer.PrintStructured(struct {
-		WorkflowId              string
-		RunId                   string
-		Type                    string
-		Namespace               string
-		TaskQueue               string
-		AssignedBuildId         string
-		StartTime               time.Time
-		CloseTime               time.Time                  `cli:",cardOmitEmpty"`
-		ExecutionTime           time.Time                  `cli:",cardOmitEmpty"`
-		Memo                    map[string]*common.Payload `cli:",cardOmitEmpty"`
-		SearchAttributes        map[string]*common.Payload `cli:",cardOmitEmpty"`
-		StateTransitionCount    int64
-		HistoryLength           int64
-		HistorySize             int64
-		ExecutionExpirationTime time.Time `cli:",cardOmitEmpty"`
-		RunExpirationTime       time.Time `cli:",cardOmitEmpty"`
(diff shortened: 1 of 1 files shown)


temporalcli/commands.workflow_view.go
```diff
@@ -102,49 +102,56 @@ func (c *TemporalWorkflowDescribeCommand) run(cctx *CommandContext, args []strin
...
+		Memo:                 info.Memo.GetFields(),
+		SearchAttributes:     info.SearchAttributes.GetIndexedFields(),
+		StateTransitionCount: info.StateTransitionCount,
+		HistoryLength:        info.HistoryLength,
+		HistorySize:          info.HistorySizeBytes,
 	}, printer.StructuredOptions{})
 
+	extendedInfo := resp.WorkflowExtendedInfo
+	if extendedInfo != nil {
+		cctx.Printer.Println(color.MagentaString("Extended Execution Info:"))
```
cretz, line 139: I think this can remain in the original structure and just not be omitted if empty. Just because we call this extended info in the protos doesn't mean it has to be broken out for users. Can just leave old code and do `resp.GetWorkflowExtendedInfo().GetWhatever()`. I understand that means the absence of "cancel requested" can either mean it wasn't requested _or_ it's too old of a server, but I think that's ok.

ychebotarev: resp.GetWorkflowExtendedInfo().GetCancelRequested() will return false if extended info doesn't exist.
Even if cancel was requested. This is misleading, I rather provide no information then false information.

ychebotarev: I may remove this:
`cctx.Printer.Println(color.MagentaString("Extended Execution Info:"))`

cretz: > Even if cancel was requested. This is misleading, I rather provide no information then false information.

I mentioned in my comment I don't think this is that misleading. But if the concern is strong enough, you can make a new field that has three states (e.g. maybe a `*bool`). But ideally all these fields are in the same card field structure and aligned the same with the other fields.

ychebotarev: Yes, you did mention it. If it shows "false" while it is "true" it is still misleading. 
Lots of information in this output is in separate card, like: `if info.VersioningInfo != nil {`, `if running {`, etc.
So I see no harm is clearly separating it. Also we did it for a reason - extended info is a workflow related information that is not available in visibility. I don't see much hard if any of putting it into separate card.

cretz: If we were fine having it together before this PR I figure the fix doesn't really need to add that feature

cretz: EDIT: Marking approved, I am not too concerned with this, we can change later since there is no compatibility requirement for text output

cretz (approved):

PR #783, opened 2025-03-25
Update server to v1.28.0

antlai-temporal:
## What was changed

This PR updates `next-server` to catch-up with `main` changes and then updates server (and api)
to the latest (v1.28.0-130.0 and v1.46.0 respectively)


Sushisource (approved):
Many thanks

PR #784, opened 2025-03-26
Pin modernc/sqlite version 1.34.1

### LINKED ISSUE #777: [Feature Request] Pin the sqlite driver version due to sqlite driver issue

There is an issue that was solved but reappeared in the sqlite driver. We have an issue opened at https://gitlab.com/cznic/sqlite/-/issues/196, but in the meantime we need to do [like server does](https://github.com/temporalio/temporal/blob/1ec4badd2b66846cc8e7e1b7d2302ad82e6d9345/Makefile#L121) and pin the sqlite driver version to a specific version in `go.mod` (probably need to have some kind of `import _ "modernc.org/sqlite"` to keep it in `go.mod` properly).

Bonus points if we can add a test like in https://github.com/temporalio/temporal/pull/6815 so we can catch it if it regresses again.

yuandrew:
## What was changed

Pinned v1.34.1 for modernc/sqlite

## Why?

1.34.2 has a regression, waiting on https://gitlab.com/cznic/sqlite/-/issues/196 to be resolved.
See https://github.com/temporalio/temporal/pull/7333 for more details

## Checklist

1. Closes #777

2. How was this tested:

3. Any docs updates needed?


Makefile
```diff
@@ -9,3 +9,15 @@ temporalcli/commands.gen.go: temporalcli/commandsgen/commands.yml
...
+# Pinning modernc.org/sqlite to this version until https://gitlab.com/cznic/sqlite/-/issues/196 is resolved.
+PINNED_DEPENDENCIES := \
+	modernc.org/sqlite@v1.34.1 \
+	modernc.org/libc@v1.55.3
+
+
+update-dependencies:
+	@printf $(COLOR) "Update dependencies..."
+	@go get -u -t $(PINNED_DEPENDENCIES) ./...
+	@go mod tidy
```
cretz, line 23: Hrmm, not really sure this is that necessary or that this is how contributors will update dependencies. Is there any way for us to prevent a contributor from using `go get -u` themselves to update the dependency past this version? Were you able to find a test in the server that proves the concerning bug is solved? How does the server prevent a contributor from doing `go get -u` wiping away this pinned dependency? Worst case if we can't find a test that checks the bug does not exist, can we maybe add a test that uses https://pkg.go.dev/runtime/debug#ReadBuildInfo (or some other approach) to check the expected pinned versions until upstream is fixed?

cretz:
Concerning CI failures, I would suggest changing `fail-fast: true` to `fail-fast: false` in `ci.yml` so that other jobs continue to run even if one fails. (won't help solve a CI bug, but at least lets you know whether it's unique/flaky to that one environment)

Makefile
yuandrew: > Were you able to find a test in the server that proves the concerning bug is solved?

Prathyush said some of their functional tests would fail, but not that commonly. 

Added a test that just manually checks the go.mod file for the versions of these 2 deps that we expect (ReadBuildInfo didn't contain the information we need)

temporalcli/sqlite_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Pinning modernc.org/sqlite to this version until https://gitlab.com/cznic/sqlite/-/issues/196 is resolved
+func TestSqliteVersion(t *testing.T) {
+	content, err := os.ReadFile("../go.mod")
+	if err != nil {
+		t.Fatalf("Failed to read go.mod: %v", err)
+	}
+	contentStr := string(content)
+	fmt.Println("content", contentStr)
```
cretz, line 18:
```suggestion
```

Makefile
cretz: Works for me, thanks!

cretz (approved):
LGTM. I have seen the same nil issue multiple times now for macos intel for `TestWorkflow_Batch_Update_Options_Versioning_Override` test and it's even occurring on main, so something is happening there that we may have to fix separately. I will bring up the CI failure internally.

PR #785, opened 2025-03-28
Use `require` instead of `assert` in test

antlai-temporal:
## What was changed

One of the versioning tests incorrectly used assert instead of require, and that did not 
protect from a nil pointer panic.


temporalcli/commands.workflow_test.go
```diff
@@ -524,15 +525,15 @@ func (s *SharedServerSuite) TestWorkflow_Batch_Update_Options_Versioning_Overrid
 				"-w", run.GetID(),
 				"--output", "json",
 			)
-			assert.NoError(t, res.Err)
+			require.NoError(t, res.Err)
```
cretz, line 528: A little concerned with the above `temporal workflow update-options --query whatever` above in that visibility is eventually consistent, so how do you know it found anything with that `CustomKeywordField`? Or are you just relying on the speed of dev server visibility that the workflow will always be in visibility by this time?

cretz (approved):

temporalcli/commands.workflow_test.go
antlai-temporal: Adding extra test for visibility

PR #791, opened 2025-04-16
Match tags in documentation repo

yuandrew:
## What was changed

Added a tags section to match what Documentation side has

## Why?

We previously just used keywords, but that could be seen as keyword stuffing. The tags were updated on the docs side a few months ago, changes were never updated on this side.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
Nothing blocking

PR #796, opened 2025-04-29
Auto generate and publish CLI docs on each release

yuandrew:
## What was changed

Created a new Github Action that auto-generates and publishes a PR to the [documentation repo](https://github.com/temporalio/documentation) with the latest generated CLI docs. Corresponding docs side PR is already merged, https://github.com/temporalio/documentation/pull/3525, and first PR of the auto-generated docs is merged in https://github.com/temporalio/documentation/pull/3528.

## Why?

Makes keeping CLI docs up to date significantly easier.

## Checklist

1. Closes 

2. How was this tested:

Tested in this PR, but there's a small chance I missed something in the actual release trigger workflow. No way to fully test that until a new release is cut after this action is on main.

3. Any docs updates needed?


.github/workflows/trigger-docs.yml
```diff
@@ -0,0 +1,52 @@
...
+          app-id: ${{ secrets.TEMPORAL_CICD_APP_ID }}
+          private-key: ${{ secrets.TEMPORAL_CICD_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: documentation # generate a token with permissions to trigger GHA in documentation repo
+
+      - name: Trigger Documentation Workflow
+        env:
+          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          gh workflow run update-cli-docs.yml \
```
cretz, line 46: Can you link to the PR with this workflow in it on the docs side? Just curious. I am happy to see we are delegating the generation to that workflow instead of trying to do it here and shipping data.

cretz (approved):
Nothing blocking

.github/workflows/trigger-docs.yml
yuandrew: Ahh yes, https://github.com/temporalio/documentation/blob/main/.github/workflows/update-cli-docs.yml. Will add to the PR desc as well

yuandrew: oh sorry you asked for the PR, that should already be in the PR desc, https://github.com/temporalio/documentation/pull/3525

cretz: Thanks!

PR #808, opened 2025-06-02
Add new options to reset/update activities

ychebotarev:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commandsgen/commands.yml
+++ b/temporalcli/commandsgen/commands.yml
@@ -379,6 +379,14 @@ commands:
       - name: identity
         type: string
         description: Identity of the user submitting this request.
+      - name: restore_original
+        type: bool
+        description: |
+          If set, the activity options will be restored to the default.
+          Default options are then options activity was created with.
+          They are part of the first SCHEDULE event.
+          This flag cannot be combined with any other option; if you supply
+          restore_original together with other options, the request will be rejected.
     option-sets:
       - workflow reference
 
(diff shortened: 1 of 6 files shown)


temporalcli/commandsgen/commands.yml
```diff
@@ -409,11 +417,11 @@ commands:
       ```
     options:
       - name: activity-id
-        short: a
+        short: i
         type: string
         description: Activity ID to pause.
       - name: activity-type
-        short: g
+        short: t
```
cretz, line 425: Do these same changes need to be made to other commands like "unpause" or "reset" for consistency?

cretz (commented):
Only minor PR comment, but will want to wait to merge until tests are available (i.e. when next-server can get a new tagged release)

temporalcli/commandsgen/commands.yml
ychebotarev: yes. done.

ychebotarev:
> Only minor PR comment, but will want to wait to merge until tests are available (i.e. when next-server can get a new tagged release)

Note.
I updated server version to main, because apparently we don't have "tag" for new server release.

ychebotarev:
> 👍 This is good so that we can add tests. 

tests are added.

> I think we should wait to merge until we can update go.mod with a server tag. If figure this will be quite soon anyways.

This is merge to `next-server`. Why wait?

Issue #804, opened 2025-05-14
[CLI] Update protos, use buildid/deployment name args & support versioning overrides when starting workflows


Sushisource:
### Is your feature request related to a problem? Please describe.

Need to support the versioning override field of workflow starting APIs. See SDK implementations for reference.

### Describe the solution you'd like

Probably just a flag to the relevant commands

PR #811, opened 2025-06-09
Versioning refactor

Sushisource:
## What was changed
Refactor versioning to use latest SDK & avoid string-based deployment versions

## Why?
Consistent with SDK refactorings

## Checklist

1. Closes https://github.com/temporalio/cli/issues/804

2. How was this tested:
Existing tests

3. Any docs updates needed?


go.mod
```diff
@@ -17,9 +17,10 @@ require (
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.36.0
-	go.temporal.io/api v1.49.0
-	go.temporal.io/sdk v1.34.0
-	go.temporal.io/server v1.28.0-133.1
+	go.temporal.io/api v1.50.0
+	// Set to released version once available
+	go.temporal.io/sdk v1.34.1-0.20250609225810-918fea843587
```
Sushisource: We'll need to use the newest SDK release before merging next-server into main

temporalcli/commands.workflow_test.go
```diff
@@ -541,18 +551,26 @@ func (s *SharedServerSuite) TestWorkflow_Batch_Update_Options_Versioning_Overrid
 
 			versioningInfo := jsonResp.GetWorkflowExecutionInfo().GetVersioningInfo()
 			require.NotNil(t, versioningInfo)
-			require.NotNil(t, versioningInfo.VersioningOverride)
-			require.Equal(t, version2, versioningInfo.VersioningOverride.PinnedVersion)
+			// TODO: Unclear why this just is not showing up in the response
+			// require.NotNil(t, versioningInfo.VersioningOverride)
+			// TODO: Fix
+			// require.Equal(t, version2, versioningInfo.VersioningOverride.PinnedVersion)
```
Sushisource: VersioningOverride is just not showing up in the response any more for some reason. Any idea here @ShahabT @carlydf ? Specifically this is for the test which batch-updates workflows. Seems like the override just might not be getting applied when run as a batch?

go.mod
cretz: Usually we don't even want unreleased dependencies in next-server. Can this PR wait to be merged until the SDK has been released?

cretz (commented):
LGTM, but would rather not merge even to `next-server` until dependency tagged unless there is something urgent requiring it be usable in CLI before even usable in SDK.

temporalcli/commandsgen/commands.yml
```diff
@@ -3238,9 +3221,14 @@ commands:
...
           - auto_upgrade
-      - name: versioning-override-pinned-version
+      - name: versioning-override-deployment-name
         type: string
-        description: Override Pinned Version for a Worker Deployment (Only for pinned).
+        description: Deployment name component of override Pinned Version for a Worker Deployment
+          (Only for pinned).
+      - name: versioning-override-build-id
+        type: string
+        description: Build Id component of override Pinned Version for a Worker Deployment
```
drewhoskins-temporal: This description could use a de-mangling.  
How about
Override the build ID to move your running Workflow to another build within its Worker Deployment.  The workflow must already be pinned or you must also pass `--versioning-override-behavior pinned`

temporalcli/commandsgen/commands.yml
```diff
@@ -3238,9 +3221,14 @@ commands:
           - unspecified
           - pinned
           - auto_upgrade
-      - name: versioning-override-pinned-version
+      - name: versioning-override-deployment-name
```
drewhoskins-temporal, line 3224: Suggest: Override when you want to move your running Workflow to a different Worker Deployment.  The workflow must already be pinned or you must also pass `--versioning-override-behavior pinned`

drewhoskins-temporal: Is it allowable to just pass in the build ID without this one if they're not changing deployments?  (I hope so.)  I didn't see a test for that case but forgive me if I missed it.

temporalcli/commandsgen/commands.yml
Sushisource: Well, that isn't quite right. You can only use this option if you pass pinned. I'll rephrase it though.

temporalcli/commandsgen/commands.yml
Sushisource: ~I don't know if server will accept that. I can try.~

I added a test for this. Server accepts it, but the override doesn't include the deployment name any more afterward (it does include the non-override deployment name, which is the same, so maybe that's fine).

temporalcli/commands.workflow_test.go
Sushisource: Carly told me this is a bug they've discovered in server

Sushisource: Can confirm this is fixed now

temporalcli/commandsgen/commands.yml
```diff
@@ -3211,8 +3194,8 @@ commands:
       temporal workflow update-options \
           --workflow-id YourWorkflowId \
           --versioning-override-behavior pinned \
-          --versioning-override-pinned-version \
-            YourDeploymentSeriesName.YourDeploymentBuildId
+          --versioning-override-deployment-name YourDeploymentSeriesName \
```
carlydf:
```suggestion
          --versioning-override-deployment-name YourDeploymentName \
```

cretz (approved):

PR #812, opened 2025-06-12
Reset subcommand to update version

gow:
## What was changed
Added reset subcommands so that we can do reset + update version info atomically in one operation. The server changes were made in https://github.com/temporalio/temporal/pull/7719. This is the CLI counterpart.

To allow subcommands we added a new flag `subcommands-optional`. When this flag is present a parent command can be run in isolation without any subcommands (i.e subcommands will be optional)

## Why?
So that we can reset + update_version at the same time.

## Checklist

1. Closes:  N/A

2. How was this tested:
Added tests + manually tested

3. Any docs updates needed?
N/A


.github/CODEOWNERS
cretz, line 1: Do we need a separate PR to backport `main` into `next-server`? It is a bit hard to follow this PR on what changes are yours and what changes are in `main`.

gow: Yeah. Looks like `next-server` is a lot of commits behind main?
I don't fully understand the purpose of next-server branch. I just set it as the base to get the failing tests pass. I've changed the base back to main so that the PR shows just my commits.

Sushisource (approved):
This all looks good to me, though I'm not sure what's the problem with the test.

Regardless, I'll need to update this slightly after my refactor & next-server are merged. So potentially we can just let this sit until after that since everything needs to be together in the next release anyway.

go.mod
```diff
@@ -16,10 +18,10 @@ require (
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.36.0
-	go.temporal.io/api v1.46.0
-	go.temporal.io/sdk v1.33.0
+	go.temporal.io/api v1.49.2-0.20250605210325-e57aa581185c
+	go.temporal.io/sdk v1.34.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
-	go.temporal.io/server v1.27.1
+	go.temporal.io/server v1.29.0-135.2
```
cretz: Will have to wait for stable API and OSS server tags since this now targets `main`

temporalcli/commandsgen/commands.yml
```diff
@@ -51,7 +51,8 @@
 #   * When commands have a single command-level option, include it the mandatory example.
 #   * Use square bracket overviews to present how complex commands will be used.
 #     * Yes: temporal operator [command] [subcommand] [options]
-#     Commands with subcommands can't be run on their own.
+#     Commands with subcommands can't be run on their own unless
+#     subcommands-optional is set to true.
```
cretz, line 55: When designing commands, we had intentionally decided that a command either houses children or is a standalone command. We intentionally avoided hybrids in all of our designs so that it's clear when you execute a parent command you don't have a side effect. It's also clearer from a docs/usability standpoint to know that commands cannot serve two purposes.

I'm not completely against the change in our long-held stance, but it is a change and worth confirming the CLI is ok deviating from its existing stance.

temporalcli/commandsgen/commands.yml
```diff
@@ -3521,6 +3524,14 @@ commands:
           Don't prompt to confirm.
           Only allowed when `--query` is present.
 
+  - name: temporal workflow reset with-workflow-update-options
```
cretz, line 3512: I think if this is a separate command, it's `reset-and-update-workflow-options` but is it really a separate command or is it just a "post reset operation" on top of reset?

temporalcli/commandsgen/commands.yml
Sushisource: We agreed on this decision quite some ways back when this went through design review: https://www.notion.so/temporalio/Multi-commands-with-reset-PRD-1d38fc5677388025a756ca7241453e32

cretz: Ah, I wasn't aware the matter was settled, ok

.github/CODEOWNERS
cretz: > I don't fully understand the purpose of next-server branch

It's for server features without a stable server tag

cretz (commented):
LGTM, but will have to wait until there is a stable tag for API and server before merging in to `main`

temporalcli/commandsgen/commands.yml
gow: This is not a separate command. It's a "subcommand" of `workflow reset` in cli. In server rpc it's just a post reset operation param to reset request.

gow:
@cretz @Sushisource Please take a final look? I'll merge it on Monday evening.

temporalcli/commandsgen/commands.yml
```diff
@@ -4492,3 +4503,19 @@ option-sets:
...
+        type: string-enum
+        description: |
+          Override the versioning behavior of a Workflow.
+        required: true
+        enum-values:
+          - pinned
+          - auto_upgrade
+      - name: versioning-override-pinned-version
+        type: string
+        description: Override Pinned Version for a Worker Deployment (Only for pinned).
```
Sushisource: This will need to change to match my PR https://github.com/temporalio/cli/pull/811 after it goes in, so you can do that now if you want or I can do it before we cut the release, either way is fine.

Sushisource:
Looking good to me. Thanks!

PR #819, opened 2025-07-07
Update Server to v1.28

Sushisource:
Merges the next server branch


gow:
> May want to ask @gow if we want #812 in `main` too

My changes are against main already. I'll go ahead and merge that one.

PR #824, opened 2025-07-08
fix: activity pause,unpause,reset accept activity type OR activity id

spkane31:
## What was changed
`temporal activity {pause,unpause,reset}` commands take either an `--activity-id` or `activity-type` argument but not both. If both are provided the command will fail

## Why?
We want to fail when both are provided, the user doesn't need to provide activity type and id as type can be inferred from type.

## Checklist

1. Closes no issue

2. How was this tested:
* new unit test showing failure

3. Any docs updates needed?
documentation updates will be needed, @drewhoskins-temporal  or @ychebotarev can share where and I will update

### DIFF
--- a/temporalcli/commands.activity.go
+++ b/temporalcli/commands.activity.go
@@ -208,6 +208,10 @@ func (c *TemporalActivityPauseCommand) run(cctx *CommandContext, args []string) 
 		Identity: c.Identity,
 	}
 
+	if c.ActivityType != "" && c.ActivityId != "" {
+		return fmt.Errorf("either Activity Type or Activity Id, but not both")
+	}
+
 	if c.ActivityType != "" {
 		request.Activity = &workflowservice.PauseActivityRequest_Type{Type: c.ActivityType}
 	} else if c.ActivityId != "" {
@@ -259,6 +263,10 @@ func (c *TemporalActivityUnpauseCommand) run(cctx *CommandContext, args []string
 			ResetHeartbeat: c.ResetHeartbeats,
 			Jitter:         durationpb.New(c.Jitter.Duration()),
 			Identity:       c.Identity,
+		}
+
+		if c.ActivityType != "" && c.ActivityId != "" {
+			return fmt.Errorf("either Activity Type or Activity Id, but not both")
 		}
 
 		if c.ActivityType != "" {
@@ -316,6 +324,10 @@ func (c *TemporalActivityResetCommand) run(cctx *CommandContext, args []string) 
 		Identity:       c.Identity,
 		KeepPaused:     c.KeepPaused,
 		ResetHeartbeat: c.ResetHeartbeats,
+	}
+
+	if c.ActivityType != "" && c.ActivityId != "" {
+		return fmt.Errorf("either Activity Type or Activity Id, but not both")
 	}
 
 	if c.ActivityType != "" {
--- a/temporalcli/commands.activity.go
+++ b/temporalcli/commands.activity.go
@@ -208,16 +208,15 @@ func (c *TemporalActivityPauseCommand) run(cctx *CommandContext, args []string) 
 		Identity: c.Identity,
 	}
 
-	if c.ActivityType != "" && c.ActivityId != "" {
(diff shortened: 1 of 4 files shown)


temporalcli/commands.activity.go
```diff
@@ -208,6 +208,10 @@ func (c *TemporalActivityPauseCommand) run(cctx *CommandContext, args []string)
 		Identity: c.Identity,
 	}
 
+	if c.ActivityType != "" && c.ActivityId != "" {
```
cretz: Can also just toss a `if c.ActivityId != ""` inside the `if c.ActivityType != ""` body below, but this is fine too

cretz (approved):

temporalcli/commands.activity.go
ychebotarev: this is "3 times" repetition, usually worth to extract into a single function.
I would introduce something like  "activity-reference" (see commands.yml, "single-workflow-or-batch" as example), and add "Validate" to it.

temporalcli/commands.gen.go
```diff
@@ -473,10 +475,35 @@ func NewTemporalActivityFailCommand(cctx *CommandContext, parent *TemporalActivi
 	return &s
 }
 
+type ActivityReferenceOptions struct {
+	ActivityId   string
+	ActivityType string
+}
+
+func (v *ActivityReferenceOptions) buildFlags(_ *CommandContext, f *pflag.FlagSet, activityType string, onlyWithoutQuery bool) {
```
cretz, line 483: You can't add code to this code-generated file, it'd get removed on regen. Rather, add in another file (or maybe better, do not add methods to generated structures beyond the `run` method needed on command, but make simple utility functions that accept them).

temporalcli/commands.gen.go
```diff
@@ -473,10 +475,35 @@ func NewTemporalActivityFailCommand(cctx *CommandContext, parent *TemporalActivi
...
+	f.StringVarP(&v.ActivityId, "activity-id", "a", "", fmt.Sprintf("Activity ID to %s.", activityType))
+	if onlyWithoutQuery {
+		f.StringVarP(&v.ActivityType, "activity-type", "g", "", fmt.Sprintf("Activity Type to %s. Can only be used without --query.", activityType))
+	} else {
+		f.StringVarP(&v.ActivityType, "activity-type", "g", "", fmt.Sprintf("Activity Type to %s.", activityType))
+	}
+}
+
+func (v *ActivityReferenceOptions) validateFlags() error {
```
cretz: You can't add code to this code-generated file, it'd get removed on regen. Rather, add in another file (or maybe better, do not add methods to generated structures beyond the run method needed on command, but make simple utility functions that accept them).

temporalcli/commands.gen.go
cretz: You cannot customize flag building in this way with our framework as it is today. You can make different option sets or, as we have done in other cases, accept that documentation is shared when option sets are shared.

temporalcli/commands.gen.go
spkane31: Removed the `ActivityReferenceOptions` struct entirely, opting for the repetitive checking. The flags for this `activity-id` and `activity-type` are already set with other option-sets which causes a panic and unraveling that doesn't seem like the correct solution here.  

cc @ychebotarev

temporalcli/commands.activity.go
```diff
@@ -208,6 +208,10 @@ func (c *TemporalActivityPauseCommand) run(cctx *CommandContext, args []string)
 		Identity: c.Identity,
 	}
 
+	if c.ActivityId != "" && c.ActivityType != "" {
```
ychebotarev, line 211: back to square 1 :) ?
somewhere
```
func ValidateActivityReference(activityId, activityType) error {
   ...
}
```
and then here and below
```
if err = ValidateActivityReference(c.ActivityId, c.ActivityType); err != nil {
   return err
}
```
up to you

ychebotarev (approved):

Issue #316, opened 2023-08-24
[Feature Request] Publish Docker image


cretz:
### Describe the solution you'd like
Need a docker image for this single binary

slamdev:
while waiting for the official image you can use this one: https://github.com/slamdev/temporalite-container-image
```yaml
version: "3.8"
services:
  temporal:
    # https://hub.docker.com/r/slamdev/temporalite/tags
    image: slamdev/temporalite:0.10.7
    restart: on-failure
    command:
      - --namespace=default
      - --db-filename=/tmp/db
    tmpfs:
      - /tmp
    ports:
      - "7233:7233"
      - "8233:8233"
```
ignore `temporalite` name, its temporal cli under the hood, I just re-purposed the repo that was used to build deprecated temporalite.

ElanHasson:
@cretz Can i just send a PR for this?

I was thinking:

Add a GHA workflow to build, test, sign the image and push to ghcr on every release?

cretz:
We are actually in the process of rewriting the innards of the CLI of which #463 is part, yes. We just don't have the publish step yet or the rest of the rewrite completed, stay tuned.

ElanHasson:
Love it thanks!

ElanHasson:
so, `temporalio/admin-tools` has the cli as well :)

enricoschaaf:
Do I see this right that the is no replacement yet for https://github.com/slamdev/temporalite-container-image? Since the PR is merged I assumed it maybe there is but it seems the action just triggers docker builds of other containers that contain the cli but no standalone image.

slamdev:
> Do I see this right that the is no replacement yet for https://github.com/slamdev/temporalite-container-image? Since the PR is merged I assumed it maybe there is but it seems the action just triggers docker builds of other containers that contain the cli but no standalone image.

you can switch to official image now, that's how I use it:
```yaml
  temporal:
    # https://hub.docker.com/r/temporalio/server/tags
    image: temporalio/server:1.21.5.0
    restart: on-failure
    user: "0:0" # running as root to avoid permission issues with writing to /tmp/sqlite/db
    entrypoint:
      - temporal
      - server
      - start-dev
      - --namespace=default
      - --db-filename=/tmp/sqlite/db
      - --ip=0.0.0.0
    tmpfs:
      - /tmp/sqlite
    ports:
      - "7233:7233"
      - "8233:8233"
```

joelmarty:
[...]

PR #826, opened 2025-07-09
Updated Dockerfile in preparation for public Docker image

maciejdudko:
## What was changed

Updated Dockerfile to be based on Alpine and to package binaries produced by Goreleaser. Added instructions to CONTRIBUTORS.md on how to build Docker image.

Note that this PR does NOT contain any automation for Docker image publishing. Release 1.4.0 will be published manually after this PR is merged.

## Why?

Feature request: https://github.com/temporalio/cli/issues/316

## Checklist

2. How was this tested:

    goreleaser build --snapshot --clean
    docker build --tag temporalio/temporal:snapshot --platform=linux/amd64,linux/arm64 .

3. Any docs updates needed?
Internal release instructions will need to be updated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,9 @@ COPY ./dist/nix_linux_arm64/temporal /dist/arm64/temporal
 
 FROM alpine:3.22
 ARG TARGETARCH
+RUN apk add --no-cache ca-certificates
 COPY --from=dist /dist/$TARGETARCH/temporal /usr/local/bin/temporal
+RUN adduser -u 1000 -D temporal
+USER temporal
 
 ENTRYPOINT ["temporal"]


Dockerfile
```diff
@@ -1,21 +1,9 @@
...
-COPY . ./
-
-# Build
-RUN go build ./cmd/temporal
-
-# Use slim container for running
-FROM debian:bookworm-slim
-RUN set -x && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    ca-certificates && \
-    rm -rf /var/lib/apt/lists/*
```
bergundy, line 15: Can you confirm that this container has the alpine equivalent of `ca-certificates`?

Dockerfile
```diff
@@ -1,21 +1,9 @@
-FROM golang:1.24-bookworm AS builder
+FROM --platform=$BUILDARCH scratch AS dist
+COPY ./dist/nix_linux_amd64_v1/temporal /dist/amd64/temporal
+COPY ./dist/nix_linux_arm64/temporal /dist/arm64/temporal
 
-WORKDIR /app
+FROM alpine:3.22
+ARG TARGETARCH
+COPY --from=dist /dist/$TARGETARCH/temporal /usr/local/bin/temporal
```
bergundy: Can't we do this in one step? Why do we need a separate `dist`step?

Dockerfile
```diff
@@ -1,21 +1,9 @@
-FROM golang:1.24-bookworm AS builder
+FROM --platform=$BUILDARCH scratch AS dist
+COPY ./dist/nix_linux_amd64_v1/temporal /dist/amd64/temporal
+COPY ./dist/nix_linux_arm64/temporal /dist/arm64/temporal
 
-WORKDIR /app
+FROM alpine:3.22
```
bergundy, line 5: Doesn't this also need a `--platform` flag?

cretz, line 5: So we don't test against this OS very well FWIW, though Go is self-contained and doesn't use libc/musl so it should be fine. But just curious why alpine instead of e.g. `debian:bookworm-slim` or maybe even `gcr.io/distroless/static-debian12` or others often seen in the wild? It seems the latter (https://github.com/GoogleContainerTools/distroless) is quite common for this use case.

CONTRIBUTING.md
```diff
@@ -46,3 +46,18 @@ Note that inclusion of space characters in the value supplied via `-ldflags` is
 Here's an example that adds branch info from a local repo to the version string, and includes a space character:
 
     go build -ldflags "-X 'github.com/temporalio/cli/temporalcli.buildInfo=ServerBranch $(git -C ../temporal rev-parse --abbrev-ref HEAD)'" -o temporal ./cmd/temporal/main.go
+
+## Building Docker image
+
+Docker image build requires [Goreleaser](https://goreleaser.com/) to build the binaries first, although it doesn't use
+Goreleaser for the Docker image itself.
```
cretz, line 53: > although it doesn't use Goreleaser for the Docker image itself.

So was looking at https://goreleaser.com/customization/docker/ and it seems they won't publish for you, is that correct? If so, it makes sense we don't want to rely on it.

Dockerfile
maciejdudko: The `dist` step acts as a manual per-architecture mapping to the right binary. Docker doesn't support conditional `COPY` so due to the differences in Go and Docker target naming, it would be complicated to do in a single step.

Dockerfile
maciejdudko: `ca-certificates` exists in Alpine repo but is missing in the image. I will push a new revision that has this fixed.

Dockerfile
maciejdudko: No. `--platform` defaults to `$TARGETPLATFORM`, which is what we want here. `dist` uses `$BUILDPLATFORM` so it can be potentially cached for multiplatform builds.

CONTRIBUTING.md
maciejdudko: It's not documented but it can in fact push to Dockerhub. But since it's undocumented I don't want to rely on it. The bigger reason is that it can't build a true multiplatform image, it needs to build them separately for each platform and then combine them with a manifest, and it can only do one tag at a time it seems (and we need two - the version and the `latest`). It too is very poorly documented, and also we're using Goreleaser v1 which they don't publish docs for anymore (I had to use Web Archive). Overall, it feels very not worth it.

Dockerfile
maciejdudko: I was basing the image on our `admin-tools` image which uses Alpine. I'm not very well versed in Docker conventions, but it works, and Alpine is only 4MB compared to `debian:bookworm-slim`'s 28MB. But I don't have a strong opinion here.

cretz: :+1: Curious about the distroless one since it's fairly common these days (e.g. it's what Docker uses in their guide sample at https://docs.docker.com/guides/golang/build-images/#multi-stage-builds). But Alpine is fine too. Not a blocker.

cretz (approved):

Dockerfile
bergundy: I was thinking of recommending distroless too here. Alpine is a bit nicer if you ever want a shell to inspect the image, we may want that if we add support for config files later. Agree it's not a blocker.

bergundy: I see.

Dockerfile
bergundy: Thanks.

Dockerfile
bergundy: Not sure I follow, can't this step be `COPY ./dist/$TARGETARCH/temporal /usr/local/bin/temporal`?

maciejdudko: It won't work because the directory name is in the wrong format (arch is `amd64_v1` in Goreleaser but `amd64` in Docker).

bergundy: So use a build arg to say where to copy from instead?

maciejdudko: It wouldn't work for building multiplatform image with a single `docker build` command as all platforms would receive the same argument. I guess we could have a Docker bake file or a shell script wrapper or something but I think we should wait with that kind of stuff until we have proper release automation so that we don't do any throwaway work, and keep this PR as simple as possible. Especially since the extra step doesn't affect the final container image.

CONTRIBUTING.md
robross0606: Wondering why, with this conversation seemingly open-ended, the MR was merged.

PR #827, opened 2025-07-10
Add "Autogenerated" comment to each generated CLI docs page

yuandrew:
## What was changed

Added a comment at the top of each generated .mdx file that the file is auto-generated

## Why?

Make it more obvious that source of truth for CLI docs lives in this repo

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):
Might want to make it take up even more lines so it's extra obvious

PR #828, opened 2025-07-11
Added Docker instructions to README

maciejdudko:
Added information how to use official CLI Docker image to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,17 @@ Reference [the documentation](https://docs.temporal.io/cli) for detailed install
 2. Extract the downloaded archive.
 3. Add the `temporal` binary to your `PATH` (`temporal.exe` for Windows).
 
+### Run via Docker
+
+[Temporal CLI on DockerHub](https://hub.docker.com/r/temporalio/temporal)
+
+    docker run --rm temporalio/temporal
+
+Note that for dev server to be accessible from host system, it needs to listen on external IP and the ports need to be forwarded:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,17 @@ Reference [the documentation](https://docs.temporal.io/cli) for detailed install
 2. Extract the downloaded archive.
 3. Add the `temporal` binary to your `PATH` (`temporal.exe` for Windows).
 
+### Run via Docker
+
+[Temporal CLI on DockerHub](https://hub.docker.com/r/temporalio/temporal)
+
+    docker run --rm temporalio/temporal
```
dandavison: As written this command fails with a non-zero exit code and "unknown command".

```suggestion
    docker run --rm temporalio/temporal --help
```

README.md
```diff
@@ -23,6 +23,17 @@ Reference [the documentation](https://docs.temporal.io/cli) for detailed install
...
+
+Note that for dev server to be accessible from host system, it needs to listen on external IP and the ports need to be forwarded:
+
+    docker run --rm -p 7233:7233 -p 8233:8233 temporalio/temporal:latest server start-dev --ip 0.0.0.0
```
dandavison, line 34: Can we make `--ip 0.0.0.0` not be required for `start-dev` when it's running inside docker?

maciejdudko: Right now the default IP is hardcoded to 127.0.0.1, we'd have to introduce a new env var or a config file in the command's code to allow changing the default within the container. We can create an issue for that.

PR #831, opened 2025-07-17
activity: support batch operations in reset and update-options

spkane31:
## What was changed
* The `temporal activity {pause,update-options}` support batch operations through `--query`
* Consistently passing the `--identity` parameter to server requests

## Why?
Improved developer experience, can batch these requests through a visibility query instead of sending multiple requests

## Checklist

1. Closes: N/A

2. How was this tested:
New unit tests

3. Any docs updates needed?
Documentation updates are automatically generated with releases.


temporalcli/client.go
```diff
@@ -255,7 +255,11 @@ func payloadCodecInterceptor(
 	)
 }
 
-func clientIdentity() string {
+// if providedIdentity is empty, use the hostname and username to generate a default identity
+func clientIdentity(providedIdentity string) string {
```
spkane31: if `--identity` is provided use that, otherwise use the custom version using hostname and username.

go.mod
```diff
@@ -16,11 +16,11 @@ require (
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.39.0
-	go.temporal.io/api v1.50.0
+	go.temporal.io/api v1.50.1-0.20250717230205-2e8ba1309322
 	go.temporal.io/sdk v1.35.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
-	go.temporal.io/server v1.28.0
-	google.golang.org/grpc v1.71.0
+	go.temporal.io/server v1.29.0-135.0.0.20250718192231-215876cb9fec
```
cretz: Cannot merge with untagged server/API. Also, if the tag is not a server OSS release, this PR should go to `next-server` branch instead.

temporalcli/commands.activity_test.go
```diff
@@ -436,3 +436,89 @@ func (s *SharedServerSuite) TestUnpauseActivity_BatchSuccess() {
 	// unblock the activities to let them finish
 	failActivity.Store(false)
 }
+
+
+func (s *SharedServerSuite) TestResetActivity_BatchSuccess() {
+	var failActivity atomic.Bool
+	failActivity.Store(true)
+	s.Worker().OnDevActivity(func(ctx context.Context, a any) (any, error) {
+		time.Sleep(100 * time.Millisecond)
```
cretz: Ideally we can avoid sleeps in tests to avoid flakiness, but I have not checked this test in general. How long does this test take to run usually? Does the presence/absence of this sleep affect the assertion success?

temporalcli/commandsgen/commands.yml
```diff
@@ -3814,6 +3865,10 @@ commands:
           --workflow-id YourWorkflowId \
           --update-id YourUpdateId
       ```
+    options: 
+      - name: identity
```
cretz: Is this not already available via the new available-to-all-clients `identity` arg?

temporalcli/client.go
cretz: I think we should consider making this a method on the `*ClientOptions` object. But really we should probably just get rid of this method, inline it in `dialClient` and document on `dialClient` that after the call is made successfully, `ClientOptions.Identity` is guaranteed to be non-empty and it should be used to access identity if needed. We saw the same thing with `ClientOptions.Namespace`.

temporalcli/commandsgen/commands.yml
spkane31: Fixed this everywhere

temporalcli/commands.activity_test.go
spkane31: Removed this in a few tests.

temporalcli/commandsgen/commands.yml
```diff
@@ -4093,6 +4144,9 @@ option-sets:
           HTTP headers for requests to codec server.
           Format as a `KEY=VALUE` pair.
           May be passed multiple times to set multiple headers.
+      - name: identity
+        type: string
+        description: The identity of the user or client submitting this request.
```
cretz: May be worth clarifying that there is a default that includes `user@hostname`, but no biggie

cretz (commented):
LGTM, nothing blocking. Will approve if/when server and API dependencies are on tags (but doesn't have to be server OSS release since it's going to `next-server` not main).

temporalcli/commandsgen/commands.yml
```diff
@@ -335,9 +331,11 @@ commands:
   - name: temporal activity update-options
     summary: Update Activity options
     description: |
-      Update Activity options. Specify the Activity and Workflow IDs, and
-      options you want to update.
-      Updates are incremental, only changing the specified options.
+      Update the options of a running Activity that were passed into it from
+       a Workflow. Updates are incremental, only changing the specified 
+       options.
```
bergundy: nit: fix the indentation

```suggestion
      Update the options of a running Activity that were passed into it from
      a Workflow. Updates are incremental, only changing the specified 
      options.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -352,13 +350,35 @@ commands:
           --retry-maximum-interval DURATION \
           --retry-backoff-coefficient NewBackoffCoefficient \
           --retry-maximum-attempts NewMaximumAttempts
+      ```
 
+      You may follow this command with `temporal activity reset`, and the new values will apply after the reset.
+
+      Activity options can be updated in bulk with a visibility Query list filter.
```
bergundy: Either this:
```suggestion
      Activity options can be updated in bulk with a visibility query list filter. 
```
Or capitalize visibility too.

temporalcli/commandsgen/commands.yml
```diff
@@ -352,13 +350,35 @@ commands:
...
       ```
     options:
+      
       - name: activity-id
+        short: a
         type: string
-        description: Activity ID.
-        required: true
+        description: |
+          The Activity ID to update options. Can only be used without --query or --match-all. Either `activity-id` or `activity-type` must be provided, but not both.
```
bergundy: I would use this language across all of the mutually exclusive fields and add to the top level description saying that one of `--query` or `--workflow-id` must be supplied.

```suggestion
          The Activity ID to update options. Mutually exclusive with `--query`, `--match-all` and `activity-type`. Requires `--workflow-id` to be specified.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -352,13 +350,35 @@ commands:
...
+      - name: activity-type
+        short: g
+        type: string
+        description: |
+          Activities of this Type will be updated. Can only be used without --match-all. Either `activity-id` or `activity-type` must be provided, but not both.
+      - name: match-all
+        type: bool
+        description: |
+          Every activity should be updated. This flag is ignored if
+          activity-type is provided. Can only be used with --query.
```
bergundy: Why can't match all be used with a single activity?

temporalcli/commandsgen/commands.yml
```diff
@@ -554,26 +585,50 @@ commands:
...
+          --query 'TemporalResetInfo="property:activityType=Foo"'
+      ```
     options:
       - name: activity-id
         short: a
         type: string
-        description: Activity ID to pause.
+        description: The Activity ID to reset. Either `activity-id` or `activity-type` must be provided, but not both.
       - name: activity-type
         short: g
```
bergundy: Why is g the short flag form here?

temporalcli/commandsgen/commands.yml
```diff
@@ -554,26 +585,50 @@ commands:
...
+        description: |
+          Reset the Activity's heartbeats. Only works with --reset-attempts.
+      - name: match-all
+        type: bool
+        description: |
+          Every activity should be reset. This flag is ignored if
+          activity-type is provided. Can only be used with --query.
+      - name: jitter
+        type: duration
+        short: j
```
bergundy: Don't feel like we need a short flag here, let's use those sparingly.

temporalcli/commandsgen/commands.yml
cretz: To clarify my previous suggestion, don't actually set the `default` here, just inform users about it in the doc part. By using `default` here it actually sets this string as the default value which we don't want.

temporalcli/commandsgen/commands.yml
spkane31: I don't have the context here, @ychebotarev might remember why this decision was made. Found it introduced in this PR: https://github.com/temporalio/cli/pull/746/files#diff-1bf9d89c09230b7fb737937c657147b516bac4d39fa100086a30871c7abad39e

temporalcli/commandsgen/commands.yml
spkane31: match all is used to perform an action on all activities in a query, if you have a single activity (I'm assuming single activity id here) then why would you want to use match all in the first place? lmk if I am misunderstanding your question

temporalcli/commandsgen/commands.yml
ychebotarev: mistyped and was not noticed?  make them "i" and "t"? or remove altogether.

spkane31: Removed

temporalcli/commandsgen/commands.yml
spkane31: I think in an ideal world we could put the code here as a function but our generator isn't up to that. I'll remove the default field and put this in the description.

.github/workflows/trigger-docs.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: |
           gh workflow run update-cli-docs.yml \
             -R temporalio/documentation \
-            -r cli-docs-autoupdate \
+            -r main \
```
cretz: I wonder if some stuff from `main` has leaked into this PR. Should we make a separate PR to backport `main` to `next-server` @bergundy?

cretz (approved):
Nothing blocking, but would like this PR to not include things added in `main` since last `next-server` merge (i.e. need a `main` merge into `next-server` outside of this PR IMO or this branch needs to be based off the `next-server` branch head instead of `main` head bringing in extra things).

temporalcli/commandsgen/commands.yml
bergundy: Sorry, single workflow

.github/workflows/trigger-docs.yml
bergundy: We can it would just be a bit awkward to see a squashed no-op commit when we merge back into main later.

spkane31: PR here: https://github.com/temporalio/cli/pull/834

Hoping #834 will make this PR cleaner

cretz: So should this branch for this PR be recreated to be based off `next-server` instead of `main`?

temporalcli/commandsgen/commands.yml
spkane31: Not sure I understand here, this is for activity update-options. We don't take any information about workflows into account right?

.github/workflows/trigger-docs.yml
spkane31: Updated this PR to only include relevant changes, the previous unrelated changes were merged in w/ #834.

temporalcli/commandsgen/commands.yml
bergundy: > `Can only be used with --query.`

Why? Can't I pause all activities by providing `--workflow-id`?

spkane31: You have to provide activity-id or activity-type as well, you can't pause all activities in a workflow right now.

temporalcli/client.go
```diff
@@ -27,6 +27,18 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error)
 		return nil, fmt.Errorf("root command unexpectedly missing when dialing client")
 	}
 
+	if c.Identity == "" {
```
bergundy, line 30: I wonder why this is in your PR, it seems unrelated.

temporalcli/commandsgen/commands.yml
bergundy: I'm asking why not, is there a reason not to support it?

temporalcli/client.go
cretz: See https://github.com/temporalio/cli/pull/831#discussion_r2219239978. I think the issue is we apparently have a need for an `--identity` arg for all client uses not just this one, so it was added. So this code changed from `clientOptions.Identity = clientIdentity()` to `clientOptions.Identity = <user-provided-or-default>` and the `clientIdentity()` helper was removed since people should rely on the options value.

temporalcli/commandsgen/commands.yml
```diff
@@ -359,11 +353,31 @@ commands:
...
+        short: a
         type: string
-        description: Activity ID.
-        required: true
+        description: |
+          The Activity ID to update options. Mutually exclusive with `--query`, `--match-all`, and `--activity-type`. Requires `--workflow-id` to be specified.
+      - name: activity-type
+        type: string
+        description: |
+          Activities of this Type will be updated. Can only be used without --match-all. Either `activity-id` or `activity-type` must be provided, but not both.
```
bergundy:
```suggestion
          Activities of this Type will be updated. Mutually exclusive with `--match-all` and `activity-id`.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -359,11 +353,31 @@ commands:
...
+      - name: match-all
+        type: bool
+        description: |
+          Every activity should be updated. This flag is ignored if
+          activity-type is provided.
```
bergundy:
```suggestion
          Every activity should be updated. Mutually exclusive with `--activity-type` and `activity-id`.
```

temporalcli/commandsgen/commands.yml
```diff
@@ -359,11 +353,31 @@ commands:
       ```
 
       You may follow this command with `temporal activity reset`, and the new values will apply after the reset.
+
```
bergundy, line 356:
```suggestion

      Either `--activity-id`, `activity-type`, or `--match-all` must be specified.

```

go.mod
cretz: Reminder to update this and `go.temporal.io/api` to an actual tag before merging

cretz: Reminder to update this and `go.temporal.io/api` with an actual tag before merging

PR #834, opened 2025-07-23
Update next-server branch with changes in main

spkane31:
## What was changed
Updating `next-server` branch with changes to `main` 

## Why?
Consistency

## Checklist

1. Closes: NA

2. How was this tested:
NA

3. Any docs updates needed?
No

### DIFF
--- a/temporalcli/commandsgen/commands.yml
+++ b/temporalcli/commandsgen/commands.yml
@@ -304,7 +304,7 @@ commands:
         type: string
         description: Identity of the user submitting this request.
     option-sets:
-      - workflow reference
+      - workflow-reference
 
   - name: temporal activity fail
     summary: Fail an Activity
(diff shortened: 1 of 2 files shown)


bergundy:
I would have preferred a rebase to avoid the history being awkward but this is fine. I hope it doesn't complicate the merge back into main later.

spkane31:
> I would have preferred a rebase to avoid the history being awkward but this is fine. I hope it doesn't complicate the merge back into main later.

@bergundy is this rebase from `main` process something worth automating or should most of our PRs into this repo be made into `next-server`? The exception would be PRs made right before releases

bergundy:
> > I would have preferred a rebase to avoid the history being awkward but this is fine. I hope it doesn't complicate the merge back into main later.
> 
> @bergundy is this rebase from `main` process something worth automating or should most of our PRs into this repo be made into `next-server`? The exception would be PRs made right before releases

We should automate this and even consider blocking merges to `next-server` if it is out of sync with `main`. It's just not trivial since `next-server` may have conflicts with anything in `main`.

We should think about this process a bit more.

PR #836, opened 2025-07-25
Support multiple --input-meta flags for the same key

### LINKED ISSUE #833: [Feature Request] Support signaling with multiple message types from CLI and UI

I have a signal definition that takes in two parameters/args that are of two different message types. Currently, both the Temporal UI and CLI do not support this:

- Temporal UI accepts the signal, but it is dropped by the Temporal Worker saying only 1 parameter was specified
  - The Temporal UI does warn that only single payloads are accepted
  - exception
    ```  
    System.InvalidOperationException: Signal MySignalWithTwoParams given 1 parameter(s), but more than that are required by the signature
    at Temporalio.Worker.WorkflowInstance.DecodeArgs(MethodInfo method, RepeatedField`1 payloads, String itemName, Boolean dynamic, String dynamicArgPrepend)
    at Temporalio.Worker.WorkflowInstance.<>c__DisplayClass165_0.<<ApplySignalWorkflow>b__1>d.MoveNext()
    ```
- Temporal CLI accepts the signal, but it is dropped by the Temporal Worker saying it cannot decode one of the parameters (specified `messageType` twice - it takes the last one passed)
  ```
  Failed decoding signal args for {SignalName}, dropping the signal","exception.message":"Signal MySignalWithTwoParams had failure decoding parameters","exception.stacktrace":"System.InvalidOperationException: Signal MySignalWithTwoParams had failure decoding parameters\n ---> System.ArgumentException: Payload has protobuf message type protos.v1.bar but given type's message type is {desc.FullName}\n
  ```

I would like the Temporal UI and CLI to support sending signals that take in parameters with different message types

- Temporal UI example
  Image
- Temporal CLI
  ``` powershell
  temporal workflow signal `
    --workflow-id "MyWorkflow" `
    --type "MySignalWithTwoParams" `
    --input '{ \"foo\": \"foo\" }' `
    --input '{ \"bar\": \"bar\" }' `
    --input-meta "encoding=json/protobuf" `
    --input-meta "messageType=protos.v1.foo"
    --input-meta "messageType=protos.v1.bar"
  ```

### Additional context

We used two message types for reusability among different workflows and signals.

bergundy:
Closes #833

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -246,7 +246,7 @@ type PayloadInputOptions struct {
 func (v *PayloadInputOptions) buildFlags(cctx *CommandContext, f *pflag.FlagSet) {
 	f.StringArrayVarP(&v.Input, "input", "i", nil, "Input value. Use JSON content or set --input-meta to override. Can't be combined with --input-file. Can be passed multiple times to pass multiple arguments.")
 	f.StringArrayVar(&v.InputFile, "input-file", nil, "A path or paths for input file(s). Use JSON content or set --input-meta to override. Can't be combined with --input. Can be passed multiple times to pass multiple arguments.")
-	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Passing the same key more than once will apply its values in sequence to the corresponding --input arguments.")
+	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Repeated metadata keys are applied to the corresponding inputs in the provided order.")
 	f.BoolVar(&v.InputBase64, "input-base64", false, "Assume inputs are base64-encoded and attempt to decode them.")
 }
 
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -246,7 +246,7 @@ type PayloadInputOptions struct {
 func (v *PayloadInputOptions) buildFlags(cctx *CommandContext, f *pflag.FlagSet) {
 	f.StringArrayVarP(&v.Input, "input", "i", nil, "Input value. Use JSON content or set --input-meta to override. Can't be combined with --input-file. Can be passed multiple times to pass multiple arguments.")
 	f.StringArrayVar(&v.InputFile, "input-file", nil, "A path or paths for input file(s). Use JSON content or set --input-meta to override. Can't be combined with --input. Can be passed multiple times to pass multiple arguments.")
-	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Repeated metadata keys are applied to the corresponding inputs in the provided order.")
+	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Passing the same key more than once will apply its values in sequence to the corresponding --input arguments.")
 	f.BoolVar(&v.InputBase64, "input-base64", false, "Assume inputs are base64-encoded and attempt to decode them.")
 }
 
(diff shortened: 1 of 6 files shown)


temporalcli/commands.gen.go
```diff
@@ -244,7 +244,7 @@ type PayloadInputOptions struct {
 func (v *PayloadInputOptions) buildFlags(cctx *CommandContext, f *pflag.FlagSet) {
 	f.StringArrayVarP(&v.Input, "input", "i", nil, "Input value. Use JSON content or set --input-meta to override. Can't be combined with --input-file. Can be passed multiple times to pass multiple arguments.")
 	f.StringArrayVar(&v.InputFile, "input-file", nil, "A path or paths for input file(s). Use JSON content or set --input-meta to override. Can't be combined with --input. Can be passed multiple times to pass multiple arguments.")
-	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times.")
+	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Repeated metadata keys are applied to the corresponding inputs in the provided order.")
```
Sushisource, line 249:
```suggestion
	f.StringArrayVar(&v.InputMeta, "input-meta", nil, "Input payload metadata as a `KEY=VALUE` pair. When the KEY is \"encoding\", this overrides the default (\"json/plain\"). Can be passed multiple times. Passing the same key more than once will apply its values in sequence to the corresponding --input arguments.")
```

Honestly not sure this is better, but, I had no idea what was going on here until I saw the test. Maybe we just want an example in this help text?

bergundy: It's already pretty long so not sure we want to include the example in this single option's description.

temporalcli/payload.go
```diff
@@ -9,12 +9,23 @@ import (
 	"go.temporal.io/api/common/v1"
 )
 
-func CreatePayloads(data [][]byte, metadata map[string][]byte, isBase64 bool) (*common.Payloads, error) {
+func CreatePayloads(data [][]byte, metadata map[string][][]byte, isBase64 bool) (*common.Payloads, error) {
```
cretz, line 14: It's a little confusing following this code. May need some godoc on this method explaining this `map[string][][]byte` data structure. As it looks today, I'd call this expecting 1:1 with `data`, but that's not the case. Rather it seems that the first array value is applied unless there is one for the data index. So if I have 3 pieces of `data` and my `metadata` has 2 values for "encoding", say "foo" and "bar", the first data item and the third data item get "foo" and only the middle one gets "bar".

Not saying this is wrong necessarily, just confusing and worth documenting somewhere clearly so it is clear that the logic below is by intention and what callers should expect. Maybe just toss a "If metadata has an entry at a data index, it is used, otherwise it uses the metadata entry at index 0" above the function would be enough.

temporalcli/payload.go
```diff
@@ -9,12 +9,23 @@ import (
...
 	ret := &common.Payloads{Payloads: make([]*common.Payload, len(data))}
 	for i, in := range data {
-		// If it's JSON, validate it
-		if strings.HasPrefix(string(metadata["encoding"]), "json/") && !json.Valid(in) {
-			return nil, fmt.Errorf("input #%v is not valid JSON", i+1)
+		var metadataForIndex = map[string][]byte{}
```
cretz:
```suggestion
		metadataForIndex := make(map[string][]byte, len(metadata))
```

Pedantic, can ignore, but a nice thing to do

temporalcli/commands.workflow_exec.go
```diff
@@ -623,13 +623,23 @@ func (p *PayloadInputOptions) buildRawInputPayloads() (*common.Payloads, error)
...
 		}
-		metadata[metaPieces[0]] = []byte(metaPieces[1])
+		if vals, ok := metadata[metaPieces[0]]; ok {
+			if len(vals) == len(inData) {
+				return nil, fmt.Errorf("received more --input-meta flags for key %q than number of inputs", metaPieces[0])
+			}
+			metadata[metaPieces[0]] = append(vals, []byte(metaPieces[1]))
+		} else {
+			metadata[metaPieces[0]] = [][]byte{[]byte(metaPieces[1])}
+		}
```
cretz, line 639:
```suggestion
		metadata[metaPieces[0]] = append(metadata[metaPieces[0]], []byte(metaPieces[1]))
		if len(metadata[metaPieces[0]]) > len(inData) {
			return nil, fmt.Errorf("received more --input-meta flags for key %q than number of inputs", metaPieces[0])
		}
```

Pedantic, but for a map of slices, it's a bit more idiomatic Go to append to the nil result of the absent key rather than have conditionals (can use intermediate vars if you are concerned with the accesses)

temporalcli/commands.gen.go
Sushisource: If we don't include the example, maybe still go with this wording change? I'm worried almost no one is going to understand this from just reading the helptext.

Luckily it probably won't come up much either, so I suppose there's that.

Sushisource (approved):

temporalcli/payload.go
cretz: It looks like a comment was made as a result of this comment (it's a bit hard to know without an update to this thread).

PR #840, opened 2025-08-01
Update task queue config cli

sivagirish81:
## What was changed
+ Added a new `report-config` field in the describe command to retrieve updated task queue configs if exists.
+ Implement display logic for config table along with truncation of large update reasons or update identities.
**Note :  These changes are only a part of the legacy mode of the describeTaskQueueApi.**

```
./temporal task-queue describe \                                                               
 --task-queue=test-display-queue \
 --task-queue-type-legacy=activity \
 -n=default \
 --report-config \
 --legacy-mode
```

Sample Response
```
./temporal task-queue describe \
 --task-queue=hello-world \
 --task-queue-type-legacy=activity \
 -n=default \
 --report-config \
 --legacy-mode
Pollers:
  Identity  LastAccessTime  RatePerSecond

Task Queue Configuration:
Note: Long content may be truncated. Use --output json for full details.
              Setting                     Value              Reason        UpdatedBy       UpdatedTime    
  Fairness Key Rate Limit Default  100 requests/second  Fairness key test  api-tester  2025-08-12 04:31:46
```

---

+ Added separate config Subcommand for getting/setting configs.
```
./temporal task-queue config --help  
Manage Task Queue configuration:

temporal task-queue config [command] [options]

Available commands:
- get: Retrieve the current configuration for a task queue
- set: Update the configuration for a task queue

Usage:
  temporal task-queue config [command]

Available Commands:
  get         Get Task Queue configuration
  set         Set Task Queue configuration
```
---

+ Get Subcommand
```
./temporal task-queue config get --help
Retrieve the current configuration for a Task Queue:

temporal task-queue config get \
    --task-queue YourTaskQueue \
    --task-queue-type activity

This command returns the current configuration including:
- Queue rate limit: The overall rate limit of the task queue.
  This setting overrides the worker rate limit if set.
  Unless modified, this is the system-defined rate limit.
- Fairness key rate limit defaults: Default rate limits for fairness keys.
  If set, each individual fairness key will be limited to this rate,
  scaled by the weight of the fairness key.

Usage:
  temporal task-queue config get [flags]
```
Sample Response
```
./temporal task-queue config get \
  --task-queue=hello-world \
  --task-queue-type=activity \
  --namespace=default
Note: Long content may be truncated. Use --output json for full details.
              Setting                     Value              Reason        UpdatedBy       UpdatedTime    
  Fairness Key Rate Limit Default  100 requests/second  Fairness key test  api-tester  2025-08-12 04:31:46
```

---

+ Set Subcommand
```
./temporal task-queue config set --help
Update a Task Queue's overall rate limit and the default rate limit for all fairness keys:

temporal task-queue config set \
    --task-queue YourTaskQueue \
    --task-queue-type activity \
    --namespace YourNamespace \
    --queue-rps-limit <requests_per_second:float> \
    --queue-rps-limit-reason <reason_string> \
    --fairness-key-rps-limit-default <requests_per_second:float> \
    --fairness-key-rps-limit-reason <reason_string>

This command supports updating:
- Queue rate limits: Controls the overall rate limit of the task queue.
  This setting overrides the worker rate limit if set.
  Unless modified, this is the system-defined rate limit.
- Fairness key rate limit defaults: Sets default rate limits for fairness keys.
  If set, each individual fairness key will be limited to this rate,
  scaled by the weight of the fairness key.

To unset a rate limit, pass in 'default', for example: --queue-rps-limit default

Usage:
  temporal task-queue config set [flags]

Flags:
      --fairness-key-rps-limit-default float|default   Fairness key rate limit default in requests per second. Accepts a float; or 'default' to unset.
      --fairness-key-rps-limit-reason string           Reason for fairness key rate limit update.
  -h, --help                                           help for set
      --queue-rps-limit float|default                  Queue rate limit in requests per second. Accepts a float; or 'default' to unset.
      --queue-rps-limit-reason string                  Reason for queue rate limit update.
  -t, --task-queue string                              Task Queue name. Required.
      --task-queue-type string                         Task Queue type. Accepted values: workflow, activity, nexus. Accepted values: workflow, activity, nexus. Required.
```
   

Sample response : 
```
./temporal task-queue config set \
  --task-queue=hello-world \
  --task-queue-type=activity \
  --fairness-key-rps-limit-default=100 \
  --fairness-key-rps-limit-reason="Fairness key test" \
  --identity="api-tester" \
  --namespace=default
Successfully updated task queue configuration
  Config  {"fairnessKeysRateLimitDefault":{"rateLimit":{"requestsPerSecond":100},"metadata":{"reason":"Fairness key test","updateIdentity":"api-tester","updateTime":"2025-08-12T04:31:46.640Z"}}}
```

## Why?
+ Cli support for `UpdateTaskQueueConfig` api -  new api that allows updates of rate limits against task queues.

## Checklist

1. How was this tested:
+ Added tests.

2. Any docs updates needed?
Pending.

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -586,6 +586,6 @@ func (o *overrideDisplayTypeFlagValue) Type() string {
 	return o.displayType
 }
 
-func OverrideFlagDisplayType(flag *pflag.Flag, displayType string) {
+func overrideFlagDisplayType(flag *pflag.Flag, displayType string) {
 	flag.Value = &overrideDisplayTypeFlagValue{Value: flag.Value, displayType: displayType}
 }
--- a/temporalcli/commandsgen/code.go
+++ b/temporalcli/commandsgen/code.go
@@ -426,7 +426,7 @@ func (o *Option) writeFlagBuilding(selfVar, flagVar string, w *codeWriter) error
 		w.writeLinef("%v.%v(&%v.%v, %q%v, %q)", flagVar, flagMeth, selfVar, o.fieldName(), o.Name, defaultLit, desc)
 	}
 	if o.DisplayType != "" {
-		w.writeLinef("OverrideFlagDisplayType(%v.Lookup(%q), %q)", flagVar, o.Name, o.DisplayType)
+		w.writeLinef("overrideFlagDisplayType(%v.Lookup(%q), %q)", flagVar, o.Name, o.DisplayType)
 	}
 	if o.Required {
 		w.writeLinef("_ = %v.MarkFlagRequired(%v, %q)", w.importCobra(), flagVar, o.Name)
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -3,14 +3,15 @@
 package temporalcli
 
 import (
-	"os"
-	"time"
-
 	"github.com/mattn/go-isatty"
 
 	"github.com/spf13/cobra"
 
 	"github.com/spf13/pflag"
+
+	"os"
+
+	"time"
 )
 
 var hasHighlighting = isatty.IsTerminal(os.Stdout.Fd())
@@ -2252,10 +2253,10 @@ func NewTemporalTaskQueueConfigSetCommand(cctx *CommandContext, parent *Temporal
 	s.Command.Flags().Var(&s.TaskQueueType, "task-queue-type", "Task Queue type. Accepted values: workflow, activity, nexus. Accepted values: workflow, activity, nexus. Required.")
 	_ = cobra.MarkFlagRequired(s.Command.Flags(), "task-queue-type")
 	s.Command.Flags().StringVar(&s.QueueRpsLimit, "queue-rps-limit", "", "Queue rate limit in requests per second. Accepts a float; or 'default' to unset.")
-	OverrideFlagDisplayType(s.Command.Flags().Lookup("queue-rps-limit"), "float|default")
+	overrideFlagDisplayType(s.Command.Flags().Lookup("queue-rps-limit"), "float|default")
 	s.Command.Flags().StringVar(&s.QueueRpsLimitReason, "queue-rps-limit-reason", "", "Reason for queue rate limit update.")
 	s.Command.Flags().StringVar(&s.FairnessKeyRpsLimitDefault, "fairness-key-rps-limit-default", "", "Fairness key rate limit default in requests per second. Accepts a float; or 'default' to unset.")
-	OverrideFlagDisplayType(s.Command.Flags().Lookup("fairness-key-rps-limit-default"), "float|default")
+	overrideFlagDisplayType(s.Command.Flags().Lookup("fairness-key-rps-limit-default"), "float|default")
 	s.Command.Flags().StringVar(&s.FairnessKeyRpsLimitReason, "fairness-key-rps-limit-reason", "", "Reason for fairness key rate limit update.")
 	s.Command.Run = func(c *cobra.Command, args []string) {
 		if err := s.run(cctx, args); err != nil {


go.mod
```diff
@@ -174,3 +175,7 @@ require (
 	modernc.org/strutil v1.2.1 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
+
+replace go.temporal.io/api => go.temporal.io/api v1.51.1-0.20250725211336-3d6e39249ecf
+
+replace go.temporal.io/server => go.temporal.io/server v1.29.0-135.0.0.20250730212656-fcd84c4be641
```
sivagirish81: Will revert this when the server and api versions are updated.

go.mod
```diff
@@ -16,11 +16,11 @@ require (
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.39.0
-	go.temporal.io/api v1.50.0
+	go.temporal.io/api v1.51.1-0.20250725211336-3d6e39249ecf
```
cretz: Note, we can't merge with untagged API, but feel free to keep developing

temporalcli/commandsgen/commands.yml
```diff
@@ -3089,6 +3095,76 @@ commands:
...
+        description: |
+          Task Queue type.
+          Accepted values: workflow, activity, nexus.
+        enum-values:
+          - workflow
+          - activity
+          - nexus
+      - name: identity
+        type: string
+        description: Identity for the operation.
```
cretz: This is a client-level option as was done in #831 (can add that to this PR if it's going to `main` or make this PR go to `next-server`, your choice)

temporalcli/commandsgen/commands.yml
```diff
@@ -3089,6 +3095,76 @@ commands:
...
+      - name: namespace
+        type: string
+        description: Namespace for the operation.
```
cretz: Namespace is already part of every client-based command

temporalcli/commandsgen/commands.yml
```diff
@@ -2534,6 +2534,12 @@ commands:
       - name: disable-stats
         type: bool
         description: Disable task queue statistics.
+      - name: report-config
```
dnr, line 2564: What do people think about having a separate subcommand tree, like temporal task-queue config get and temporal task-queue config set instead of bundling it into describe? (or we could bundle it into describe also.) I'm feeling like describe is too much of a grab-bag now and when a user wants to mess with config, they would appreciate a config-focused command. The fact that it uses the same DescribeTaskQueue rpc under the hood doesn't matter.

temporalcli/commandsgen/commands.yml
```diff
@@ -2534,6 +2534,12 @@ commands:
...
+        type: bool
+        description: |
+          Include task queue configuration in the response.
+          When enabled, the command will return the current rate limit
+          configuration for the task queue.
```
dnr, line 2569:
```suggestion
          When enabled, the command will return the current configuration for the task queue.
```
don't limit things to rate limits, we can put other things in the config later

temporalcli/commandsgen/commands.yml
```diff
@@ -3089,6 +3095,76 @@ commands:
...
+        description: Identity for the operation.
+      - name: namespace
+        type: string
+        description: Namespace for the operation.
+      - name: queue-rate-limit
+        type: float
+        description: |
+          Queue rate limit in requests per second.
+          Use -1 to unset.
+        default: -99999
```
dnr: this is kind of weird. is this value supposed to be "unset"? can't you just not a default and check if the value is set/unset directly, or does cobra not support that?

if we do need a sentinel value, `math.NaN()` would make more sense

temporalcli/commands.taskqueue.go
```diff
@@ -360,14 +363,25 @@ func (c *TemporalTaskQueueDescribeCommand) runLegacy(cctx *CommandContext, args
 				Versioning: pi.WorkerVersionCapabilities,
 			})
 		}
+		// Capture config from the first partition (they should all be the same)
```
dnr, line 365: hmm, maybe it's time to remove this client-side fan-out feature? we have Describe doing fan-out now. or is that just for [deprecated] enhanced mode?

if we do keep client-side fan-out, I think we should either not ReportConfig on the ones other than the root, or disallow partition count > 1 with ReportConfig.

temporalcli/commands.taskqueue.go
```diff
@@ -382,7 +396,117 @@ func (c *TemporalTaskQueueDescribeCommand) runLegacy(cctx *CommandContext, args
...
+		// Queue Rate Limit
+		if config.QueueRateLimit != nil {
+			rateLimit := config.QueueRateLimit
+			value := "Not Set"
+			reason := ""
+			updatedBy := ""
+			updatedTime := ""
+
+			if rateLimit.RateLimit != nil && rateLimit.RateLimit.RequestsPerSecond > 0 {
+				value = fmt.Sprintf("%.0f requests/second", rateLimit.RateLimit.RequestsPerSecond)
```
dnr: "requests/second" is really verbose, let's use "rps" here?

also "%.0f" seems wrong, what if they set the value to 0.5 or 0.1?

temporalcli/commands.taskqueue.go
```diff
@@ -382,7 +396,117 @@ func (c *TemporalTaskQueueDescribeCommand) runLegacy(cctx *CommandContext, args
...
+			if rateLimit.Metadata != nil {
+				if rateLimit.Metadata.Reason != "" {
+					reason = truncateString(rateLimit.Metadata.Reason, 50)
+				}
+				if rateLimit.Metadata.UpdateIdentity != "" {
+					updatedBy = truncateString(rateLimit.Metadata.UpdateIdentity, 50)
+				}
+				if rateLimit.Metadata.UpdateTime != nil {
+					updateTime := rateLimit.Metadata.UpdateTime.AsTime()
+					updatedTime = updateTime.Format("2006-01-02 15:04:05")
```
dnr: don't write explicit time formats, use one of the standards like `time.RFC3339` or `time.DateTime`

temporalcli/commands.taskqueue.go
```diff
@@ -382,7 +396,117 @@ func (c *TemporalTaskQueueDescribeCommand) runLegacy(cctx *CommandContext, args
...
+				Setting:     "Queue Rate Limit",
+				Value:       value,
+				Reason:      reason,
+				UpdatedBy:   updatedBy,
+				UpdatedTime: updatedTime,
+			})
+		}
+
+		// Fairness Key Rate Limit Default
+		if config.FairnessKeysRateLimitDefault != nil {
```
dnr: this code is very repetitive, can we factor some out?

temporalcli/commands.taskqueue.go
```diff
@@ -382,7 +396,117 @@ func (c *TemporalTaskQueueDescribeCommand) runLegacy(cctx *CommandContext, args
...
+				if len(row.Reason) > 0 && strings.HasSuffix(row.Reason, "...") ||
+					len(row.UpdatedBy) > 0 && strings.HasSuffix(row.UpdatedBy, "...") {
+					hasTruncatedContent = true
+					break
+				}
+			}
+
+			if hasTruncatedContent {
+				cctx.Printer.Println(color.YellowString("Note: Long content has been truncated. Use --output json for full details."))
+			}
```
dnr: this seems like unnecessary complexity, the `...` is plenty. or wrap instead of truncate.

temporalcli/commands.taskqueue_update_config.go
```diff
@@ -0,0 +1,128 @@
...
+	cl, err := c.Parent.ClientOptions.dialClient(cctx)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	taskQueue := c.TaskQueue
+	taskQueueType := enums.TASK_QUEUE_TYPE_WORKFLOW // default
+	if c.TaskQueueType.String() != "" {
+		switch c.TaskQueueType.String() {
```
dnr: don't we have a function to parse task queue types somewhere already?

temporalcli/commands.taskqueue_update_config.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Build the request
+	if taskQueue == "" {
+		return fmt.Errorf("taskQueue name is required")
+	}
+	if taskQueueType == enums.TASK_QUEUE_TYPE_WORKFLOW {
+		if c.QueueRateLimit != 0 && c.QueueRateLimit != UnsetRateLimit {
+			return fmt.Errorf("setting rate limit on workflow task queues is not allowed")
+		}
+		return fmt.Errorf("taskQueueType is required")
```
dnr: ?

temporalcli/commands.taskqueue_update_config.go
```diff
@@ -0,0 +1,128 @@
...
+			request.UpdateQueueRateLimit = &workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate{
+				RateLimit: &taskqueue.RateLimit{
+					RequestsPerSecond: c.QueueRateLimit,
+				},
+				Reason: c.QueueRateLimitReason,
+			}
+		}
+	}
+
+	// Add fairness key rate limit default if specified (including unset)
```
dnr: factor things out a little?

temporalcli/commands.taskqueue_update_config.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Call the API
+	_, err = cl.WorkflowService().UpdateTaskQueueConfig(cctx, request)
+	if err != nil {
+		return fmt.Errorf("error updating task queue config: %w", err)
+	}
+
+	cctx.Printer.Println("Successfully updated task queue configuration")
+
+	// Print summary of what was updated
```
dnr: don't write special code here (we're going to forget to update it), output the final config that Update returns

temporalcli/commandsgen/commands.yml
cretz: This is a challenge we run into in a few places. I personally like CLI/SDK commands/API to be 1:1 with HTTP/gRPC API, otherwise we have seen it get really complicated as we add more things to RPC. But I do not have a strong opinion here.

dnr: The obvious counter-example is batch operations :shrug:

cretz: Yup, that's a `tctl` carry-over and one of the things I was referring to as "as we add more things to RPC". Batch not being a clear-cut command to represent it means as batch things are added we can't easily represent them. So now you have `--rps` on every workflow interaction because #675 needed it for batch. Granted making a single command multiple RPCs is the inverse of the problem here of multiple commands a single RPC, but there are still challenges for not being 1:1 and adding things (or being able to perform an invocation to get multiple things as results as you can with the HTTP API).

cretz: Having said the above, I'd be ok splitting it out, but I think we should discuss what this looks like in the SDK, because at least having the CLI and SDK share some client commonalities helps here. Regardless, not blocking and I'm ok with whatever y'all decide.

temporalcli/commandsgen/commands.yml
sivagirish81: Added identity as a client level option in this PR.

temporalcli/commandsgen/commands.yml
sivagirish81: Removed namespace and used the client level option.

temporalcli/commandsgen/commands.yml
sivagirish81: Implemented.

temporalcli/commandsgen/commands.yml
sivagirish81: Got rid of the sentinel value for unset values. Used cobra to handle it internally.

temporalcli/commands.taskqueue.go
sivagirish81: Changed to rps and used %.2f to report upto 2 decimal places.

temporalcli/commands.taskqueue.go
sivagirish81: Removed the check for truncation logic and simply left it at truncation and gave a hint message that tells the user to if they want to see the truncated content then they can view the json dump.

temporalcli/commands.taskqueue_update_config.go
sivagirish81: Created one in the commands.taskqueue.helper.go to make it resuable and parse string values in general i.e to handle string enum inputs.

temporalcli/commands.taskqueue_update_config.go
sivagirish81: Removed the incorrectly placed error message.

temporalcli/commands.taskqueue_update_config.go
sivagirish81: Done.

temporalcli/commands.taskqueue_update_config.go
sivagirish81: Changed to simply return the config that update returns as suggested.

temporalcli/commands.taskqueue.go
sivagirish81: Changed to time.RFC3339.

temporalcli/commands.taskqueue.go
sivagirish81: The legacy version of DescribeTaskQueue does not do fanout the fanout happens only in the enhanced version.

> if we do keep client-side fan-out, I think we should either not ReportConfig on the ones other than the root, or disallow partition count > 1 with ReportConfig.

Reading the data from config is not that expensive do we need to add this additional check?

temporalcli/commands.taskqueue.go
sivagirish81: Done.

temporalcli/commandsgen/commands.yml
sivagirish81: Added the report-config flag to the DescribeTaskQueue api to maintain 1:1 with HTTP/gRPC API.
Also introduced two additional subcommands task-queue config get/set to retrieve the task queue config and set the config.
+ Get - uses the DescribeTaskQueue rpc under the hood to retrieve the configs.
+ Set - uses the UpdateTaskQueueConfig rpc under the hood to set/update the configs.

temporalcli/commandsgen/commands.yml
```diff
@@ -3160,6 +3134,117 @@ commands:
...
+          Use -1 to unset.
+      - name: queue-rate-limit-reason
+        type: string
+        description: Reason for queue rate limit update.
+        hidden: true
+      - name: fairness-key-rate-limit-default
+        type: float
+        description: |
+          Fairness key rate limit default in requests per second.
+          Use -1 to unset.
```
cretz: Are we incorporating overrides as appear in https://github.com/temporalio/api/pull/626 or will that come later? The reason I ask is I am not sure the "-1 to unset" is the best way vs being able to provide what you want to unset. Are we expecting to force users to use this special `-1` value for overrides too?

temporalcli/commandsgen/commands.yml
```diff
@@ -3160,6 +3134,117 @@ commands:
...
+
+      ```
+      temporal task-queue config [command] [options]
+      ```
+
+      Available commands:
+      - `get`: Retrieve the current configuration for a task queue
+      - `set`: Update the configuration for a task queue
+
+  - name: temporal task-queue config get
```
cretz, line 3138: Will start new thread carrying over from https://github.com/temporalio/cli/pull/840/files#r2252052265.

I am not expecting `task-queue stats get` or `task-queue rate-limit get` or adding a new command for every thing we add as a response to task queue describe. This is just a shortcut for `task-queue describe --report-config` which we don't normally do (have multiple CLI commands doing the same thing). I guess I don't have a strong opinion against this shortcut so it has symmetry with `set`, but hopefully it remains no more than a shortcut.

temporalcli/commandsgen/commands.yml
sivagirish81: We haven’t implemented fairness weight overrides in the CLI yet. We were thinking on the lines of supporting repeated flags for setting and unsetting fairness weight overrides for fairness keys, for example:

```
--fairness-key-weight userA=5
--fairness-key-weight userB=10
--fairness-key-weight userC=-1   # Unsets the weight override for fairness key userC
```

cretz: :+1: I think we'll have to if we choose to use `-1` here too. An alternative is to have separate flags and API fields for unsetting things so that magic numbers are not in use. But if we have decided that the magic number -1 will be used in all of our interfaces (API and CLI for now, maybe UI and SDK later), ok.

stephanos: I'm also not a big fan of the magic `-1` number. Similarly. `0` here also means "stop all traffic" and not "unlimited" (like it often does). Seems dangerous.

I'm wondering if it's too much too have 3 flags (ie. `--queue-rate-limit`, `--unset-queue-rate-limit` and ... maybe `--queue-rate-stop` or sth). Or mix numbers and words; not sure if we allow that, like `--queue-rate-limit=reset` and `--queue-rate-limit=block`) - technically also a sentinel value but maybe a better one.

sivagirish81: `queue-rate-limit` and `fairness-key-rate-limit-default` are single values tied to a task queue. In these cases, we could avoid sentinel values and instead provide explicit unset flags for each.

[ Not implemented yet ]
For fairness-weight-overrides map, however, it’s different: if we don’t use sentinel values, then we’d need to introduce separate fields in the request (e.g. a list of fairness keys to unset). At the time, we chose sentinel values instead of adding extra fields to keep the request simple.

That said, if we do end up using sentinel values for fairness weight overrides, it might be more consistent to use them here as well.
>Or mix numbers and words; not sure if we allow that, like --queue-rate-limit=reset and --queue-rate-limit=block).

In cobra, By default I think a flag is always bound to a specific type so if we need to do this it needs to be a custom interface but I don't think this would be necessary.

>0 here also means "stop all traffic" and not "unlimited"

Having a separate `--queue-rate-stop` flag to set to 0, might be misleading considering users are allowed to set the rate limit to 0 in the api. But it would be helpful to let users know that 0 effectively halts traffic. Maybe we could add a line about the significance of 0 in the help/documentation of the command.

stephanos: > That said, if we do end up using sentinel values for fairness weight overrides, it might be more consistent to use them here as well.

I agree; let's be consistent. I'd prefer to be consistent about not using `-1`; at least user-facing.

> For fairness-weight-overrides [ ...] At the time, we chose sentinel values instead of adding extra fields to keep the request simple.

On the API level there, since it's not merged yet, we could consider using a nullable message instead, so that the API is clear about what it means to "unset".

> In cobra, By default I think a flag is always bound to a specific type so if we need to do this it needs to be a custom interface but I don't think this would be necessary.

That's a technical detail I'm sure we could overcome if we want to; let's work back from the user experience (of the CLI overall and this particular command). 

> Having a separate --queue-rate-stop flag to set to 0, might be misleading considering users are allowed to set the rate limit to 0 in the api. But it would be helpful to let users know that 0 effectively halts traffic. Maybe we could add a line about the significance of 0 in the help/documentation of the command.

I was actually adding a line to that in the docs and I realized that that's not enough, given the magnitude of the impact that the change potentially has. I'm thinking either using a separate flag _or_ add a confirmation step ("y/n") to make sure the user really knows what they're about to do.

cretz: > On the API level there, since it's not merged yet, we could consider using a nullable message instead, so that the API is clear about what it means to "unset".

The API this CLI refers to is merged and it does use `-1` IIUC. To be consistent on the API you'd need to affect that too. You can't use a nullable message because there is a difference between proto "unset" (i.e. don't change) and "remove". If we do want to do this, we would need separate API fields (and separate CLI flags) clearly saying to remove config on things. I do think we should either use `-1` everywhere (which is the current approach here but can be changed) or use unset flags/fields everywhere.

stephanos: > The API this CLI refers to is merged and it does use -1 IIUC.

Actually; that's not true: https://github.com/temporalio/cli/pull/840/files#diff-32890f13f9441f095cb29ab32b7bd127282d33234c3d381b2d45cabc9b282dd7R77 - ie the way to unset in the API is to set RateLimit to null. So it's not too late for the CLI and upcoming API to be changed.

cretz: Ah, makes sense, I misread an older commit I think. Yes, if server doesn't expect `-1`, awesome.

temporalcli/commandsgen/code.go
```diff
@@ -420,12 +420,26 @@ func (o *Option) writeFlagBuilding(selfVar, flagVar string, w *codeWriter) error
...
-		w.writeLinef("%v.%v(&%v.%v, %q%v, %q)",
-			flagVar, flagMeth, selfVar, o.fieldName(), o.Name, defaultLit, desc)
+		// Use standard flag
+		if o.Short != "" {
+			w.writeLinef("%v.%vP(&%v.%v, %q, %q%v, %q)",
+				flagVar, flagMeth, selfVar, o.fieldName(), o.Name, o.Short, defaultLit, desc)
+		} else {
+			w.writeLinef("%v.%v(&%v.%v, %q%v, %q)",
+				flagVar, flagMeth, selfVar, o.fieldName(), o.Name, defaultLit, desc)
+		}
```
stephanos: This has changed since the review.

temporalcli/commandsgen/commands.yml
```diff
@@ -3162,6 +3136,119 @@ commands:
...
+      - name: fairness-key-rps-limit-default
+        type: string
+        display-type: float|default
+        description: |
+          Fairness key rate limit default in requests per second.
+          Accepts a float; or 'default' to unset.
+      - name: fairness-key-rps-limit-reason
+        type: string
+        description: Reason for fairness key rate limit update.
+        hidden: true
```
stephanos, line 3236: This has changed since the review.

The rate limit parameters now carry `-rps-` in their name.
And `display-type` was added.

temporalcli/commandsgen/parse.go
```diff
@@ -215,6 +216,9 @@ func (o *Option) processSection() error {
 	if o.Type == "" {
 		return fmt.Errorf("missing option type")
 	}
+	if o.Type != "string" && o.DisplayType != "" {
+		return fmt.Errorf("display-type is only allowed for string options")
+	}
```
stephanos, line 221: This has changed since the review.

temporalcli/commands.taskqueue_config.go
```diff
@@ -0,0 +1,163 @@
...
+
+	var fairnessKeyRpsLimitDefaultParsed *taskqueue.RateLimit
+	if c.Command.Flags().Changed("fairness-key-rps-limit-default") {
+		var err error
+		if fairnessKeyRpsLimitDefaultParsed, err = parseRPS("fairness-key-rps-limit-default"); err != nil {
+			return err
+		}
+	} else if c.Command.Flags().Changed("fairness-key-rps-limit-default-reason") {
+		return fmt.Errorf("fairness-key-rps-limit-default-reason can only be set if fairness-key-rps-limit-default is updated")
+	}
```
stephanos, line 124: This has changed since the review.

temporalcli/commands.taskqueue_config.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	// Add fairness key rate limit default if specified (including unset)
+	if c.Command.Flags().Changed("fairness-key-rps-limit-default") {
+		request.UpdateFairnessKeyRateLimitDefault = &workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate{
+			RateLimit: fairnessKeyRpsLimitDefaultParsed,
+			Reason:    c.FairnessKeyRpsLimitReason,
+		}
+	}
```
stephanos, line 153: This has changed since the review. The helper is gone.

temporalcli/flags_custom_string.go
```diff
@@ -0,0 +1,42 @@
...
+func (v *customStringFlag) String() string {
+	if v == nil || v.target == nil || *v.target == "" {
+		return ""
+	}
+	return *v.target
+}
+
+func (v *customStringFlag) Type() string {
+	return v.displayType
+}
```
stephanos: This has changed since the review. It's a helper to allow control over the type that's displayed in the help text.

temporalcli/commandsgen/commands.yml
stephanos: I've pushed the changes to resolve this issue.

go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	go.temporal.io/api v1.53.0
 	go.temporal.io/sdk v1.36.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
-	go.temporal.io/server v1.28.1
+	go.temporal.io/server v1.29.0-142.2
```
stephanos: Bump to de-facto release candidate.

temporalcli/commandsgen/commands.yml
```diff
@@ -3162,6 +3136,119 @@ commands:
...
+        required: true
+        enum-values:
+          - workflow
+          - activity
+          - nexus
+
+  - name: temporal task-queue config set
+    summary: Set Task Queue configuration
+    description: |
+      Update a Task Queue's overall rate limit and the default rate limit for all fairness keys:
```
sivagirish81:
```suggestion
      Update configuration settings for a Task Queue.
```
I guess I missed this but like the get we can try to keep this generic to task queue configuration. Let's not limit this to the rate limit, we can add more things to this command later.

go.mod
cretz: I think increasing server version should ideally be done separately and can be merged into main with all of `next-server` instead of buried in this PR and not bring over things in `next-server`. I also think we need to be increasing the UI server version at the same time.

temporalcli/commandsgen/commands.yml
```diff
@@ -3162,6 +3136,119 @@ commands:
...
+        description: |
+          Queue rate limit in requests per second.
+          Accepts a float; or 'default' to unset.
+      - name: queue-rps-limit-reason
+        type: string
+        description: Reason for queue rate limit update.
+        hidden: true
+      - name: fairness-key-rps-limit-default
+        type: string
+        display-type: float|default
```
cretz, line 3229: If introducing a new property into this YAML format, can we update the admittedly-terse docs at the top of this file with information about it?

temporalcli/flags_custom_string.go
cretz, line 1: I wonder if this would be better as a utility, e.g.

```go
type overrideDisplayTypeFlagValue struct {
    pflag.Value
    displayType string
}

func (o *overrideDisplayTypeFlagValue) Type() string {
    return o.displayType
}

func OverrideFlagDisplayType(flag *pflag.Flag, displayType string) {
    flag.Value = &overrideDisplayTypeFlagValue{Value: flag.Value, displayType: displayType}
}
```

This is untested (I just hand-typed here in PR), but the idea is there. Then you can call with `OverrideFlagDisplayType(s.Command.Flags().Lookup("task-queue-type"), "string|float")` after you have created the flag the regular way. This is less code and more generally applicable. Also, not sure such a small amount of code really deserves its own file, can probably just put at the bottom of `commands.go`, but up to you.

temporalcli/commandsgen/code.go
```diff
@@ -420,12 +420,26 @@ func (o *Option) writeFlagBuilding(selfVar, flagVar string, w *codeWriter) error
 		// set default before calling Var so that it stores thedefault value into the flag
 		w.writeLinef("%v.%v = %v", selfVar, o.fieldName(), setDefault)
 	}
-	if o.Short != "" {
-		w.writeLinef("%v.%vP(&%v.%v, %q, %q%v, %q)",
-			flagVar, flagMeth, selfVar, o.fieldName(), o.Name, o.Short, defaultLit, desc)
+	if o.DisplayType != "" {
+		// Use custom flag
+		w.writeLinef("{")
```
cretz: I think we can just inline the work we need to do into one line of code instead of introducing a new block

go.mod
stephanos: Oh but it doesn't build without this change.

In this case, do we wait for the server release? I thought we would only switch to the stable release once that's out. Otherwise you have several PRs blocked for merging, right?

temporalcli/flags_custom_string.go
stephanos: I like that; definitely more composable.

go.mod
cretz: > Oh but it doesn't build without this change.

Right, neither does anything in `next-server` I assume

> In this case, do we wait for the server release?

You can target the `next-server` branch with this PR, that's where we put things that depend on non-OSS-stable server tags. We make unstable/RC CLI releases off of unstable OSS tags from there if there is an urgent need to get something in front of a user. But yes, for this to land in stable CLI, it needs to refer to stable OSS version.

temporalcli/commands.go
```diff
@@ -576,3 +576,16 @@ func encodeMapToPayloads(in map[string]any) (map[string]*commonpb.Payload, error
...
+type overrideDisplayTypeFlagValue struct {
+	pflag.Value
+	displayType string
+}
+
+func (o *overrideDisplayTypeFlagValue) Type() string {
+	return o.displayType
+}
+
+func OverrideFlagDisplayType(flag *pflag.Flag, displayType string) {
```
cretz: Realized this is same package, could make this non-exported (i.e. lowercase), but not a big deal

cretz (approved):
LGTM. Note, if this command/feature is not GA, we need to make sure we mark it as such in the description.

PR #845, opened 2025-08-25
Add new client-connect-timeout arg

### LINKED ISSUE #841: [Feature Request] Add `--client-connect-timeout`

Some users need to timeout client connect independent of the command timeout. Also, confirm this includes the get-system-info call that the Go SDK makes.

yuandrew:
## What was changed

Added new `client-connect-timeout` param

## Why?

Users requested

## Checklist

1. Closes #841

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/client.go
+++ b/temporalcli/client.go
@@ -207,13 +207,12 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error) 
 		// This is needed because the Go SDK overwrites the contextTimeout for GetSystemInfo, if not set
 		clientOptions.ConnectionOptions.GetSystemInfoTimeout = cctx.Options.ClientConnectTimeout
 
-		ctx := context.Background()
-		ctx, _ = context.WithTimeoutCause(ctx, cctx.Options.ClientConnectTimeout,
+		ctxWithTimeout, _ := context.WithTimeoutCause(cctx, cctx.Options.ClientConnectTimeout,
 			fmt.Errorf("command timed out after %v", cctx.Options.ClientConnectTimeout))
-		return client.DialContext(ctx, clientOptions)
+		return client.DialContext(ctxWithTimeout, clientOptions)
 	}
 
-	return client.Dial(clientOptions)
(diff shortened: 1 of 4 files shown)


temporalcli/client.go
```diff
@@ -203,6 +203,16 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error)
 	clientOptions.ConnectionOptions.DialOptions = append(
 		clientOptions.ConnectionOptions.DialOptions, cctx.Options.AdditionalClientGRPCDialOptions...)
 
+	if cctx.Options.ClientConnectTimeout != 0 {
+		// This is needed because the Go SDK overwrites the contextTimeout for GetSystemInfo, if not set
+		clientOptions.ConnectionOptions.GetSystemInfoTimeout = cctx.Options.ClientConnectTimeout
+
+		ctx := context.Background()
+		ctx, _ = context.WithTimeoutCause(ctx, cctx.Options.ClientConnectTimeout,
+			fmt.Errorf("command timed out after %v", cctx.Options.ClientConnectTimeout))
```
cretz:
```suggestion
		ctx, cancel := context.WithTimeoutCause(cctx, cctx.Options.ClientConnectTimeout,
			fmt.Errorf("command timed out after %v", cctx.Options.ClientConnectTimeout))
		defer cancel()
```

1. Need to have this based of the overall context (for things like canceling on SIGINT and such)
2. Might as well call `cancel()` as just a good practice

yuandrew: 1. Can do, we've previously only been using`context.Background`, since that's what `Dial` uses
2. Wouldn't calling defer here immediately cancel the context before the client can even be used? Since this function returns the client

cretz: > Wouldn't calling defer here immediately cancel the context before the client can even be used? Since this function returns the client

It'd cancel the _dial_ context, which is only for the purposes of the dial call IIUC

yuandrew: ahh you're right, not sure why I just assumed canceling that context would cancel the client's context

PR #846, opened 2025-08-27
Adds --client-authority flag to set the ConnectionOptions.Authority field on clients

wcharczuk-stripe:
## What was changed
Sets up (a) required field to add a `--client-authority` flag to the `temporal` cli.

## Why?
We will need this to generate the command flags file in github.com/temporalio/cli.

## Checklist
1. Flagged in support channel (2025-08-26)

2. How was this tested:
Ran local tests + validated the client options file contains the new field.

3. Any docs updates needed?

- https://docs.temporal.io/cli/cmd-options#address as well we will want to add the flag here


temporalcli/client.go
```diff
@@ -82,9 +82,10 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error)
 	// Override some values in client config profile that come from CLI args. Some
 	// flags, like address and namespace, have CLI defaults, but we don't want to
 	// override the profile version unless it was _explicitly_ set.
-	var addressExplicitlySet, namespaceExplicitlySet bool
+	var addressExplicitlySet, grpcAuthorityExplicitlySet, namespaceExplicitlySet bool
```
wcharczuk-stripe: if the env config isn't a factor here this is likely ~not necessary

temporalcli/client.go
```diff
@@ -172,6 +173,11 @@ func (c *ClientOptions) dialClient(cctx *CommandContext) (client.Client, error)
 	if err != nil {
 		return nil, fmt.Errorf("failed creating client options: %w", err)
 	}
+
+	if grpcAuthorityExplicitlySet {
```
wcharczuk-stripe: again, possible we don't need this

temporalcli/commandsgen/commands.yml
```diff
@@ -4051,6 +4051,10 @@ option-sets:
         description: Temporal Service gRPC endpoint.
         default: 127.0.0.1:7233
         implied-env: TEMPORAL_ADDRESS
+      - name: grpc-authority
```
cretz: Hrmm, I wonder if we should just call this `client-authority`. I know we have `--grpc-meta` since we needed to clarify that it is "gRPC metadata" aka headers without calling them headers, but all of these other gRPC options I'm not sure we need to prefix.

temporalcli/commandsgen/commands.yml
```diff
@@ -4051,6 +4051,10 @@ option-sets:
         description: Temporal Service gRPC endpoint.
         default: 127.0.0.1:7233
         implied-env: TEMPORAL_ADDRESS
+      - name: grpc-authority 
+        type: string
+        description: Temporal gRPC client :authority pseudoheader.
+        implied-env: TEMPORAL_GRPC_AUTHORITY
```
cretz:
```suggestion
```

I don't think this environment variable is leveraged, nor do we really need it for this option IMO

temporalcli/client.go
cretz:
```suggestion
	if c.GrpcAuthority != "" {
```

Don't need the flag-changed stuff above. That's for an advanced client env config stuff.

temporalcli/docs/activity.mdx
cretz, line 1: Looks like these docs were committed inadvertently (maybe we are missing a gitignore entry we should have)

temporalcli/commandsgen/commands.yml
wcharczuk-stripe: `ConnectionOptions.Authority` really is specific to grpc/http2 as the `:authority` header isn't relevant in other protocols, but something like `address` is reasonable to keep the same if you ever switched off grpc.

defer to you but imho given that there are no other `--client-*` prefixed options might want to keep it consistent with the other grpc specific flag.

temporalcli/docs/activity.mdx
wcharczuk-stripe: ok! i can remove

temporalcli/commandsgen/commands.yml
cretz: I would expect basically everything here to be grpc/http2 specific, and for everything that can technically apply to non-grpc/http2, authority (or "host name" in http1 parlance) is one of them.

We didn't prefix `grpc-meta` with `grpc-` because it was gRPC specific, we did it to disambiguate what "meta" means (we originally had headers which caused confusion), which we don't need here and not sure we want `grpc-` prefixes anywhere else we don't have to.

cretz (approved):

PR #850, opened 2025-09-02
fix: unknown command prints out before unknown flag

### LINKED ISSUE #849: [Bug] Error showing "unknown flag" before "unknown command"
### Describe the bug

Running:

    temporal foo bar --address localhost:7233

is showing "unknown flag: --address" instead of "unknown command" as expected. Flags are command specific, so it doesn't make sense to show unknown flag before unknown command. `temporal foo bar` does properly show "unknown command".

spkane31:
## What was changed
* When an invalid command and invalid flag is passed to the CLI then the error message will report on the invalid flag. This change will flip the order, and alert on an invalid command before invalid flags. 

## Why?
This is an improved developer experience, fixing an invalid flag on an invalid command will still fail. The more important issue is using a valid command, which might have the flag referenced.

image

## Checklist

1. Closes #849 

2. How was this tested:
Tested manually and added unit tests

3. Any docs updates needed?
No

### DIFF
--- a/temporalcli/commands.go
+++ b/temporalcli/commands.go
@@ -552,59 +552,19 @@ func aliasNormalizer(aliases map[string]string) func(f *pflag.FlagSet, name stri
 	}
 }
 
-// firstUnknownCommandToken returns the first positional argument that is not a flag
-// or a value for a known root-level flag. This helps identify which token is the
-// unknown command when no command actually ran (e.g., "temporal foo --address ...").
 func unknownCommandFromArgs(cctx *CommandContext, args []string) string {
-	knownTop := map[string]struct{}{}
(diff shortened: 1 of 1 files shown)


temporalcli/commands.go
```diff
@@ -21,7 +21,6 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/temporalio/cli/temporalcli/internal/printer"
 	"github.com/temporalio/ui-server/v2/server/version"
-	"go.temporal.io/api/common/v1"
```
spkane31, line 24: My editor flagged this because it was imported twice.

cretz:
Closing PR as we're concerned the complication of the solution outweighs the need. See issue at #849.

PR #852, opened 2025-09-04
Added backticks around ascii warnings to format correctly in docs

flippedcoder:
## What was changed
The ASCII warning blocks were rendering incorrectly on the docs site. The fix is to add backticks around these blocks. It looks like the format will still be ok in the CLI `--help` messages, they will just be bolded like some of the commands in the `--help` results.

Without fix:
image

With fix:
Screenshot 2025-09-04 at 10 40 05

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


yuandrew (approved):
LGTM! you'll need to run `go run ./temporalcli/internal/cmd/gen-commands` to re-generate the `commands.gen.go` file

PR #854, opened 2025-09-25
Annotate --activity-type as experimental in activity pause command

gow:
## What was changed
Marking ` --activity-type` parameter as experimental param in activity pause command

## Why?
The behavior of this may change in the future.


bergundy (approved):
Thanks!

cretz (approved):

PR #857, opened 2025-10-03
Bump server to 1.29.0

bergundy:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/temporalio/cli
 
-go 1.24.5
+go 1.25.0
```
bergundy, line 3: Looks like running `go mod tidy` also bumped the Go version.

PR #858, opened 2025-10-06
Add `temporal worker deployment manager-identity` commands

carlydf:
## What was changed
Add `temporal worker deployment manager-identity` commands (`set` and `unset`)

## Why?
So that CLI clients of Worker Deployments that are managed by an automated system (such as the Kubernetes Worker Controller) can safely coordinate with that client to make intended changes and ensure they are not overwritten.

## Checklist

1. Closes 

2. How was this tested:
Working on it. I had some tests failing that were related to activity commands, which I suspect will pass once I pull in the changes on `next-server` branch.

3. Any docs updates needed?


go.mod
```diff
@@ -16,18 +16,18 @@ require (
...
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.39.0
-	go.temporal.io/api v1.50.0
-	go.temporal.io/sdk v1.35.0
+	go.temporal.io/api v1.53.0
+	go.temporal.io/sdk v1.36.1-0.20251002083308-f069ff4410ad
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
-	go.temporal.io/server v1.28.0
-	google.golang.org/grpc v1.71.0
+	go.temporal.io/server v1.29.0-141.4
```
cretz: If this is going to a non-stable OSS server tag, I think you may need to change the target branch to `next-server` (may need to wait a bit since `next-server` may be under maintenance by @bergundy). Or if this works with stable 1.29, can wait until #856 merged and merge main back into this branch.

temporalcli/commandsgen/commands.yml
```diff
@@ -1233,6 +1234,157 @@ commands:
...
+    option-sets:
+      - deployment-name
+    options:
+      - name: manager-identity
+        type: string
+        description: New Manager Identity. Required unless --self is specified.
+      - name: self
+        type: bool
+        description: Set Manager Identity to the identity of the user submitting this request. Required unless --manager-identity is specified.
+      - name: identity
```
cretz: As of #831, `--identity` will be present on _all_ client-based commands, so will probably be able to remove from here and next command and reference it from the parent after dialing the client (because that's when the default is set).

go.mod
bergundy: This should work with 1.29.0, if it doesn't it can't go into this CLI release.

temporalcli/commands.worker.deployment_test.go
```diff
@@ -569,3 +570,73 @@ func (s *SharedServerSuite) TestDeployment_Ramping() {
...
+	defer w.Stop()
+
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		res := s.Execute(
+			"worker", "deployment", "list",
+			"--address", s.Address(),
+		)
+		assert.NoError(t, res.Err)
+		assert.Contains(t, res.Stdout.String(), deploymentName)
+	}, 30*time.Second, 100*time.Millisecond)
```
cretz, line 600: Can this test really take 30s or is this just a common value we're using everywhere?

EDIT: I saw in downloaded junit files that this test takes 1.2s. That's not too bad. Tests `TestDeployment_Describe_Drainage` and `TestDeployment_List` take over 2s, tests `TestDeployment_Ramping` and `TestDeployment_Set_Current_Version` take over 5s, and test `TestWorkflow_Batch_Update_Options_Versioning_Override` takes over 10s (the slowest test in the repo). May want to open an issue to look into improving those, but not blocking this PR.

carlydf: Glad you pointed this out! I just copied that timeout from the tests Antonio wrote, didn't base it on a specific time expectation. Things may have taken longer in the initial pre-release because we weren't doing the task queue user data in batches, which we now are.

PR #860, opened 2025-10-09
Add --allow-no-pollers flag to set-current-version and set-ramping-version

carlydf:
## What was changed
Add --allow-no-pollers flag to set-current-version and set-ramping-version

## Why?
So that users can make a version current / ramping before it has pollers.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.worker.deployment.go
+++ b/temporalcli/commands.worker.deployment.go
@@ -15,16 +15,16 @@ import (
 
 type versionSummariesRowType struct {
 	DeploymentName string    `json:"deploymentName"`
-	BuildId        string    `json:"buildId"`
+	BuildID        string    `json:"BuildID"`
 	DrainageStatus string    `json:"drainageStatus"`
 	CreateTime     time.Time `json:"createTime"`
 }
 
 type formattedRoutingConfigType struct {
 	CurrentVersionDeploymentName        string    `json:"currentVersionDeploymentName"`
-	CurrentVersionBuildId               string    `json:"currentVersionBuildId"`
+	CurrentVersionBuildID               string    `json:"currentVersionBuildID"`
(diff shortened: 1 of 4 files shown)


temporalcli/commands.worker.deployment.go
```diff
@@ -549,7 +551,7 @@ func (c *TemporalWorkerDeploymentSetCurrentVersionCommand) run(cctx *CommandCont
 		safeModeMessage: "Current",
 		deploymentName:  c.DeploymentName,
 	})
-	if err != nil {
+	if err != nil && !(errors.As(err, new(*serviceerror.NotFound)) && c.AllowNoPollers) {
```
cretz, line 624: What if it's namespace not found? May not be a big deal to accidentally continue in the face of that, just want to check

carlydf: If it's namespace not found, the subsequent API call will also return namespace not found and the user will see the error then. This call is only doing a "DescribeWorkerDeployment" to get a conflict token to use in the SetCurrent/SetRamping API call that the user is invoking. If there are no pollers, it's possible that the Worker Deployment doesn't exist yet, in which case there is no conflict token to get.

TL;DR, any important NotFound errors will surface in the actual API call that happens after the conflict token is fetched

bergundy: `serviceerror.NotFound` is different than `serviceerror.NamespaceNotFound`: https://github.com/temporalio/api-go/blob/0a581a07855239d71a73ce3fa44eab81e34a7a32/serviceerror/namespace_not_found.go#L14

cretz: Ah, makes sense, unsure if there are any other possible gRPC not found status codes for this call

PR #863, opened 2025-10-10
fix: improve consistency of backticks in usage messages

stpierre:
## What was changed

* Remove extraneous trailing backticks
* Add missing backticks where obviously needed
* Improve consistency of quoting

This doesn't fix every inconsistent use of backticks, but it does remove a few dangling backticks, and in places where quoting was inconsistent (i.e., a mix of backticks, quotes, and/or no quotes in a single usage message) makes it slightly more consistent.

## Why?

Prettier usage :)

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.gen.go
+++ b/temporalcli/commands.gen.go
@@ -379,7 +379,7 @@ func NewTemporalCommand(cctx *CommandContext) *TemporalCommand {
 	s.Command.PersistentFlags().StringVar(&s.Env, "env", "default", "Active environment name (`ENV`).")
 	cctx.BindFlagEnvVar(s.Command.PersistentFlags().Lookup("env"), "TEMPORAL_ENV")
 	s.Command.PersistentFlags().StringVar(&s.EnvFile, "env-file", "", "Path to environment settings file. Defaults to `$HOME/.config/temporalio/temporal.yaml`.")
-	s.Command.PersistentFlags().StringVar(&s.ConfigFile, "config-file", "", "File path to read TOML config from, defaults to `$CONFIG_PATH/temporal/temporal.toml` where `$CONFIG_PATH` is defined as `$HOME/.config` on Unix, \"$HOME/Library/Application Support\" on macOS, and %AppData% on Windows. EXPERIMENTAL.")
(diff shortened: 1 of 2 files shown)


temporalcli/commands.gen.go
```diff
@@ -3506,9 +3506,9 @@ func NewTemporalWorkflowListCommand(cctx *CommandContext, parent *TemporalWorkfl
 	s.Command.Use = "list [flags]"
 	s.Command.Short = "Show Workflow Executions"
 	if hasHighlighting {
-		s.Command.Long = "List Workflow Executions. The optional \x1b[1m--query\x1b[0m limits the output to\nWorkflows matching a Query:\n\n\x1b[1mtemporal workflow list \\\n    --query YourQuery\x1b[1m\x1b[0m\n\nVisit https://docs.temporal.io/visibility to read more about Search Attributes\nand Query creation. See \x1b[0mtemporal batch --help` for a quick reference.\n\nView a list of archived Workflow Executions:\n\n\x1b[1mtemporal workflow list \\\n    --archived\x1b[0m"
+		s.Command.Long = "List Workflow Executions. The optional \x1b[1m--query\x1b[0m limits the output to\nWorkflows matching a Query:\n\n\x1b[1mtemporal workflow list \\\n    --query YourQuery\x1b[0m\n\nVisit https://docs.temporal.io/visibility to read more about Search Attributes\nand Query creation. See \x1b[1mtemporal batch --help\x1b[0m for a quick reference.\n\nView a list of archived Workflow Executions:\n\n\x1b[1mtemporal workflow list \\\n    --archived\x1b[0m"
```
dandavison, line 3568: Anyone know why the ANSI escapes changed here? Haven't tested in a terminal but the new version LGTM. (`1m` is a transition to bold and `0m` is a transition that removes all flags.)

stpierre: I believe this is related to [this change](https://github.com/temporalio/cli/pull/863/files#diff-1bf9d89c09230b7fb737937c657147b516bac4d39fa100086a30871c7abad39eL3668); previously there was an extraneous hanging backtick on the code block for `temporal workflow list --query YourQuery`, so the parser that creates the highlighted version of the help text may have gotten a bit confused. Here's how it looks today: 

image

With this patch:

image

This is actually the exact typo that I noticed that sent me down this rabbit hole 😄

dandavison: Excellent, thanks for looking into this.

PR #867, opened 2025-10-24
Implement ListWorkers and DescribeWorker

### LINKED ISSUE #868: [Feature Request] Support ListWorkers and DescribeWorker

New API, we want to add CLI commands to make these gRPC requests for users to be able to query this information easily

yuandrew:
## What was changed

Implemented the ListWorkers and DescribeWorker gRPC requests

note: list workers doesn't display all heartbeat information. Open to feedback on if we should display more/all/less fields. See screenshots in testing section for an example.

PageSize is also not supported in server for the ListWorkers call, so that was left out. Added a `limit` field to allow users to limit how many results are displayed, similar to other CLI commands.

Sample output of the two new commands
image

## Why?

New feature.

## Checklist

1. Closes #868

2. How was this tested:

Added tests, because Go SDK doesn't support worker heartbeating yet, we manually send a `RecordWorkerHeartbeatRequest` gRPC request to mimic the behavior of a worker.

3. Any docs updates needed?


temporalcli/commandsgen/commands.yml
```diff
@@ -1404,6 +1404,78 @@ commands:
         type: bool
         description: Don't prompt to confirm unset Manager Identity.
 
+  - name: temporal worker heartbeat
```
cretz: I don't think the mechanism by which we get worker information (heartbeat atm) should affect the command name. If we have a concept in our API of "list workers", we should have a concept in our CLI of `worker list`, and if we have a concept in our API of "describe worker", we should have a concept in our CLI of `worker describe`.

yuandrew: that's fair, users don't really need to know the heartbeat mechanism exists

Sushisource: Yes, plus 1 on that front

temporalcli/commandsgen/commands.yml
```diff
@@ -1404,6 +1406,44 @@ commands:
...
+      Get a list of workers to the specified namespace.
+      
+      ```
+      temporal worker list --namespace YourNamespace --query 'taskQueue="YourTaskQueue"'
+      ```
+    options:
+      - name: namespace
+        short: n
+        type: string
+        description: Temporal Service Namespace.
```
cretz: This is already implied on all client calls (available via `c.Parent.Namespace` IIRC), and is actually important that you use the option after the `dialClient` call (see that call's godoc). You can remove the explicit option for namespace (here and describe)

temporalcli/commandsgen/commands.yml
```diff
@@ -1404,6 +1406,44 @@ commands:
...
+      temporal worker describe --namespace YourNamespace --worker-instance-key YourKey
+      ```
+    options:
+      - name: namespace
+        short: n
+        type: string
+        description: Namespace this worker belongs to.
+      - name: worker-instance-key
+        type: string
+        description: Worker instance key to describe.
```
cretz, line 1437: Should be set to `required: true` I think

temporalcli/commands.worker.go
cretz, line 1: Still some identifiers in here referencing "heartbeat" even though it doesn't apply here

cretz, line 1: Don't forget tests

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+		Namespace:         c.Parent.Namespace,
+		WorkerInstanceKey: c.WorkerInstanceKey,
+	})
+	if err != nil {
+		return err
+	}
+
+	if cctx.JSONOutput {
+		return cctx.Printer.PrintStructured(resp, printer.StructuredOptions{})
+	}
```
cretz, line 104: I understand we may not have done this with workflow listing because they are different values, but here if we're willing to return a proto-JSON-serialized form of `WorkerInfo` for "worker describe", I think for JSON of "worker list" we should be willing to do the same. It is a nice comfort from a user POV that the JSON is clear/obvious and consistent from describe to list if it's actually the same object (which it is here).

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+	Status            string    `json:"status"`
+	TaskQueue         string    `json:"taskQueue"`
+	WorkerIdentity    string    `json:"workerIdentity"`
+	HostName          string    `json:"hostName"`
+	Deployment        string    `json:"deployment,omitempty" cli:",cardOmitEmpty"`
+	HeartbeatTime     time.Time `json:"heartbeatTime"`
+	Elapsed           string    `json:"elapsedSinceLastHeartbeat"`
+}
+
+type formattedWorkerDeploymentVersionRef struct {
```
rkannan82: Do we have to introduce this prefix - formatted? The server and CLI structs are anyway in different packages.

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+	LastSuccessfulPollTime time.Time `json:"lastSuccessfulPollTime"`
+	IsAutoscaling          bool      `json:"isAutoscaling"`
+}
+
+type formattedPluginInfo struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+type formattedWorkerHeartbeatDetail struct {
```
rkannan82: Suggest dropping Heartbeat from here and other places?

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	deploymentpb "go.temporal.io/api/deployment/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	workerpb "go.temporal.io/api/worker/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+type workerHeartbeatListRow struct {
```
rkannan82: WorkerStatus or WorkerSummary? So that it is aligned with the semantics. IIUC, the list workers is just giving some basic information and the DescribeWorker is giving more details.

temporalcli/commandsgen/commands.yml
yuandrew: yes ty

temporalcli/commands.worker.go
yuandrew: good point, ty

yuandrew: Have a blurb in the PR description, 

> Didn't add tests due to worker heartbeating not being supported in the Go SDK. But verified locally after running a python worker.

Happy to add tests once the Go SDK is implemented, if that sounds alright with you

temporalcli/commands.worker.go
yuandrew: good point, changed this to resp.GetWorkerInfo()

temporalcli/commands.worker.go
yuandrew: I guess if every struct has the prefix "formatted", that doesn't really add much value. removed

temporalcli/commands.worker.go
yuandrew: we can probably be vague enough and just remove `heartbeat` for  `workerListRow`

temporalcli/commands.worker.go
cretz: Can we wait on merging until those tests become available? Alternatively, due to what I assume are some SDK-built-in limitations on heartbeat frequency, it might be better if you manually report heartbeats via gRPC in your tests here as if you had workers instead of re-testing the SDK side of the heartbeat behavior if it won't have good hooks for easy testing.

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/temporalio/cli/temporalcli/internal/printer"
+	deploymentpb "go.temporal.io/api/deployment/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	workerpb "go.temporal.io/api/worker/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+type workerListRow struct {
+	WorkerInstanceKey string    `json:"workerInstanceKey"`
```
cretz: Do you still need all of these JSON annotations?

temporalcli/commands.worker.go
yuandrew: good point! I can just manually send a heartbeat gRPC to mimic a worker

temporalcli/commands.worker.go
```diff
@@ -0,0 +1,381 @@
...
+	Status            string
+	TaskQueue         string
+	WorkerIdentity    string
+	HostName          string
+	Deployment        string
+	HeartbeatTime     time.Time
+	Elapsed           string
+}
+
+type workerDeploymentVersionRef struct {
```
cretz, line 27: I see the JSON tags were removed for the above struct but remain for all of these others. Are they still needed?

yuandrew: oops! Didn't realize we were no longer json printing **any** of these structs. Removed

temporalcli/commandsgen/commands.yml
```diff
@@ -1404,6 +1406,37 @@ commands:
...
+      
+      ```
+      temporal worker list --namespace YourNamespace --query 'taskQueue="YourTaskQueue"'
+      ```
+    options:
+      - name: query
+        short: q
+        type: string
+        description: Content for an SQL-like `QUERY` List Filter.
+      - name: limit
```
spkane31, line 1422: Is there a `skip` parameter for this endpoint? Wondering how/if we support pagination

yuandrew: `limit` and `query` are both optional parameters, so I don't think we need a `skip` param. Users can just omit if they just want the default return value

Pagination is built into the [API def](https://github.com/temporalio/api/blob/060dbc3216ca0032c1b1f4f82a7b7c37a035448c/temporal/api/workflowservice/v1/request_response.proto#L2478), but not currently supported in server (@rkannan82  correct me if I'm wrong). But we do intend to support pagination down the line

spkane31: 🚢

PR #875, opened 2025-11-18
Commands generator binary

stephanos:
## What was changed

Make `gen-commands` and `gen-docs` a binary that can be used any YAML input.

## Why?

Temporal wants to re-use the `commandsgen` tool for other _internal_ CLIs.

**It is not supported for users outside of Temporal.** 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


.github/workflows/ci.yaml
```diff
@@ -10,10 +10,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, macos-13, windows-latest, ubuntu-arm]
+        os:
+          [
+            ubuntu-latest,
+            macos-latest,
+            macos-15-intel,
```
stephanos, line 17: Image

internal/commandsgen/code.go
```diff
@@ -8,12 +8,10 @@ import (
 	"regexp"
 	"sort"
 	"strings"
-
-	"go.temporal.io/server/common/primitives/timestamp"
```
stephanos, line 12: Copied `parseDuration` to remove server dependency here.

internal/temporalcli/commands.schedule_test.go
```diff
@@ -222,14 +222,17 @@ func (s *SharedServerSuite) TestSchedule_List() {
...
-
-	res = s.Execute(
-		"schedule", "list",
-		"--address", s.Address(),
-		"--really-long",
-	)
-	s.NoError(res.Err)
-	out := res.Stdout.String()
+	var out string
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
stephanos, line 226: Was flaky.

internal/temporalcli/commands.worker.deployment_test.go
```diff
@@ -393,7 +393,11 @@ func (s *SharedServerSuite) TestDeployment_List() {
 			"worker", "deployment", "list",
 			"--address", s.Address(),
 		)
-		s.NoError(res.Err)
+		assert.NoError(t, res.Err)
+		assert.Contains(t, res.Stdout.String(), deploymentName1)
+		assert.Contains(t, res.Stdout.String(), version1.BuildID)
+		assert.Contains(t, res.Stdout.String(), deploymentName2)
+		assert.Contains(t, res.Stdout.String(), version2.BuildID)
```
stephanos, line 398: Was flaky.

.github/workflows/ci.yaml
```diff
@@ -51,11 +58,16 @@ jobs:
           retention-days: 14
 
       - name: Regen code, confirm unchanged
-        if: ${{ matrix.checkGenCodeTarget }}
+        if: ${{ matrix.checkGenCommands }}
         run: |
-          go run ./internal/cmd/gen-commands
+          go run ./cmd/gen-commands -pkg temporalcli -context "*CommandContext" < internal/temporalcli/commands.yaml > internal/temporalcli/commands.gen.go
```
stephanos: This is where I would invoke `make gen` but received strong objections last time.

.github/workflows/ci.yaml
```diff
@@ -51,11 +58,16 @@ jobs:
...
           git diff --exit-code
 
+      - name: Generate docs, confirm working
+        if: ${{ matrix.checkGenCommands }}
+        run: |
+          go run ./cmd/gen-docs -output dist/docs < internal/temporalcli/commands.yaml
+
```
stephanos, line 70: Better to know immediately if this is broken.

internal/commandsgen/code.go
```diff
@@ -59,6 +83,7 @@ type codeWriter struct {
 	buf         bytes.Buffer
 	allCommands []Command
 	OptionSets  []OptionSets
+	contextType string
```
stephanos, line 86: Allow custom context name.

internal/temporalcli/commands.gen.go
```diff
@@ -4227,3 +4235,119 @@ func NewTemporalWorkflowUpdateOptionsCommand(cctx *CommandContext, parent *Tempo
 	}
 	return &s
 }
+
+var reDays = regexp.MustCompile(`(\d+(\.\d*)?|(\.\d+))d`)
```
stephanos, line 4239: As this is generated code, I think it's fine for the `var` and `type` to not be at the top here to keep the codegen simple. Similarly, a single generated file seems easier do manage and generate then multiple.

internal/commandsgen/code.go
```diff
@@ -98,6 +123,36 @@ func (c *codeWriter) importPkg(pkg string) string {
 
 func (c *codeWriter) importCobra() string { return c.importPkg("github.com/spf13/cobra") }
 
+// appendGoSource parses a Go source file, registers its imports, and appends
+// everything after the import block to the output buffer.
+func (c *codeWriter) appendGoSource(src string) error {
```
stephanos, line 128: It was either this, string literals or regex; this seems like the best approach to me.

.github/workflows/ci.yaml
cretz: Yes, IMO we should avoid adding platform dependent code. Any contributor on any supported platform should be able to basically run the same commands as our CI does

internal/cmd/gen-commands/main.go
cretz, line 1: I think this and the other could also be considered a "hidden" `internal` sub-command of the existing CLI (probably manually added, not via existing yaml file, which allows it to be easily versioned/downloaded and doesn't require these top-level dirs. The bootstrapping problem is very slight.

cmd/gen-commands/main.go
```diff
@@ -0,0 +1,53 @@
...
+		pkg         string
+		contextType string
+	)
+
+	flag.StringVar(&pkg, "pkg", "main", "Package name for generated code")
+	flag.StringVar(&contextType, "context", "*CommandContext", "Context type for generated code")
+	flag.Parse()
+
+	// Read input from stdin
+	yamlBytes, err := io.ReadAll(os.Stdin)
```
cretz: Personal preference is to (also?) accept files for both directions, but it's not a big deal for an internal tool

CONTRIBUTING.md
```diff
@@ -20,18 +20,18 @@ Example to run a single test case:
 
 ## Adding/updating commands
 
-First, update [commands.yml](internal/commandsgen/commands.yml) following the rules in that file. Then to regenerate the
-[commands.gen.go](internal/commands.gen.go) file from code, simply run:
+First, update [commands.yaml](internal/temporalcli/commands.yaml) following the rules in that file. Then to regenerate the
+[commands.gen.go](internal/temporalcli/commands.gen.go) file from code, simply run:
 
-    go run ./internal/cmd/gen-commands
+    make gen
```
cretz: We support contributors on all of our supported platforms and therefore contributing guide should work across all platforms we support IMO

stephanos: Consider this a Freudian slip. Fixed.

cmd/gen-commands/main.go
stephanos: 👍 I'll change it to a file input, no strong opinions on my end

internal/cmd/gen-commands/main.go
stephanos: That's fair; since it is an internal tool only. 👍

cretz: I was saying they could be just commands on the existing binary (but hidden from "help"). Technically a completely different binary is fine too, but you lose the ability to easily download/version this binary.

gen-commands
cretz, line 1: Were these committed intentionally?

internal/cmd/gen-commands/main.go
stephanos: Oh, sorry, I misread. I have mixed feelings about including it directly in the Temporal CLI; it feels like it shouldn't be including the dev tooling itself.

AFAICS `go run github.com/temporalio/cli/internal/cmd/gen-commands@<version>` is how you'd run the command generator. And it would respond to the same git tags as the Temporal CLI as it's in the same repo.

Or should we consider moving it to a separate repo; now that we a standalone binary?

cretz: I'm ok w/ how you have it now, just a thought of making it a separate command gets all of the packaging/deployment benefits, but what you have here is fine

cretz (approved):

internal/cmd/gen-commands/main.go
stephanos: 👍 we can always make adjustments later once we some learnings.

PR #877, opened 2025-11-20
Automate Docker Image Build and Publish

chaptersix:
## What was changed
* added GHA to build and publish docker images
* Updated go releaser to v2
* Minor tweaks to other GHA

## Why?
Make is easier to publish docker images on release.


.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,162 @@
...
+          docker buildx bake \
+            --file .github/docker/docker-bake.hcl \
+            cli
+        env:
+          CLI_SHA: ${{ steps.meta.outputs.cli_sha }}
+          IMAGE_SHA_TAG: ${{ steps.meta.outputs.image_sha_tag }}
+          IMAGE_BRANCH_TAG: ${{ steps.meta.outputs.image_branch_tag }}
+          VERSION: ${{ steps.meta.outputs.version }}
+          TAG_LATEST: false
+          IMAGE_REPO: temporalio
```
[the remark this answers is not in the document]
chaptersix: /fp this is a re-useable workflow and has the permissions set where it's called.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,214 @@
...
+        id: meta
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_PUBLISH: ${{ inputs.publish }}
+          INPUT_REGISTRY: ${{ inputs.registry }}
+          INPUT_REGISTRY_NAMESPACE: ${{ inputs.registry_namespace }}
+          INPUT_IMAGE_NAME: ${{ inputs.image_name }}
+          REPO_OWNER: ${{ github.repository_owner }}
+        run: |
+          echo "cli_sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
```
chaptersix: This logic mostly existing to enable forks to publish and image. I can remove it if we don't need it.

.github/docker/cli.Dockerfile
```diff
@@ -0,0 +1,34 @@
+# syntax=docker/dockerfile:1
+
+ARG ALPINE_IMAGE
```
bergundy: Why does this need to be an argument?

.github/docker/cli.Dockerfile
bergundy, line 1: Can't we put the dockerfile in the root of the repo? and name it `Dockerfile`?

.github/docker/cli.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Copy passwd and group files for non-root user
+COPY --from=certs /etc/passwd /etc/passwd
+COPY --from=certs /etc/group /etc/group
+
+# Copy the appropriate binary for target architecture
+COPY --from=dist /dist/$TARGETARCH/temporal /temporal
+
+# Run as non-root user temporal (uid 1000)
+USER 1000:1000
```
bergundy: Does this work?

```suggestion
USER temporal:temporal
```

Makefile
```diff
@@ -9,3 +9,23 @@ internal/commands.gen.go: internal/commandsgen/commands.yml
 
 build:
 	go build ./cmd/temporal
+
+update-alpine: ## Update Alpine base image to latest version and digest (usage: make update-alpine [ALPINE_TAG=3.22])
```
bergundy: It was a mistake to add a Makefile to this repo in the first place. Let's not make the situation worse.

.goreleaser.yml
```diff
@@ -1,3 +1,5 @@
+version: 2
```
bergundy, line 1: Thanks!

.github/workflows/update-latest-tag.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: ${{ github.event.release.tag_name }}
+
+      - name: Check if release is latest
+        id: check_latest
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        uses: actions/github-script@v7
```
bergundy: Fancy!

.github/workflows/update-latest-tag.yml
```diff
@@ -0,0 +1,120 @@
...
+            console.log(`Should tag as latest: ${isLatest}`);
+
+      - name: Set up Docker Buildx
+        if: steps.check_latest.outputs.is_latest == 'true'
+        uses: docker/setup-buildx-action@v3
+
+      - name: Get registry configuration
+        if: steps.check_latest.outputs.is_latest == 'true'
+        id: registry
+        run: |
```
bergundy: Make this JS too, that's much nice than bash :)

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,214 @@
...
+    name: Build and Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
```
bergundy: Why fix this to a specific hash and not a major version to get automatic upgrades like we do everywhere else?

.github/docker/cli.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Copy CA certificates from certs stage
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+# Copy passwd and group files for non-root user
+COPY --from=certs /etc/passwd /etc/passwd
+COPY --from=certs /etc/group /etc/group
+
+# Copy the appropriate binary for target architecture
+COPY --from=dist /dist/$TARGETARCH/temporal /temporal
```
bergundy: Wouldn't it be easier to start from alpine and save all of these extra layers?

Dockerfile
```diff
@@ -1,11 +1,16 @@
...
+COPY dist/nix_linux_arm64_v8.0/temporal /dist/arm64/temporal
+
+FROM alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
 
-FROM alpine:3.22
 ARG TARGETARCH
+
 RUN apk add --no-cache ca-certificates
-COPY --from=dist /dist/$TARGETARCH/temporal /usr/local/bin/temporal
+COPY --chmod=755 --from=dist /dist/${TARGETARCH}/temporal /usr/local/bin/temporal
```
chaptersix:
```suggestion
COPY --from=dist /dist/${TARGETARCH}/temporal /usr/local/bin/temporal
```

chaptersix:
> Semgrep found **1** [`missing-explicit-permissions`](https://semgrep.dev/playground/r/yeTZ9lD/temporal.security.gha.missing-explicit-permissions?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule) finding:
> 
> * .github/workflows/trigger-docs.yml
>   
>   * [L8-55](https://github.com/temporalio/cli/blob/363f893470355b718d9cf9576fb5bd69b664e37a/.github/workflows/trigger-docs.yml#L8-55) - [Triage](https://semgrep.dev/orgs/temporal/findings/302399672)
> 
> No explicit `GITHUB_TOKEN` permissions found at the workflow or job level. Add a `permissions:` block at the workflow root (applies to all jobs) or per job with least privilege (e.g., `contents: read` and only specific writes like `pull-requests: write` if needed).

this is fixed

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+      registry_namespace:
+        description: "Registry namespace/organization"
+        required: false
+        type: string
+        default: ""
+      image_name:
+        description: "Image name"
+        required: false
+        type: string
+        default: "temporal"
```
bergundy: Wondering if this will actually be used but not opposed to having the flexibility if it is needed.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+      DOCKER_PASSWORD:
+        required: false
+
+jobs:
+  build:
+    name: Build and Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
bergundy: nit: there's a v5 already.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
bergundy: Looks like v6 is out already.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_PUBLISH: ${{ inputs.publish }}
+          INPUT_REGISTRY: ${{ inputs.registry }}
+          INPUT_REGISTRY_NAMESPACE: ${{ inputs.registry_namespace }}
+          INPUT_IMAGE_NAME: ${{ inputs.image_name }}
+          REPO_OWNER: ${{ github.repository_owner }}
+        uses: actions/github-script@v7
+        with:
+          script: |
```
bergundy: With a script this long, it'd be best not to inline the code here IMHO. Here's an alternative: https://github.com/actions/github-script?tab=readme-ov-file#run-a-separate-file

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+            }
+            core.setOutput('version', version);
+
+            // Determine registry (with auto-detection for temporalio vs forks)
+            let registry = inputRegistry;
+            if (!registry) {
+              if (repoOwner === 'temporalio') {
+                registry = 'docker.io';
+              } else {
+                registry = 'ghcr.io';
```
bergundy: Is this a requirement? Why even publish to `ghcr`?

chaptersix: I used it to test the release process in my fork. I can remove it.

.github/workflows/build-and-publish.yml
chaptersix:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/update-latest-tag.yml
```diff
@@ -0,0 +1,127 @@
...
+  contents: read
+  packages: write
+
+jobs:
+  update-latest:
+    name: Update Latest Tag
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
chaptersix:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/build-and-publish.yml
chaptersix:
```suggestion
        uses: actions/setup-go@v6
```

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+        run: echo "platform=$(go version | cut -d ' ' -f 4)" >> $GITHUB_OUTPUT
+
+      - name: Get Go version
+        id: go
+        run: echo "go=$(go version | cut -d ' ' -f 3)" >> $GITHUB_OUTPUT
+
+      - name: Check if release is latest
+        if: inputs.publish
+        id: check_latest_release
+        uses: actions/github-script@v7
```
chaptersix:
```suggestion
        uses: actions/github-script@v8
```

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+
+            const isLatest = !release.prerelease && !release.draft;
+            core.setOutput('is_latest', isLatest);
+            console.log(`Release: ${release.tag_name}`);
+            console.log(`Prerelease: ${release.prerelease}, Draft: ${release.draft}`);
+            console.log(`Should tag as latest: ${isLatest}`);
+
+      - name: Run GoReleaser (release)
+        if: inputs.publish
+        uses: goreleaser/goreleaser-action@v6.4.0
```
chaptersix:
```suggestion
        uses: goreleaser/goreleaser-action@v6
```

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BUILD_DATE: ${{ steps.date.outputs.date }}
+          BUILD_TS_UNIX: ${{ steps.timestamp.outputs.timestamp }}
+          GIT_BRANCH: ${{ steps.branch.outputs.branch }}
+          BUILD_PLATFORM: ${{ steps.platform.outputs.platform }}
+          GO_VERSION: ${{ steps.go.outputs.go }}
+
+      - name: Run GoReleaser (snapshot)
+        if: ${{ !inputs.publish }}
+        uses: goreleaser/goreleaser-action@v6.4.0
```
chaptersix:
```suggestion
        uses: goreleaser/goreleaser-action@v6
```

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,243 @@
...
+      - name: Get build metadata
+        id: meta
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_PUBLISH: ${{ inputs.publish }}
+          INPUT_REGISTRY: ${{ inputs.registry }}
+          INPUT_REGISTRY_NAMESPACE: ${{ inputs.registry_namespace }}
+          INPUT_IMAGE_NAME: ${{ inputs.image_name }}
+          REPO_OWNER: ${{ github.repository_owner }}
+        uses: actions/github-script@v7
```
chaptersix:
```suggestion
        uses: actions/github-script@v8
```

.github/workflows/update-latest-tag.yml
```diff
@@ -0,0 +1,127 @@
...
+        with:
+          ref: ${{ github.event.release.tag_name }}
+
+      - name: Check if release is latest
+        id: check_latest
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        uses: actions/github-script@v7
```
chaptersix:
```suggestion
        uses: actions/github-script@v8
```

.github/workflows/update-latest-tag.yml
```diff
@@ -0,0 +1,127 @@
...
+            console.log(`Should tag as latest: ${isLatest}`);
+
+      - name: Set up Docker Buildx
+        if: steps.check_latest.outputs.is_latest == 'true'
+        uses: docker/setup-buildx-action@v3
+
+      - name: Get registry configuration
+        if: steps.check_latest.outputs.is_latest == 'true'
+        id: registry
+        uses: actions/github-script@v7
```
chaptersix:
```suggestion
        uses: actions/github-script@v8
```

.github/workflows/build-and-publish.yml
chaptersix: greatly simplified it.

.github/workflows/build-and-publish.yml
chaptersix: Primary motivation was to use the same process for building images in PR for releases. Left a few more things configurable just in case.

chaptersix: We can set different permissions based on the trigger.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,191 @@
...
+          docker buildx bake \
+            --file docker-bake.hcl \
+            cli
+        env:
+          CLI_SHA: ${{ steps.meta.outputs.cli_sha }}
+          IMAGE_SHA_TAG: ${{ steps.meta.outputs.image_sha_tag }}
+          VERSION: ${{ steps.meta.outputs.version }}
+          TAG_LATEST: false
+          IMAGE_NAMESPACE: ${{ inputs.registry_namespace }}
+          IMAGE_NAME: ${{ steps.meta.outputs.image_name }}
```
chaptersix:
```suggestion
          IMAGE_NAME: ${{ inputs.image_name }}
```

.github/scripts/check-latest-tag.js
bergundy, line 1: Seems like it would be simpler to require an argument to decide whether to tag latest. That can be a separate step at the end of the release process once the release on GH is no longer marked as pre-release. We intentionally mark new releases as pre-release to allow goreleaser to build binaries first.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,191 @@
...
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_PUBLISH: ${{ inputs.publish }}
+          INPUT_REGISTRY_NAMESPACE: ${{ inputs.registry_namespace }}
+          INPUT_IMAGE_NAME: ${{ inputs.image_name }}
+          REPO_OWNER: ${{ github.repository_owner }}
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const inputVersion = process.env.INPUT_VERSION;
```
bergundy: It would have been easier to parse and also run locally if this was a script written in Go and we called `docker build` directly as a subprocess than using the bakefile and these CI scripts IMHO but I am definitely not going to block on that.

Dockerfile
```diff
@@ -1,11 +1,16 @@
-FROM --platform=$BUILDARCH scratch AS dist
-COPY ./dist/nix_linux_amd64_v1/temporal /dist/amd64/temporal
-COPY ./dist/nix_linux_arm64/temporal /dist/arm64/temporal
+# Intermediate stage to normalize goreleaser output paths
+# This copies both architecture binaries and renames them to clean paths,
+# allowing the final stage to select the correct binary using TARGETARCH
+FROM scratch AS dist
+COPY dist/nix_linux_amd64_v1/temporal /dist/amd64/temporal
+COPY dist/nix_linux_arm64_v8.0/temporal /dist/arm64/temporal
+
```
bergundy:
```suggestion
RUN cp /dist/${BUILDARCH}/temporal /dist/temporal

```

Dockerfile
```diff
@@ -1,11 +1,16 @@
...
+FROM alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
 
-FROM alpine:3.22
 ARG TARGETARCH
+
 RUN apk add --no-cache ca-certificates
-COPY --from=dist /dist/$TARGETARCH/temporal /usr/local/bin/temporal
+COPY --from=dist /dist/${TARGETARCH}/temporal /usr/local/bin/temporal
```
bergundy:
```suggestion
RUN apk add --no-cache ca-certificates tzdata
COPY --from=dist /dist/temporal /usr/local/bin/temporal
```

bergundy: Note that I added tzdata.

Dockerfile
bergundy: See also below.

bergundy (commented):
Overall LGTM. Didn't have anything major.

.github/scripts/check-latest-tag.js
chaptersix: This workflow listens to both the creation of the release and edits. The `release.yaml` workflow only listens to the creation. I tried to make things work for both the pre-release -> release and just release. In case someone messes up the order or the need to do pre-release -> release goes away. 

Also building the binaries with `goreleaser` takes quite a while on existing runners so I didn't want to rebuild unnecessarily. I prefer the binary and image building to run in PR and it look very similar to what happens during a release.

Dockerfile
chaptersix: Not sure I understand this change. build arch will always be the same and target arch will differ based on the platform the image is being built for. We need to select to proper binary from the dist stage.

bergundy: My thought was that the dist container would be arch specific and we could save the `TARGETARCH` build arg from the final image.

.github/scripts/check-latest-tag.js
bergundy: Just note that the check you have here won't work because we don't automatically mark a release as latest when we cut it.

Dockerfile
chaptersix: Sounds like the goal is to reduce the layer size. I don't believe the dist steps results in an additional layer. I'll test to see if arch specific dist step impact changes the total image size, layer count, or layer size.

.github/scripts/check-latest-tag.js
chaptersix:
```
            const isLatest = !release.prerelease && !release.draft;
            core.setOutput('is_latest', isLatest);
            console.log(`Release: ${release.tag_name}`);
            console.log(`Prerelease: ${release.prerelease}, Draft: ${release.draft}`);
            console.log(`Should tag as latest: ${isLatest}`);
```
     
   I misspoke, latest in this context is not pre-release and not draft. I see your point. This will not work. I'll update this to be a separate action that can be run after the manually.

chaptersix: done

Dockerfile
chaptersix: no impact, on the resulting image. Sent you some results

chaptersix:
just realized that this repo uses release candidates and those are not currently published to docker hub.

chaptersix:
moving the image build to the manually triggered workflow.

Dockerfile
chaptersix: ended up remove this

.github/workflows/build-and-publish-docker.yml
```diff
@@ -0,0 +1,123 @@
...
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_TAG_LATEST: ${{ inputs.tag_latest }}
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const inputVersion = process.env.INPUT_VERSION;
+            const inputTagLatest = process.env.INPUT_TAG_LATEST;
+
+            const version = inputVersion.startsWith('v') ? inputVersion.slice(1) : inputVersion;
+            const releaseTag = inputVersion.startsWith('v') ? inputVersion : `v${inputVersion}`;
```
bergundy, line 46: I would also be okay with requiring that the tag has the right format and failing the workflow.

chaptersix: The workflow effectively insures the proper format because the release has to exist for it to run. One of the complaints about the existing release process is different steps use `v` and others do not. We'd either need to accept `v` or not `v` or accept both due to the docker tag excluding `v` and the github release including `v`.

PR #881, opened 2025-12-03
Wire enableReplication into cli

michaely520:
## What was changed
Wiring in new enableReplication flag into cluster upsert cli.

Deleting sql pinning test as we no longer need this, see historical context:

https://github.com/temporalio/cli/pull/784
https://github.com/temporalio/temporal/pull/7333

then unpinning in OSS https://github.com/temporalio/temporal/pull/8489

## Why?

Allows for users to leverage the new field

## Checklist

1. Closes 

2. How was this tested:

built locally and tested in conjunction with https://github.com/temporalio/temporal/pull/8744, used the cli manually to make a replication based test pass

3. Any docs updates needed?


go.mod
```diff
@@ -10,21 +10,21 @@ require (
...
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/ui-server/v2 v2.42.1
-	go.temporal.io/api v1.53.0
+	go.temporal.io/api v1.58.1-0.20251203002935-472c572d9d48
 	go.temporal.io/sdk v1.37.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
-	go.temporal.io/server v1.29.1
+	go.temporal.io/server v1.30.0-147.0
```
cretz: This PR needs to target `next-server` branch IMO (unsure if has had `main` merged into it recently or if it needs to)

Issue #876, opened 2025-11-20
[Feature Request] Allow setting Temporal headers when starting workflow


cretz:
### Describe the solution you'd like

From user request at https://github.com/temporalio/features/issues/703.

PR #883, opened 2025-12-08
Enable setting of Temporal headers through cli

Tamoghnasen9:
## What was changed
Support setting of Temporal headers in workflows

## Why?
The cli doesn't support the setting of workflow headers. With this change we would be able to pass the headers to Workflows using cli.

## Checklist

1. Closes - https://github.com/temporalio/cli/issues/876

2. How was this tested: 
- Brought up the dev server using : `temporal server start-dev`
- Build the cli locally and passed the `--headers` flag to start a workflow. Checked in Temporal UI, the Header section shows the headers set in Workflow.

3. Any docs updates needed?
https://docs.temporal.io/cli/workflow


internal/temporalcli/commands.yaml
```diff
@@ -4464,6 +4464,12 @@ option-sets:
           May be passed multiple times to set multiple headers.
           Can also be made available via environment variable as
           `TEMPORAL_GRPC_META_[name]`.
+      - name: headers
```
cretz: I think this setting should only be on commands where it makes sense, namely workflow start, execute, query, signal, and update (and any signal-with-start or update-with-start).

Tamoghnasen9: Ok. Limited it to:
- workflow  execute
- workflow  execute-update-with-start
- workflow  signal
- workflow signal-with-start
- workflow  start
- workflow  start-update-with-start
- workflow query
- workflow update start
- workflow update execute

internal/temporalcli/commands.workflow.go
```diff
@@ -449,7 +463,21 @@ func workflowUpdateHelper(cctx *CommandContext,
...
+		newOpts := cctx.StoredClientOptions
+		cliHeaderPropagator, err := cliHeaderPropagator(updateStartOpts.Headers)
+		if err != nil {
+			return err
+		}
+		newOpts.ContextPropagators = append(newOpts.ContextPropagators, cliHeaderPropagator)
+		execClient, err = client.NewClientFromExisting(cl, newOpts)
+		if err != nil {
+			return fmt.Errorf("unable to create header-propagator client: %w", err)
+		}
```
cretz: I think instead of this approach where we store options and make a new client, I think we should have an always-set "header context propagator" that extracts the headers from a known key in the context, and just call something like `cl.UpdateWorkflow(ContextWithHeaders(updateStartOpts.Headers), ...` where `ContextWithHeaders` creates a context with that key if there are any headers.

internal/temporalcli/client.go
```diff
@@ -314,3 +318,44 @@ func (rawValuePayloadConverter) Encoding() string {
 	// Should never be used
 	return "raw-value-encoding"
 }
+
+type headerPropagator struct{}
+
+const cliHeaderContextKey = "worflow-headers"
```
cretz:
```suggestion
const cliHeaderContextKey = "workflow-headers"
```

Or better:

```suggestion
type cliHeaderContextKey struct{}
```

And use the empty struct when accessing

internal/temporalcli/client.go
```diff
@@ -314,3 +318,44 @@ func (rawValuePayloadConverter) Encoding() string {
...
+
+func (headerPropagator) Inject(ctx context.Context, writer workflow.HeaderWriter) error {
+	val := ctx.Value(cliHeaderContextKey)
+	if headers, ok := val.(map[string]string); ok {
+		for k, v := range headers {
+			writer.Set(k, payload.EncodeString(v))
```
cretz: Hrmm, in addition to not wanting to use `server` utilities in here, I think this may need to be raw data. A user should be able to provide `foo="var"` or `foo=1` and not have the latter become `foo="1"`. They should be able to provide `foo={"bar": "baz"}`. This is the same approach that is taken everywhere else that accepts payloads, like for `--search-attribute` and `--memo`, can we match those (i.e. use `stringKeysJSONValues`)?

Tamoghnasen9: Thanks, implemented this.

spkane31:
Can you add a test for this change? You should be able to do something like this (heavy pseudocode):

```go
res := s.Execute("workflow", "start", ..., "--headers", "someHeader=123",...)

s.Client.GetWorkflowHistory(...)

startedEvent := <first enums.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED eventType>

headers := startedEvent.GetWorkflowExecutionStartedEventAttributes().GetHeaders()

require.Equal(t, headers.Get("someHeader"), "123")
```

that's not perfect by any means but that's the gist of what a test for this should look like.

PR #887, opened 2025-12-12
Move gen commands

stephanos:
## What was changed

Moved `‎internal/cmd/gen-commands` to `cmd/gen-commands`; and same for `gen-docs`.

## Why?

When they are in `internal/` they cannot be invoked via `go run` from outside the repository; making it cumbersome to use.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,13 +60,13 @@ jobs:
       - name: Regen code, confirm unchanged
         if: ${{ matrix.checkGenCommands }}
         run: |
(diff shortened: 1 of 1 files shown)


cmd/gen-commands/README.md
cretz, line 1: FWIW, I think it's not the best that these internal tools are alongside the primary binary as an equal and would have liked to see us entertain an option of making these just undocumented commands inside the existing binary, but not a strong objection

PR #889, opened 2025-12-15
Invoke CLI extension

stephanos:
## What was changed

CLI now invokes extensions in accordance with [CLI Extensions proposal](https://github.com/temporalio/proposals/blob/master/cli/cli-extensions.md).

Note that the help command integration part is going to be in a follow-up PR.

## Why?

Allow Temporal CLI Extensions.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/temporalcli/commands.extension.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+)
+
+const (
+	// extensionPrefix is the required prefix for all Temporal CLI extensions.
+	extensionPrefix = "temporal-"
+)
+
+// lookupExtension searches for an extension executable from the given args.
+// Extension commands must come before any flags.
```
stephanos: This still is an active research question; but in order to move forward and gain more data we make a decision here. This can/will be revisited.

internal/temporalcli/commands.extension.go
```diff
@@ -0,0 +1,74 @@
...
+
+// tryExecuteExtension attempts to find and execute an extension for the given options.
+// Returns (err, true) if an extension was found and executed, or (nil, false) if no match.
+func tryExecuteExtension(ctx context.Context, opts *CommandOptions) (error, bool) {
+	extPath, extArgs := lookupExtension(opts.Args)
+	if extPath == "" {
+		return nil, false
+	}
+
+	cmd := exec.CommandContext(ctx, extPath, extArgs...)
```
stephanos: Interrupt signal handling was called out in the CLI extension proposal as an area of active search; here we decide to go with the option of aborting the sub-command on interrupt to unblock this work. This will be re-evaluated after gathering more data.

internal/temporalcli/commands.extension.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Args starting with "-" are skipped when building extension names.
+// Dashes within args are converted to underscores when building extension names.
+// For example, given args ["foo", "--flag=value", "bar-baz"], it tries:
+//
+//	temporal-foo-bar_baz, temporal-foo
+//
+// If temporal-foo-bar_baz is found, returns (path, ["--flag=value"]).
+// If temporal-foo is found, returns (path, ["--flag=value", "bar-baz"]).
+func lookupExtension(args []string) (path string, extArgs []string) {
```
stephanos: The CLI extension proposal describes the question of "whether flags of parent commands can still be placed before the extension command" as an area of active research. For now we require them to be placed after the extension command to disambiguate. This will be re-evaluated after gathering more data.

cretz: I think you should find the most applicable built-in sub-command, remove all common flags on that subcommand, then do your extension resolution (and pass along all of those flags to the subcommand). And you should fail if there are unrecognized parent/common flags before the subcommand. For instance, I would expect:

* `temporal --output json my-command --some-bool-arg` to run `temporal-something --output json --some-bool-arg`
* `temporal --no-json-shorthand-payloads my-command` to run `temporal-my-command --no-json-shorthand-payloads`
* `temporal --address 1.2.3.4:5678 workflow --tls diagram` to run `temporal-workflow-diagram --address 1.2.3.4:5678 --tls`
* `temporal workflow --some-bool-arg diagram` to fail with "unrecognized flag --some-bool-arg"
* `temporal --some-arg some-value workflow diagram` to fail with "unrecognized flag --some-arg"

I have not investigated how to do this yet or how much of it is possible, but you cannot comfortably know which flags accept values and which are standalone to do some kind of arbitrary parsing. May have to put a `ArbitraryArgs` on every command. I can help with a PoC of this if needed.

internal/temporalcli/commands.extension.go
cretz: This may be good enough. On nix-based platforms, I think we should consider tunnelling signals (e.g. sigint) and on context timeout, sending sigterm in this case instead of sigkill, but on Windows hard kill like this is acceptable. We will need to test I think.

PR #892, opened 2025-12-18
Add OAuth support for CLI extensions

stephanos:
## What was changed

Redo of https://github.com/temporalio/cli/pull/886.

Adding:

1. ability to code-gen just options-sets
2. ability to reference option-sets across codegen YAML input files
3. shared `CommonOptions` and `ClientOptions` generated for cliext
4. `ClientOptionsBuilder` for building client dial options
5. integration of OAuth token into client dial options (plus integration tests)
6. `StoreClientOAuthOptions` and `LoadClientOAuth` for oauth config
7. `NewLogger` helper for creating a new logger from flags

## Why?

To make OAuth a first-class authentication mechanism in the Temporal CLI and CLI extensions.

## Checklist

1. Closes 

8. How was this tested:

9. Any docs updates needed?

### DIFF
--- a/internal/temporalcli/commands.go
+++ b/internal/temporalcli/commands.go
@@ -83,7 +83,6 @@ type CommandOptions struct {
 	Fail func(error)
 
 	AdditionalClientGRPCDialOptions []grpc.DialOption
-	ClientConnectTimeout            time.Duration
 }
 
 type DeprecatedEnvConfig struct {
@@ -559,10 +558,6 @@ func (c *TemporalCommand) preRun(cctx *CommandContext) error {
 			fmt.Errorf("command timed out after %v", c.CommandTimeout.Duration()),
 		)
 	}
-	if c.ClientConnectTimeout.Duration() > 0 {
-		cctx.Options.ClientConnectTimeout = c.ClientConnectTimeout.Duration()
-	}
-
 	return nil
 }
 


cliext/client.go
```diff
@@ -0,0 +1,321 @@
...
+	EnvLookup envconfig.EnvLookup
+	// Logger is the slog logger to use for the client. If set, it will be
+	// wrapped with the SDK's structured logger adapter.
+	Logger *slog.Logger
+}
+
+// BuildClientOptions creates SDK client options from a ClientOptionsBuilder.
+// If OAuth is configured and no APIKey is set, OAuth will be used to obtain an access token.
+// Returns the client options and the resolved namespace (which may differ from input if loaded from profile).
+func BuildClientOptions(ctx context.Context, opts ClientOptionsBuilder) (client.Options, string, error) {
```
cretz: Pedantic, but If the second parameter isn't really an "options", might as well just put `Build` as a method on `*ClientOptionsBuilder`

cliext/client.go
```diff
@@ -0,0 +1,321 @@
...
+	// If nil, environment variables are not used for profile loading.
+	EnvLookup envconfig.EnvLookup
+	// Logger is the slog logger to use for the client. If set, it will be
+	// wrapped with the SDK's structured logger adapter.
+	Logger *slog.Logger
+}
+
+// BuildClientOptions creates SDK client options from a ClientOptionsBuilder.
+// If OAuth is configured and no APIKey is set, OAuth will be used to obtain an access token.
+// Returns the client options and the resolved namespace (which may differ from input if loaded from profile).
```
cretz: Isn't namespace already on the options?

cliext/client.oauth.go
```diff
@@ -0,0 +1,98 @@
...
+		return OAuthToken{}, fmt.Errorf("OAuth config is nil")
+	}
+	client := newOAuthClient(config.OAuthClientConfig)
+	return client.token(ctx, config)
+}
+
+// NewOAuthDynamicTokenProvider creates a function that provides OAuth access tokens dynamically.
+// The returned function loads OAuth configuration on-demand from the config file and refreshes
+// tokens as needed. Returns empty string if no OAuth is configured.
+func NewOAuthDynamicTokenProvider(opts ClientOptionsBuilder) func(context.Context) (string, error) {
```
cretz: Pedantic, but would also consider putting this as a method on the builder (don't even have to have it return a func, can just make the method the func, e.g. `func (c *ClientOptionsBuilder) GetOAuthToken(ctx context.Context) (string, error) {`, but see comment a couple of lines down above why we'd want some before-anon-func work)

cliext/client.oauth.go
```diff
@@ -0,0 +1,98 @@
...
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  c.options.AuthURL,
+			TokenURL: c.options.TokenURL,
+		},
+		Scopes: c.options.Scopes,
+	}
+}
+
+// token returns a valid access token, refreshing if necessary.
+func (c *oauthClient) token(ctx context.Context, config *OAuthConfig) (OAuthToken, error) {
```
cretz: I would have expected `oauth2` package to handle expiration, refresh, etc. Is this logic basically what `oauth2.ReuseTokenSourceWithExpiry` does? I wonder if we even need to provide config/token structs ourselves instead of reusing `*oauth2.Config` and `*oauth2.Token` (and documenting which values of those we support for load/store). If we do still want our oauth config structs, IMO we should offer a `NewTokenSource()` func on `*OAuthConfig`. This will help with some libraries that want the oauth data types, e.g. https://pkg.go.dev/google.golang.org/grpc/credentials/oauth#TokenSource.

cliext/client.oauth.go
```diff
@@ -0,0 +1,98 @@
...
+	client := newOAuthClient(config.OAuthClientConfig)
+	return client.token(ctx, config)
+}
+
+// NewOAuthDynamicTokenProvider creates a function that provides OAuth access tokens dynamically.
+// The returned function loads OAuth configuration on-demand from the config file and refreshes
+// tokens as needed. Returns empty string if no OAuth is configured.
+func NewOAuthDynamicTokenProvider(opts ClientOptionsBuilder) func(context.Context) (string, error) {
+	return func(ctx context.Context) (string, error) {
+		result, err := LoadClientOAuth(LoadClientOAuthOptions{
```
cretz: IMO we should not re-lookup oauth config on each token need, it should be bound to original the loaded configuration. This means long-running CLI commands do not get logged in as maybe another user if the file changes out from under. I'd move this loading part out of to before the `func`.

cliext/client.go
```diff
@@ -0,0 +1,321 @@
...
+	if cfg.Identity != "" {
+		clientOpts.Identity = cfg.Identity
+	}
+
+	// Set logger if provided.
+	if opts.Logger != nil {
+		clientOpts.Logger = log.NewStructuredLogger(opts.Logger)
+	}
+
+	// Set OAuth credentials if configured and no API key is set.
```
cretz: Not seeing where this is checking whether oauth is configured

cliext/flags.gen.go
```diff
@@ -0,0 +1,248 @@
...
+	f.Var(&v.CommandTimeout, "command-timeout", "The command execution timeout. 0s means no timeout.")
+	v.ClientConnectTimeout = 0
+	f.Var(&v.ClientConnectTimeout, "client-connect-timeout", "The client connection timeout. 0s means no timeout.")
+	f.StringVar(&v.Env, "env", "default", "Active environment name (`ENV`).")
+	ctx.BindFlagEnvVar(f.Lookup("env"), "TEMPORAL_ENV")
+	f.StringVar(&v.EnvFile, "env-file", "", "Path to environment settings file. Defaults to `$HOME/.config/temporalio/temporal.yaml`.")
+	f.StringVar(&v.ConfigFile, "config-file", "", "File path to read TOML config from, defaults to `$CONFIG_PATH/temporal/temporal.toml` where `$CONFIG_PATH` is defined as `$HOME/.config` on Unix, `$HOME/Library/Application Support` on macOS, and `%AppData%` on Windows. EXPERIMENTAL.")
+	ctx.BindFlagEnvVar(f.Lookup("config-file"), "TEMPORAL_CONFIG_FILE")
+	f.StringVar(&v.Profile, "profile", "", "Profile to use for config file. EXPERIMENTAL.")
+	ctx.BindFlagEnvVar(f.Lookup("profile"), "TEMPORAL_PROFILE")
```
cretz: I think it's clearer if the `--config-file` and `--profile` flags are not bound to env vars on the flags themselves. IMO they should default to unset and let `envconfig` resolve them at runtime as the true canonical place for knowing what these are based on environment (in case we want to change that some day).

If a programmatic user needs to know the defaults of these values when the flag is not otherwise specified, we can provide helpers if what we have is not good enough already.

Also, `BindFlagEnvVar` is a relic of the `env` days and since envconfig came about in #764, `BindFlagEnvVar` was reduced to only applying for `env`. But now that we are exporting some of this functionality, I would be ok just hardcoding the `TEMPORAL_ENV` env var mapping somewhere and removing the concept of `optionContext` altogether. And then, at that point, we might as well remove support for the `env:` in the YAML file.

cliext/client.oauth.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// Check if token is still valid: not expired and not expiring within next minute.
+	if !config.AccessTokenExpiresAt.IsZero() && time.Now().Before(config.AccessTokenExpiresAt.Add(-1*time.Minute)) {
+		return config.OAuthToken, nil
+	}
+
+	// Token is expired or about to expire, refresh it.
+	cfg := c.createOAuth2Config("")
+	tokenSource := cfg.TokenSource(ctx, &oauth2.Token{RefreshToken: config.RefreshToken})
+	newToken, err := tokenSource.Token()
```
cretz: I am curious on this, do we think we should be re-persisting this refreshed token to disk? What does `tcld` do?

cliext/client.oauth.go
stephanos: Good call! I must have missed `ReuseTokenSourceWithExpiry`.

re `OAuthClientConfig` and `OAuthToken` have the benefit of hiding the `oauth2` as an implementation detail.  Another benefit is that they allow attaching `RequestParams` which we need for the Cloud CLI (to provide the `audience`).

But I can also see how `NewTokenSource` can be useful for integration as you point out. I don't know what weighs more, the benefit of hiding oauth2 as an impl detail or the convenience of exposing it via `NewTokenSource()`.

cliext/flags.gen.go
stephanos: 👍 on `--config-file` and `--profile`
👍 moved out `TEMPORAL_ENV` and removed `optionContext`

Regarding `implied-env`; is that actually used anywhere?

internal/temporalcli/commands.go
stephanos, line 590: ^ all moved to `cliext/logger.go`

cliext/client.go
```diff
@@ -0,0 +1,306 @@
...
+	// Initialize OAuth token source if no API key is set.
+	if cfg.ApiKey == "" {
+		if err := b.initOAuthTokenSource(ctx); err != nil {
+			return client.Options{}, fmt.Errorf("failed to initialize OAuth: %w", err)
+		}
+		// Only set credentials if OAuth is configured
+		if b.oauthTokenSource != nil {
+			clientOpts.Credentials = client.NewAPIKeyDynamicCredentials(b.getOAuthToken)
+		}
+	}
```
stephanos: Section where we integrate OAuth into the code that was moved here from temporalcli/client.go

internal/temporalcli/commands.go
```diff
@@ -440,6 +425,9 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 func (c *TemporalCommand) initCommand(cctx *CommandContext) {
 	c.Command.Version = VersionString()
 
+	// Bind TEMPORAL_ENV environment variable to --env flag
+	cctx.BindFlagEnvVar(c.Command.PersistentFlags().Lookup("env"), "TEMPORAL_ENV")
```
stephanos, line 444: Now manually binding `TEMPORAL_ENV` here.

internal/commandsgen/code.go
```diff
@@ -486,9 +501,6 @@ func (o *Option) writeFlagBuilding(selfVar, flagVar string, w *codeWriter) error
 	if o.Required {
 		w.writeLinef("_ = %v.MarkFlagRequired(%v, %q)", w.importCobra(), flagVar, o.Name)
 	}
-	if o.Env != "" {
-		w.writeLinef("cctx.BindFlagEnvVar(%v.Lookup(%q), %q)", flagVar, o.Name, o.Env)
-	}
```
stephanos, line 491: Removed

internal/commandsgen/parse.go
```diff
@@ -228,7 +260,7 @@ func (o *Option) processSection() error {
 		return fmt.Errorf("description should end in a '.'")
 	}
 
-	if o.Env != strings.ToUpper(o.Env) {
+	if o.ImpliedEnv != strings.ToUpper(o.ImpliedEnv) {
 		return fmt.Errorf("env variables must be in all caps")
 	}
```
stephanos, line 260: Changed over to `ImpliedEnv`.

cliext/client.oauth.go
cretz: > re OAuthClientConfig and OAuthToken have the benefit of hiding the oauth2 as an implementation detail

Why hide that implementation detail? Not sure any other Go library using OAuth would (e.g. the Go gRPC library). I see code was updated, so not easily seeing where request params is used in current PR, but would have to see how other Go OAuth clients are setting such request params. Definitely can have something like:

```go
type OAuthConfig struct {
	ClientConfig *oauth2.Config
	Token *oauth2.Token
	ClientRequestParams map[string]string
}
```

cliext/client.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// Set logger if provided.
+	if b.Logger != nil {
+		clientOpts.Logger = log.NewStructuredLogger(b.Logger)
+	}
+
+	// Attempt to configure OAuth config if no API key is set.
+	if cfg.ApiKey == "" {
+		result, err := LoadClientOAuth(LoadClientOAuthOptions{
```
cretz, line 207: I don't think this properly accounts for `if !common.DisableConfigFile || !common.DisableConfigEnv` that the other envconfig loading does above.

Would be ok if we wanted to go ahead and have Go SDK's env config support "additional data" to avoid this double-load for everyone that is doing basic no-auth CLI calls (happens commonly in cases like proxies). But also ok if we want to double load while we wait for that.

cliext/flags.gen.go
cretz: > Regarding implied-env; is that actually used anywhere?

That's the only one that's used, it's `env:` that's basically not used anymore except for the one `TEMPORAL_ENV`, so now we can probably remove `env:`. FWIW, `env:` used to be used in lots of places until envconfig came about and basically took over environment variables except for this one legacy one.

cliext/option-sets.yaml
```diff
@@ -0,0 +1,195 @@
...
+        type: string
+        description: Active environment name (`ENV`).
+        default: default
+        implied-env: TEMPORAL_ENV
+      - name: env-file
+        type: string
+        description: |
+          Path to environment settings file.
+          Defaults to `$HOME/.config/temporalio/temporal.yaml`.
+      - name: config-file
```
cretz, line 19: Where did `implied-env` go for `config-file` and `profile`? `implied-env` basically is just documentation/usage help, it has no runtime component.

cretz (commented):
Looking good, I think it's only detail stuff now

cliext/client.oauth.go
stephanos: 👍 okay, let's expose `oauth2`, I agree it is simpler and seems to really be _the_ library everyone uses anyway.

re request params: since they are not needed for token refresh, it's just a config field as far as `cliext ` is concerned.

cretz: But where are the request params used by cliext code?

cliext/client.oauth.go
stephanos: Good callout; the original PR I closed wrote the new access token to disk. It got lost in the reshuffle. I'll need to update that.

cliext/flags.gen.go
stephanos: Right; I've removed `env:` but I can't see where `ImpliedEnv` is actually _used_. I can't see it. My AI coding agent suggested appending it to flag descriptions; I didn't commit it here but it seemed like a good idea.

cliext/client.oauth.go
stephanos: Right; I also just realized that since AFAIK the `audience` request param is only required during login, we could let Cloud CLI (and other CLIs) deal with that themselves. I still had it here because originally I thought it was required for token refresh or the CLI's auto-login - but both of these aren't true (anymore). I'll remove it.

cliext/option-sets.yaml
stephanos: 👍 it got lost when I moved it, added back

cliext/client.go
stephanos: 👍 Absolutely; adding that check now. (just for `DisableConfigFile` since the OAuth cannot be configured via env vars)

cliext/client.oauth.go
cretz: Does `tcld` re-persist?

stephanos: Yes, it does: https://github.com/temporalio/tcld/blob/ab73fcc36aede9a87c0ba2a79b4e743a4ef6932b/app/token_config.go#L119-L123

cliext/flags.gen.go
cretz: It's should be used for docs/help only, it has no runtime behavior

cliext/client.go
```diff
@@ -0,0 +1,318 @@
...
+	EnvLookup envconfig.EnvLookup
+	// Logger is the slog logger to use for the client. If set, it will be
+	// wrapped with the SDK's structured logger adapter.
+	Logger *slog.Logger
+	// oauthConfig is initialized during Build() if OAuth is configured.
+	oauthConfig *OAuthConfig
+	// oauthConfigFilePath is the path where OAuth config is stored.
+	oauthConfigFilePath string
+	// oauthProfileName is the profile name where OAuth config is stored.
+	oauthProfileName string
```
cretz: A bit pedantic, but I'd make the stateful thing that has these 3 private fields in a separate private struct (even if it has to put this builder struct as a field) and have `getOAuthToken` on that. This shows up in Godoc as "has hidden fields" and therefore discourages users from creating the struct directly, but we want this builder struct IMO to look like it's all public fields to be created directly.

cliext/config.oauth.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return LoadClientOAuthResult{
+		OAuth:          oauthByProfile[profileName],
+		ConfigFilePath: configFilePath,
+		ProfileName:    profileName,
+	}, nil
+}
+
+// loadOAuthConfigFromFile loads OAuth configurations for all profiles from a TOML file.
+func loadOAuthConfigFromFile(path string) (map[string]*OAuthConfig, error) {
```
cretz, line 78: I still think we should consider just having an "additional data" concept on https://pkg.go.dev/go.temporal.io/sdk/contrib/envconfig. There's a lot of logic copied here around config file and env var stuff.

stephanos: I agree; I have filed a ticket for that last time we chatted about it. And I intend to pick that up very soon.

cliext/client.go
stephanos: 👍 Good point

PR #893, opened 2025-12-19
fix `hidden` flag property

stephanos:
## What was changed

Support `hidden: true` flag. It has no effect right now.

## Why?

While there isn't a use case anymore in the CLI, I was told for the Cloud CLI we want to add Temporal-specific flags for test environments that we don't want to expose to the end user in help texts.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/temporalcli/commands.yaml
```diff
@@ -3475,7 +3475,6 @@ commands:
       - name: queue-rps-limit-reason
         type: string
         description: Reason for queue rate limit update.
-        hidden: true
```
stephanos, line 3377: Checked with the team and this is not actually desirable.

internal/temporalcli/commands.yaml
cretz, line 1: The large set of comments at the top of this file was supposed to describe all fields this file could take, but I guess it is getting stale. Regardless, I think we should add `hidden` there. Not blocking though.

cretz (approved):

internal/temporalcli/commands.yaml
stephanos: Good call; fixed!

PR #899, opened 2026-01-06
Show pause information in "workflow describe" command

gow:
## What was changed

`temporal workflow describe` now shows pause information if the workflow is paused.

## Why?
When a workflow is paused we need to show the relevant pause information like paused time, identity & reason.

## Checklist

1. Closes 

2. How was this tested:
Tested manually

3. Any docs updates needed?
N/A

### DIFF
--- a/temporalcli/commands.workflow_pause.go
+++ b/temporalcli/commands.workflow_pause.go
@@ -16,6 +16,7 @@ func (c *TemporalWorkflowPauseCommand) run(cctx *CommandContext, args []string) 
 		WorkflowId: c.WorkflowId,
 		RunId:      c.RunId,
 		Identity:   c.Parent.Identity,
+		Reason:     c.Reason,
 	})
 	if err != nil {
 		return err


temporalcli/commands.workflow_pause.go
```diff
@@ -13,10 +13,10 @@ func (c *TemporalWorkflowPauseCommand) run(cctx *CommandContext, args []string)
 
 	_, err = cl.WorkflowService().PauseWorkflowExecution(cctx, &workflowservice.PauseWorkflowExecutionRequest{
 		Namespace:  c.Parent.Namespace,
-		Reason:     c.Reason,
 		WorkflowId: c.WorkflowId,
 		RunId:      c.RunId,
 		Identity:   c.Parent.Identity,
+		Reason:     c.Reason,
```
spkane31, line 19: could revert but not required

gow: Sorry I forgot to add a note. This is needed to fix a bug. I was not capturing the reason in pause command.

spkane31: Why's that? This looks like it just moves the reason to a different line

gow: oh! yeah. That explains it. I might have accidentally removed this line locally and ran into the bug. And later I added it thinking I fixed it 🤦

spkane31: not worth making the change, just pointed it out. no biggie

PR #903, opened 2026-01-12
Suppress logger warnings during start-dev shutdown

dandavison:
## What was changed
Use `slog.LevelVar` to suppress logger warnings during start-dev shutdown if the user has not set the log level.

## Why?
The warnings during shutdown are expected and damage UX.

## Checklist
2. How was this tested:
Built and ran manually. Confirmed `--log-level` was respected. Issued ctrl-c and confirmed no shutdown warnings on `server start-dev` but shutdown warnings on `server start-dev --log-level warn`

### DIFF
--- a/internal/temporalcli/commands.server.go
+++ b/internal/temporalcli/commands.server.go
@@ -166,6 +166,7 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 	cctx.Printer.Printlnf("%-8s http://%v:%v/metrics", "Metrics:", toFriendlyIp(opts.FrontendIP), opts.MetricsPort)
 	<-cctx.Done()
 	if !t.Parent.Parent.LogLevel.ChangedFromDefault {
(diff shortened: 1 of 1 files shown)


internal/temporalcli/commands.server.go
```diff
@@ -165,7 +165,9 @@ func (t *TemporalServerStartDevCommand) run(cctx *CommandContext, args []string)
 	}
 	cctx.Printer.Printlnf("%-8s http://%v:%v/metrics", "Metrics:", toFriendlyIp(opts.FrontendIP), opts.MetricsPort)
 	<-cctx.Done()
-	cctx.Printer.Println("Stopping server...")
+	if !t.Parent.Parent.LogLevel.ChangedFromDefault {
+		s.SuppressWarnings()
+	}
```
stephanos, line 171: Maybe a short comment to explain why we're doing this? People might forget/not know how verbose the logs are on shutdown.

stephanos (approved):

PR #905, opened 2026-01-14
Create versions before setting override, and require Deployment Name for Pinned Override Version

carlydf:
## What was changed
- Create versions before using them in a Pinned Versioning Override
- Require Deployment Name and Build ID flags to be set if override is Pinned. I'm not sure how this worked before, but I don't think we need to support optional deployment name.

## Why?
So that CLI works with latest version of server.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/temporalcli/commands.workflow.go
+++ b/internal/temporalcli/commands.workflow.go
@@ -112,7 +112,7 @@ func (c *TemporalWorkflowUpdateOptionsCommand) run(cctx *CommandContext, args []
 	}
 
 	if c.VersioningOverrideBehavior.Value == "pinned" {
-		if c.VersioningOverrideDeploymentName == "" && c.VersioningOverrideBuildId == "" {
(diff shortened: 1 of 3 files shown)


internal/temporalcli/commands.workflow.go
```diff
@@ -112,7 +112,7 @@ func (c *TemporalWorkflowUpdateOptionsCommand) run(cctx *CommandContext, args []
 	}
 
 	if c.VersioningOverrideBehavior.Value == "pinned" {
-		if c.VersioningOverrideDeploymentName == "" && c.VersioningOverrideBuildId == "" {
+		if c.VersioningOverrideDeploymentName == "" || c.VersioningOverrideBuildId == "" {
```
carlydf, line 115: the error message and the command help text make me think that requiring both flags was the intention all along.

carlydf: we definitely should not allow deployment name without a build id, which is what the `&&` condition allows

PR #908, opened 2026-01-21
Update server, API, and SDK to OSS v1.30.0 release candidate tags

carlydf:
## What was changed
Update server to v1.30.0-148.2
Update API to v1.60.0
Update SDK accordingly
Change BuildId -> BuildID for new SDK
Cherry-pick PR from main where I fixed versioning override tests so that they would work with the new server, which requires that a version exists before overriding a workflow to that version.

## Why?
To make way for other PRs for Versioning GA that require the latest server version.

## Checklist

1. Closes 

2. How was this tested:
Built locally
If CI tests pass I think this is ok

3. Any docs updates needed?

### DIFF
--- a/temporalcli/commands.workflow_reset_update_options_test.go
+++ b/temporalcli/commands.workflow_reset_update_options_test.go
@@ -292,6 +292,10 @@ func (s *SharedServerSuite) TestWorkflow_ResetBatchWithWorkflowUpdateOptions_Pin
 		assert.NoError(t, res.Err)
 	}, 30*time.Second, 100*time.Millisecond)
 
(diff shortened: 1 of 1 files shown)


cretz:
>> Cherry-pick PR from main where I fixed versioning override tests so that they would work with the new server, which requires that a version exists before overriding a workflow to that version.

> Would suggest merging `main` into `next-server` (understood if a bit difficult due to large `main` changes wrt `cliext` and such, but valuable to do anyways IMO)

I agree, but ACT team is going to handle that so I don't have to be blocked on it.

PR #901, opened 2026-01-09
[DO NOT MERGE] Use envconfig AdditionalProfileFields

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

PR #913, opened 2026-01-23
Tweak OAuth behavior

stephanos:
## What was changed

(1) Ability to update the `address` env profile field when storing OAuth config. 
(2) Ability to use `${namespace}` in `address` to refer to namespace at runtime. 
(3) Show helpful error message to user in case they have OAuth but still use `default` namespace.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cliext/config.oauth.go
```diff
@@ -215,6 +217,21 @@ func StoreClientOAuth(opts StoreClientOAuthOptions) error {
 		return err
 	}
 
+	// Set address if provided.
+	if opts.Address != "" {
```
stephanos, line 221: This is awkward, but will become much simpler once https://github.com/temporalio/cli/pull/901 is merged.

cliext/client.go
```diff
@@ -214,6 +217,11 @@ func (b *ClientOptionsBuilder) Build(ctx context.Context) (client.Options, error
 		}
 		// Only set credentials if OAuth is configured with an access token
 		if result.OAuth != nil && result.OAuth.Token != nil && result.OAuth.Token.AccessToken != "" {
+			// Error if OAuth is configured but namespace is still the default to show a clearer error message.
+			if clientOpts.Namespace == "default" {
+				return client.Options{}, fmt.Errorf(
+					"Please specify a namespace with `--namespace` or configure one in your profile.")
```
stephanos: Not seeing a lot of other error messages to draw the right tone from here. It seems usually they are very technical; maybe I should drop the "please". Not sure.

cliext/client.go
```diff
@@ -109,6 +109,9 @@ func (b *ClientOptionsBuilder) Build(ctx context.Context) (client.Options, error
 		profile.Namespace = cfg.Namespace
 	}
 
+	// Replace templated namespace in address, if present.
+	profile.Address = strings.ReplaceAll(profile.Address, "${namespace}", profile.Namespace)
```
cretz: If present I think should mean that there is no doing this if either address _or_ namespace are empty

cliext/client.go
cretz: Meh, either way is fine, though we usually lowercase w/out the please, following traditional Go error expectations

cliext/client.go
```diff
@@ -214,6 +217,11 @@ func (b *ClientOptionsBuilder) Build(ctx context.Context) (client.Options, error
 		}
 		// Only set credentials if OAuth is configured with an access token
 		if result.OAuth != nil && result.OAuth.Token != nil && result.OAuth.Token.AccessToken != "" {
+			// Error if OAuth is configured but namespace is still the default to show a clearer error message.
+			if clientOpts.Namespace == "default" {
```
cretz: This is a bit confusing, but in general, I think it's a clearer rule to check whether the `${namespace}` replacer is in the address, and if namespace is not present, error. Note, I would never expect this to be `default`, I think we very intentionally left the default up to the dial call.

stephanos: I thought generally using OAuth + `default` namespace never makes sense but I can scope it to that, too, to be safe.

stephanos: Isn't the default being set there by the CLI parsing?
```
	f.StringVarP(&v.Namespace, "namespace", "n", "default", "Temporal Service Namespace.")
```
ie before this line is executed.

cliext/client.go
stephanos: Since the CLI has a default for `namespace`, I don't think it can be empty?

And if the `address` does not contain `${namespace}` (or is empty) it's a noop anyway, right?

cliext/client.go
cretz: > I thought generally using OAuth + default namespace never makes sense

_Technically_ a user may use this for their OAuth on a proxy, but can't see why they'd want this magic `${namespace}` replacer then though

> Isn't the default being set there by the CLI parsing?

Ah, indeed it is, my mistake (it's on the Go side where we don't default eagerly). So yeah, feel free to assume `default` means unset from the POV of that replacer (and we need it set, at least for now)

cliext/client.go
cretz: True, per our conversation later, we will want to error if a replacer is set and namespace is default though

cliext/client.go
stephanos: I re-discovered `cfg.FlagSet.Changed("namespace")` and will use that instead of hard-coding it.

stephanos: Okay, I changed the semantics away from "is literally default" to "has not been explicitly set". The only loophole I can see is if the env profile is set to "default" explicitly (or the flag fwiw). Then this error might be confusing. Not sure how likely that is. If that's a concern, I could change it back to check for literally "default".

PR #916, opened 2026-01-26
Update Alpine Version

chaptersix:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
+FROM alpine:3.23
```
chaptersix, line 1: using digest pinning in the server repo caused docker to select the wrong platform.

PR #920, opened 2026-01-28
Rename `update-metadata-version` command to `update-version-metadata`

ShahabT:
## What was changed
WISOTT

## Why?
The old name is misleading because the command changes the metadata of a version, not the version of the metadata!

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/temporalcli/commands.yaml
```diff
@@ -2535,13 +2532,11 @@ commands:
     docs:
       description-header: >-
         Temporal Task Queue commands facilitate operations like describing
-        poller info, displaying partitions, fetching compatible Build IDs,
-        and determining Build ID reachability for effective Workflow and
-        Activity management.
+        poller info, backlog stats, and versioning info.
```
carlydf: did you mean to include all these changes to `taskqueue describe` in this PR? It seems like maybe the descriptions are changing but not the actual code.

Shivs11: +1 to this

PR #924, opened 2026-01-30
enable the system callback url

chaptersix:
## What was changed
- enabled routing of nexus callback via system url 

## Why?
https://github.com/temporalio/temporal/pull/8308

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/devserver/server.go
+++ b/internal/devserver/server.go
@@ -53,7 +53,6 @@ import (
 	"go.temporal.io/server/temporal"
 	"google.golang.org/grpc"
 	"gopkg.in/yaml.v3"
-	"go.temporal.io/server/components/callbacks"
 )
 
 const (
@@ -245,7 +244,6 @@ func (s *StartOptions) buildServerOptions() ([]temporal.ServerOption, *slog.Leve
 	// TODO: Remove this when upgrading to server 1.31.
 	dynConf[nexusoperations.UseSystemCallbackURL.Key()] = true
 
-	dynConf[callbacks.AllowedAddresses.Key()] = true
 	// Dynamic config if set
 	for k, v := range s.DynamicConfigValues {
 		dynConf[dynamicconfig.MakeKey(k)] = v


internal/devserver/server.go
```diff
@@ -241,7 +241,10 @@ func (s *StartOptions) buildServerOptions() ([]temporal.ServerOption, *slog.Leve
 	dynConf[dynamicconfig.HistoryCacheHostLevelMaxSize.Key()] = 8096
 	// Up default visibility RPS
 	dynConf[dynamicconfig.FrontendMaxNamespaceVisibilityRPSPerInstance.Key()] = 100
+	// Enable the system callback URL for worker targets.
 	// NOTE that the URL scheme is fixed to HTTP since the dev server doesn't support TLS at the time of writing.
+	// The callback URL template is used as a fallback for external targets.
+	dynConf[nexusoperations.UseSystemCallbackURL.Key()] = true
 	dynConf[nexusoperations.CallbackURLTemplate.Key()] = fmt.Sprintf(
 		"http://%s:%d/namespaces/{{.NamespaceName}}/nexus/callback", MaybeEscapeIPv6(s.FrontendIP), s.FrontendHTTPPort)
```
chaptersix: should we remove the template? Customers can optionally use persistence. So they could upgrade the cli and have nexus callbacks using this template.

internal/devserver/server.go
```diff
@@ -241,7 +241,10 @@ func (s *StartOptions) buildServerOptions() ([]temporal.ServerOption, *slog.Leve
 	dynConf[dynamicconfig.HistoryCacheHostLevelMaxSize.Key()] = 8096
 	// Up default visibility RPS
 	dynConf[dynamicconfig.FrontendMaxNamespaceVisibilityRPSPerInstance.Key()] = 100
+	// Enable the system callback URL for worker targets.
 	// NOTE that the URL scheme is fixed to HTTP since the dev server doesn't support TLS at the time of writing.
+	// The callback URL template is used as a fallback for external targets.
+	dynConf[nexusoperations.UseSystemCallbackURL.Key()] = true
```
cretz, line 245: I am curious about these Nexus dyn configs in general. Why can't a user be asked to configure these dynamic configs the same way they would if configuring any other self-hosted server? Why does other self-hosted environment not have these defaults and why must this self-hosted environment uniquely have these defaults here?

The caching and RPC limit updates make sense for local dev server, but feature defaults to not IMO. Our self-hosted dyn config requirements should be the same regardless of whether it's dev server self-hosted, docker self-hosted, binary self-hosted, etc.

chaptersix: OSS server v1.31 will have this toggle enabled by default and we can remove the URL templates here. 

After v1.31 no dc overrides are required for nexus.

internal/devserver/server.go
chaptersix: I'm going ahead and removing the templates.

internal/devserver/server.go
chaptersix: I just removed 2 other DCs.

internal/devserver/server.go
```diff
@@ -241,19 +241,10 @@ func (s *StartOptions) buildServerOptions() ([]temporal.ServerOption, *slog.Leve
...
-	dynConf[callbacks.AllowedAddresses.Key()] = []struct {
-		Pattern       string
-		AllowInsecure bool
-	}{
-		{
-			Pattern:       fmt.Sprintf("%s:%d", MaybeEscapeIPv6(s.FrontendIP), s.FrontendHTTPPort),
-			AllowInsecure: true,
-		},
-	}
+	// Enable the system callback URL for worker targets.
```
bergundy, line 243:
```suggestion
	// Enable the system callback URL for worker targets.
	// TODO: Remove this when upgrading to server 1.31.
```

internal/devserver/server.go
bergundy: We didn't want users to have to struggle to enable Nexus. That's why these configs were here in the first place. Starting server 1.31, Nexus will work OOTB with no dynamic config overrides.

bergundy: Also note that since the HTTP port is randomly allocated by default (still considered experimental), it was deemed to difficult to configure this for a user.

cretz: > We didn't want users to have to struggle to enable Nexus

But we clearly _did_ want that when we chose the general Temporal server defaults. Users should get consistent feature opt-in/default behavior across all forms of self-hosted server (whether dev server, docker container, helm chart, release downloaded from https://github.com/temporalio/temporal/releases, etc).

chaptersix: FWIW, the behavior of nexus callbacks is the same before and after this PR. Before, callbacks could only point to the dev server and after they could only point to the dev server. 

After 1.31 this config will be inline with the default config of the normal server release.  There's a TODO to remove these configs entirely.

cretz: > After 1.31 this config will be inline with the default config of the normal server release

We should wait then IMO. This CLI is for 1.30 not 1.31, and I would expect the same dyn configs to be required for users of a feeature across all releases of 1.30 (CLI, temporal release binary, docker image, helm chart, etc)

bergundy: Just want to clarify that this isn't for enabling Nexus, it's enabled by default. This is for making Nexus easier to use in the dev server. There is no need to walk back the decision we made a while ago to make Nexus work without extra user configuration.

This config is more for QOL, similar to `HistoryCacheHostLevelMaxSize` or `FrontendMaxNamespaceVisibilityRPSPerInstance`.

cretz: > This is for making Nexus easier to use in the dev server. There is no need to walk back the decision we made a while ago to make Nexus work without extra user configuration.

Nexus should be made easier to use in all 1.30 servers. If we decided not to do that for 1.30 servers, makes sense to do it for 1.31 servers. Not sure how that is related to CLI, I assume your decision to make Nexus work without user configuration is not CLI specific.

> This config is more for QOL, similar to HistoryCacheHostLevelMaxSize or FrontendMaxNamespaceVisibilityRPSPerInstance.

Would users need to set this dyn config for all the other non-CLI 1.30 servers they set up to have teh same user/operator experience? Or is it an internal-only/non-user-facing optimization?

chaptersix: > Would users need to set this dyn config for all the other non-CLI 1.30 servers they set up to have teh same user/operator experience? 

Yes, they would have to set the templating for pre-1.30 versions

PR #925, opened 2026-01-30
bump go version

chaptersix:
## What was changed
bump go to 1.25.5

## Why?
To be inline with server 1.30 and close CVEs

### DIFF
--- a/cliext/go.mod
+++ b/cliext/go.mod
@@ -1,6 +1,6 @@
 module github.com/temporalio/cli/cliext
 
-go 1.25.3
+go 1.25.5
 
 require (
 	github.com/BurntSushi/toml v1.4.0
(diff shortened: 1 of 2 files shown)


cliext/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/temporalio/cli/cliext
 
-go 1.25.0
+go 1.25.3
```
chaptersix:
```suggestion
go 1.25.5
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/temporalio/cli
 
-go 1.25.0
+go 1.25.3
```
chaptersix:
```suggestion
go 1.25.5
```

PR #928, opened 2026-02-02
Fix schedule update removing cron/spec when not provided (fixes #927)

### LINKED ISSUE #927: [Bug] schedule update removing cron if not explicitly included

Update a schedule.

### Describe the bug

The following command results in the removal of the existing cron configuration from the schedule:
```
temporal schedule update 
--env prod 
--schedule-id "Start Invoice Processing - Daily at 09:00 UTC (01:00 PST / 02:00 PDT)" 
--overlap-policy AllowAll 
--type start-invoice-processing 
--task-queue invoices-queue
```

Adding the --cron parameter resolves the issue and preserves the schedule configuration:
`--cron "0 9 * * *"`

billrich2001:
## What was changed
- **Describe before update**: Fetch current schedule via `Describe` and merge CLI options with existing spec.
- **Preserve when not provided**:
  - **Cron/calendar**: Preserve `CronExpressions` and `Calendars` when `--cron` and `--calendar` are not passed.
  - **Intervals**: Preserve `Intervals` when `--interval` is not passed.
  - **Jitter, TimeZoneName, StartAt, EndAt**: Preserve when the corresponding flags are not set (using `FlagSet.Changed` for fully incremental update).
- **`toScheduleSpecForUpdate(spec, existing)`**: New merge logic; create path still uses `toScheduleSpec` (existing == nil).

## Why?
`temporal schedule update` was building a new schedule from scratch and sending it as a full replacement, so omitting `--cron` (or other spec flags) wiped existing values.

## Checklist

1. Closes #927

2. How was this tested:
- `TestSchedule_Update` (existing): still passes.
- `TestSchedule_Update_PreservesCronWhenNotProvided`: create with `--cron "0 9 * * *"`, update without `--cron`, assert cron preserved.
- `TestSchedule_Update_PreservesJitterWhenNotProvided`: create with `--jitter 1m`, update without `--jitter`, assert jitter preserved.

3. Any docs updates needed?
I don't think so?

###Note:
I used AI to do most of the grunt work on this one, I reviewed all pushed code, but please call out if you notice anything I missed, or if anything doesn't follow project conventions, etc.

### DIFF
--- a/internal/temporalcli/commands.schedule.go
+++ b/internal/temporalcli/commands.schedule.go
@@ -221,31 +221,82 @@ func toIntervalSpec(str string) (client.ScheduleIntervalSpec, error) {
 }
 
 func (c *ScheduleConfigurationOptions) toScheduleSpec(spec *client.ScheduleSpec) error {
-	spec.CronExpressions = c.Cron
-	// Skip not supported
-	spec.Jitter = c.Jitter.Duration()
-	spec.TimeZoneName = c.TimeZone
-	spec.StartAt = c.StartTime.Time()
-	spec.EndAt = c.EndTime.Time()
+	return c.toScheduleSpecForUpdate(spec, nil)
+}
 
-	var err error
-	for _, calPbStr := range c.Calendar {
-		var calPb schedpb.CalendarSpec
-		if err = protojson.Unmarshal([]byte(calPbStr), &calPb); err != nil {
-			return fmt.Errorf("failed to parse json calendar spec: %w", err)
+// toScheduleSpecForUpdate applies CLI options to spec, preserving existing spec
+// fields when the user did not pass the corresponding flags (fully incremental update).
+// If existing is nil, behaves like toScheduleSpec (no preservation).
+func (c *ScheduleConfigurationOptions) toScheduleSpecForUpdate(spec *client.ScheduleSpec, existing *client.ScheduleSpec) error {
+	// Jitter, TimeZoneName, StartAt, EndAt: preserve from existing when flag not set
+	if existing != nil && c.FlagSet != nil {
+		if !c.FlagSet.Changed("jitter") {
+			spec.Jitter = existing.Jitter
+		} else {
+			spec.Jitter = c.Jitter.Duration()
 		}
-		cron, err := toCronString(&calPb)
-		if err != nil {
-			return err
+		if !c.FlagSet.Changed("time-zone") {
+			spec.TimeZoneName = existing.TimeZoneName
+		} else {
+			spec.TimeZoneName = c.TimeZone
 		}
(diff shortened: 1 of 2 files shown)


internal/temporalcli/commands.schedule.go
```diff
@@ -221,31 +221,82 @@ func toIntervalSpec(str string) (client.ScheduleIntervalSpec, error) {
...
 
-	var err error
-	for _, calPbStr := range c.Calendar {
-		var calPb schedpb.CalendarSpec
-		if err = protojson.Unmarshal([]byte(calPbStr), &calPb); err != nil {
-			return fmt.Errorf("failed to parse json calendar spec: %w", err)
+// toScheduleSpecForUpdate applies CLI options to spec, preserving existing spec
+// fields when the user did not pass the corresponding flags (fully incremental update).
+// If existing is nil, behaves like toScheduleSpec (no preservation).
+func (c *ScheduleConfigurationOptions) toScheduleSpecForUpdate(spec *client.ScheduleSpec, existing *client.ScheduleSpec) error {
```
cretz, line 230: I will let CLI reviewers review this, but I wanted to clarify...

The way this CLI works is not a bug, it is very much by intention. It's a backwards incompatible alteration to change this CLI command from being full update to be partial update. For example, users have been able to remove jitter by simply not providing the `--jitter` flag, but now they can't. Not only is that an incompatible alteration, there are likely a lot of users using it in this declarative form (i.e. they make update calls that are effectively no-ops because they have the same CLI args as what's already on server).

There is an inherent desire for declarative schedules so a user can partial update what they know to be the spec in some separate file. In the meantime, to remain compatible, we need to split off the concept of "partial update" and "full update" (and therefore if the former is needed as a _new feature_, it should be opted into either via flag or separate command).

billrich2001: Makes sense, this might make more sense as a separate patch command or something where it is intended to keep everything except for updated fields.

Will wait for feedback from CLI team

chaptersix: I agree with Chad on not making a backwards incompatible change. I'm not sure off the top of my head how we'd want to address this.

chaptersix: I don't think the server api supports updating the cron only right now:
https://github.com/temporalio/api/blob/e966760430f98290e83459571f40c91126ece7d8/temporal/api/schedule/v1/message.proto#L300

dnr: You can make whatever "patch" update you want with UpdateSchedule and the conflict token guard. This is built into the SDK as well

dnr: (but keep in mind that cron strings are canonicalized to structured calendar specs)

PR #930, opened 2026-02-03
Remove Experimental from Worker Deployment APIs, label previous versioning APIs Deprecated

carlydf:
## What was changed
Remove Experimental from Worker Deployment APIs, label previous versioning APIs Deprecated

## Why?
Worker Deployment Versioning is on its way to GA, no longer experimental.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/temporalcli/commands.yaml
+++ b/internal/temporalcli/commands.yaml
@@ -912,6 +912,10 @@ commands:
       ```
     option-sets:
       - deployment-version
+    options:
+      - name: report-task-queue-stats
+        type: bool
+        description: Report stats for task queues that are present in this version.
 
   - name: temporal worker deployment delete-version
     summary: Delete a Worker Deployment Version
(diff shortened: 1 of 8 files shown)


internal/temporalcli/commands.yaml
```diff
@@ -3644,13 +3554,6 @@ commands:
   - name: temporal workflow update-options
     summary: Change Workflow Execution Options
     description: |
-      ```
```
carlydf, line 3651: Given that we might still add more workflow execution properties to this in the future (like priority key), should we keep the experimental label here?

chaptersix: It's my understanding that we can make breaking changes while it's experimental. If we still want that flexibility then I think we should keep it.

internal/temporalcli/commands.yaml
```diff
@@ -3246,7 +3156,7 @@ commands:
         description: Bypass the validation that one unconditional rule remains.
 
   - name: temporal task-queue versioning replace-redirect-rule
-    summary: Change the target for a Build ID's redirect (Experimental)
```
chaptersix, line 3253: Should we just remove these?

PR #933, opened 2026-02-04
add CI validation for server dependency version

chaptersix:
## Summary
- Adds a new CI job that validates the `go.temporal.io/server` dependency
- Ensures the dependency is a tagged version (not a pseudo-version with commit hash)
- Ensures the version doesn't exceed the next server release (major.minor+1 from latest GitHub release)

## Test plan
- [x] CI job runs successfully on this PR
- [x] Verify validation passes for current `v1.30.0-148.4` (latest release is `v1.29.2`)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,8 +16,8 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/cli/cliext v0.0.0
-	github.com/temporalio/ui-server/v2 v2.42.1
-	go.temporal.io/api v1.60.0
+	github.com/temporalio/ui-server/v2 v2.45.0
+	go.temporal.io/api v1.60.1
 	go.temporal.io/sdk v1.38.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
 	golang.org/x/mod v0.31.0
--- a/go.sum
+++ b/go.sum
@@ -323,8 +323,8 @@ github.com/temporalio/sqlparser v0.0.0-20231115171017-f4060bcfa6cb/go.mod h1:143
 github.com/temporalio/tchannel-go v1.22.1-0.20220818200552-1be8d8cffa5b/go.mod h1:c+V9Z/ZgkzAdyGvHrvC5AsXgN+M9Qwey04cBdKYzV7U=
 github.com/temporalio/tchannel-go v1.22.1-0.20240528171429-1db37fdea938 h1:sEJGhmDo+0FaPWM6f0v8Tjia0H5pR6/Baj6+kS78B+M=
 github.com/temporalio/tchannel-go v1.22.1-0.20240528171429-1db37fdea938/go.mod h1:ezRQRwu9KQXy8Wuuv1aaFFxoCNz5CeNbVOOkh3xctbY=
-github.com/temporalio/ui-server/v2 v2.42.1 h1:ajeOxqCnUiCRQQhQYLxaT7wUgF/slqZJtdW4pLjVqCs=
-github.com/temporalio/ui-server/v2 v2.42.1/go.mod h1:lKTnn50t8yQvcrarxAOjX33YcfkomkiNB5BH06wQwEE=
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,6 +10,85 @@ permissions:
...
+    name: Validate Server Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Validate server dependency version
+        uses: actions/github-script@v7
+        with:
+          script: |
```
cretz: Will defer to CLI team, but IMO any build tooling/validation like this should be written like any other software that anyone can run at any time (in these cases in other repositories, we've often made Go commands somewhere under `internal` only for use as build tooling). But up to y'all.

bergundy: I don't think it needs to be a Go command because I don't see much motivation for running this outside of CI.
I would accept any modern programming language, just not shell. JS is fine in this case but it would be great if we could put the script in its own file to get proper syntax highlighting.

.github/workflows/ci.yaml
```diff
@@ -10,6 +10,85 @@ permissions:
...
+            const fs = require('fs');
+
+            const SERVER_MODULE = 'go.temporal.io/server';
+
+            // Parse major.minor from a version string like "v1.30.0-148.4" or "v1.29.2"
+            function parseMajorMinor(version) {
+              const match = version.match(/^v?(\d+)\.(\d+)/);
```
bergundy: I don't think the v is optional.

```suggestion
              const match = version.match(/^v(\d+)\.(\d+)/);
```

.github/workflows/ci.yaml
```diff
@@ -10,6 +10,85 @@ permissions:
...
+              if (!match) return null;
+              return { major: parseInt(match[1], 10), minor: parseInt(match[2], 10) };
+            }
+
+            // 1. Read and parse go.mod
+            const goModContent = fs.readFileSync('go.mod', 'utf-8');
+            const serverVersionMatch = goModContent.match(new RegExp(`${SERVER_MODULE.replace(/\//g, '\\/')}\\s+(v[^\\s]+)`));
```
bergundy: You should be able to use [RegExp.escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape) here.

.github/workflows/ci.yaml
```diff
@@ -10,6 +10,85 @@ permissions:
   actions: write
 
 jobs:
+  validate-server-version:
```
bergundy, line 13: This should only run on merge to main.

.github/workflows/ci.yaml
```diff
@@ -10,6 +10,85 @@ permissions:
...
+            // 2. Validate it's a tagged version (not a pseudo-version)
+            const pseudoPattern = /-\d{14}-[a-f0-9]{12}$/;
+            if (pseudoPattern.test(serverVersion)) {
+              core.setFailed(`Server dependency must be a tagged version, not a pseudo-version: ${serverVersion}`);
+              return;
+            }
+            core.info('✓ Version is a valid tagged version');
+
+            // 3. Fetch latest release from GitHub
+            const { data: release } = await github.rest.repos.getLatestRelease({
```
bergundy: Make sure you use an API key here to avoid getting throttled by GH.

.github/workflows/ci.yaml
cretz: > I don't see much motivation for running this outside of CI

Even without much motivation, why not? It doesn't add anything, is not meaningfully more difficult, and is a clearer way to write software. I think you'll find as this script and the pattern in general grows, the unwieldiness of this approach will become apparent, might as well do it right while it doesn't cost anything.

> JS is fine

This cannot be run in a traditional JS runtime as is, it's "github-script" which has a bunch of special runtime affordances. I too think traditional JS approach would be good as opposed to GH-runtime-specific.

> to get proper syntax highlighting.

Arguably if you can give up type checking, you can give up syntax highlighting.

cretz (approved):

PR #944, opened 2026-02-11
Bump UI server to 2.45.3 and API to 1.60.2

laurakwhit:
## What was changed

Bumps UI server to [v2.45.3](https://github.com/temporalio/ui-server/releases/tag/v2.45.3) and API to [v1.60.2](https://github.com/temporalio/api/releases/tag/v1.60.2).

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/cli/cliext v0.0.0
-	github.com/temporalio/ui-server/v2 v2.45.0
+	github.com/temporalio/ui-server/v2 v2.45.3
 	go.temporal.io/api v1.60.2
 	go.temporal.io/sdk v1.38.0
 	go.temporal.io/sdk/contrib/envconfig v0.1.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -16,12 +16,12 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.10.0
 	github.com/temporalio/cli/cliext v0.0.0
-	github.com/temporalio/ui-server/v2 v2.45.0
-	go.temporal.io/api v1.60.1
+	github.com/temporalio/ui-server/v2 v2.45.3
```
laurakwhit, line 19: > Can you confirm that standalone activities UI changes aren't included in this UI server version? That feature didn't make it into server 1.30.x.

@bergundy standalone activities UI (see https://github.com/temporalio/ui/pull/3124) is technically in this UI server version, but we are currently 404ing the route so none of it is exposed in the UI atm.

bergundy: But is the side bar item showing by default?

bergundy: Doesn't look like it:

image

laurakwhit: Nope, it's removed from side nav bar for now as well.

PR #945, opened 2026-02-11
docs: add note for schedule update command

lennessyy:
## What was changed

- Added a docs note about the schedule update command, specifying that they need to provide the full options and the update will not retain old values. 
- Context: https://temporaltechnologies.slack.com/archives/C01H1G7J98F/p1770221845107689?thread_ts=1770221845.107689&cid=C01H1G7J98F

## Why?

The current documentation makes it sound like users can get away with only supplying the parts they want to update. We should be specific that they need to provide the full options and any unprovided option will be reset to default 

## Checklist

1. Closes [EDU-5878](https://temporalio.atlassian.net/browse/EDU-5878)

2. How was this tested:

N/A. Docs change only. 

3. Any docs updates needed?

N/A

[EDU-5878]: https://temporalio.atlassian.net/browse/EDU-5878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/temporalio/cli
 
-go 1.25.5
+go 1.26.0
 
 require (
 	github.com/BurntSushi/toml v1.4.0


internal/temporalcli/commands.yaml
```diff
@@ -2240,6 +2240,12 @@ commands:
           --schedule-id "YourScheduleId" \
           --workflow-type "NewWorkflowType"
       ```
+
+      This command performs a full replacement of the Schedule
+      configuration. Any options not provided will be reset to their default
+      values. You must re-specify all options, not just the ones you want to
+      change. To view the current configuration of a Schedule, use
+      `temporal schedule describe` before updating.
```
chaptersix, line 2249:
```suggestion
  This command performs a full replacement of the Schedule
  configuration. Any options not provided will be reset to their default
  values. You must re-specify all options, not just the ones you want to
  change. To view the current configuration of a Schedule, use
  `temporal schedule describe` before updating.

  Note: Schedule Memo and Search Attributes cannot be updated with this
  command. 
```

chaptersix: I don't think the search attributes should be listed in the options if they are not supported by the update command.

internal/temporalcli/commands.yaml
```diff
@@ -2240,6 +2241,16 @@ commands:
...
+
+      Schedule memo and search attributes cannot be updated with this
```
dnr, line 2251: This is wrong: search attributes can be updated with update. Currently memo can't, but there's an open feature request for that.

chaptersix: sorry, I didn't see this comment before merging. double checking

chaptersix: Screenshot 2026-02-22 at 8 29 53 PM

chaptersix: I didn't see this comment before merging. This line could probably be a bit more clear.

From my understanding, search attributes for the schedule itself can not be modified with this command. BUT the server does support this. 
This command does allow you to update the search attributes for the workflows.

chaptersix: FWIW, the cli should support updating schedule search attributes and there's been requests to have 'patch' like behavior.

dnr: You're right, support for updating schedule SAs was was never added to CLI, only to SDKs. It should be easy to add though since it's supported in Go SDK

PR #980, opened 2026-04-14
ci: add -subdir flag to gen-docs to support cloud CLI extension docs

lennessyy:
## Summary

The Temporal Cloud CLI extension needs auto-generated command reference documentation, just like the core CLI commands. The extension has ~130 commands across 10 top-level groups (namespace, user, apikey, account, etc.), and the documentation repo's workflow already uses `gen-docs` with multiple `-input` flags to generate docs from YAML command definitions.

The problem is that `gen-docs` currently produces one file per top-level command. Since all cloud commands share the `cloud` root, they'd all land in a single `cloud.mdx` file (~2800 lines), which is too large to be useful.

This PR adds a `-split` flag that tells `gen-docs` to output a command's subcommands as separate files in a subdirectory. The documentation repo's workflow will pass `-split cloud` to produce individual pages:

```
cloud/namespace.mdx  (1148 lines)
cloud/account.mdx    (318 lines)
cloud/user-group.mdx (417 lines)
cloud/user.mdx       (262 lines)
cloud/apikey.mdx     (253 lines)
cloud/nexus.mdx      (212 lines)
cloud/connectivity.mdx (144 lines)
cloud/login.mdx
cloud/logout.mdx
cloud/whoami.mdx
```

No changes are needed in the cloud-cli repo. The split is controlled entirely by the documentation workflow:

```bash
go run ./cmd/gen-docs \
  -input internal/temporalcli/commands.yaml \
  -input path/to/cloud-cli/commands.yml \
  -split cloud \
  -output dist/docs
```

Also fixes the `encodeJSONExample` regex to catch standalone JSON objects in single quotes (e.g., `'{"key":"value"}'`), which caused MDX rendering errors.

## Test plan

- [x] CLI-only generation produces identical flat output (no regressions)
- [x] `-split cloud` produces `cloud/` subdirectory with correct file splits
- [x] Generated content (frontmatter, headings, flag tables, global flags) renders correctly
- [x] JSON-in-tables MDX rendering fix works
- [ ] CI checks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


chaptersix:
Suggestion: rename `-split` to `-subdir`.

`-split` describes the transformation (splitting one file into many), but `-subdir` describes the output structure — "put this command's subcommands into a subdirectory." Someone reading `-subdir cloud` in a Makefile or CI workflow immediately understands what happens to the output, without needing to know the internal mechanics.

This also reads more naturally as the flag gets used alongside `-output`:

```bash
go run ./cmd/gen-docs \
  -input internal/temporalcli/commands.yaml \
  -input cliext/option-sets.yaml \
  -input ../cloud-cli/temporalcloudcli/commands.yml \
  -output dist/docs \
  -subdir cloud
```

Also suggest updating the flag description from:

> "Command name whose subcommands get separate files in a subdirectory (can be specified multiple times)"

to:

> "Place subcommands of this command into a subdirectory instead of a single file (can be specified multiple times)"

PR #987, opened 2026-04-20
CLI Support for UpdateTaskQueueConfig's fairness weight overrides

majiru:
## What was changed
Add support for setting fairness key weight overrides.

## Why?
We should expose this feature to users.

## Checklist

2. How was this tested:
Newly added tests and tested against a local dev server

3. Any docs updates needed?
Yes; https://docs.temporal.io/cli/task-queue


internal/temporalcli/commands.yaml
```diff
@@ -4582,7 +4603,7 @@ option-sets:
       - name: fairness-weight
         type: float
         description: |
-          Weight [0.001-1000] for this fairness key.
+          Weight (minimum 0.001) for this fairness key.
```
dnr: I'd leave this actually.. weights > 1000 won't really have an effect, even if they are accepted by the api

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -79,43 +168,73 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
-		if queueRpsLimitParsed, err = parseRPS("queue-rps-limit"); err != nil {
+		queueRpsLimitParsed, queueRateLimitIsZero, err = parseRPS("queue-rps-limit")
+		if err != nil {
 			return err
 		}
+
+		// Validate reason is provided
+		if c.QueueRpsLimitReason == "" {
+			return fmt.Errorf("--queue-rps-limit-reason is required when setting or unsetting queue rate limit")
+		}
```
dnr: did it do this before? this feels a little heavy-handed

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -152,10 +271,86 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+	// Validate mutual exclusivity of unset-all with other weight operations
+	if c.FairnessKeyWeightUnsetAll {
+		if len(c.FairnessKeyWeightSet) > 0 || len(c.FairnessKeyWeightUnset) > 0 {
+			return fmt.Errorf("--fairness-key-weight-unset-all cannot be used with --fairness-key-weight-set or --fairness-key-weight-unset")
+		}
+	}
+
+	// Parse and validate set operations
+	var setWeights map[string]float32
+	if len(c.FairnessKeyWeightSet) > 0 {
```
dnr: just have parseFairnessKeyWeights return nil on empty input, and you don't need a conditional and forward declaration here, it'll just make the code read better

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -152,10 +271,86 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+		var err error
+		setWeights, err = parseFairnessKeyWeights(c.FairnessKeyWeightSet)
+		if err != nil {
+			return err
+		}
+		request.SetFairnessWeightOverrides = setWeights
+	}
+
+	// Validate unset operations
+	if len(c.FairnessKeyWeightUnset) > 0 {
```
dnr: same here

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -152,10 +271,86 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+				return fmt.Errorf("fairness keys appear in both set and unset operations: %v", conflicts)
+			}
+		}
+
+		request.UnsetFairnessWeightOverrides = c.FairnessKeyWeightUnset
+	}
+
+	// Handle unset all
+	if c.FairnessKeyWeightUnsetAll {
+		// Need to fetch current config to get all keys to unset
```
dnr, line 324: I thought we had an unset-all option :(  ah well

internal/temporalcli/commands.yaml
```diff
@@ -3278,6 +3284,21 @@ commands:
...
+      - name: fairness-key-weight-unset
+        type: string[]
+        description: |
+          Unset specific fairness key weight overrides.
+          Can be specified multiple times. Example: --fairness-key-weight-unset HighPriority --fairness-key-weight-unset LowPriority.
+      - name: fairness-key-weight-unset-all
+        type: bool
+        description: |
+          Unset all fairness key weight overrides.
+          Cannot be used with --fairness-key-weight-set or --fairness-key-weight-unset.
```
chaptersix: how about:
```
 --fairness-key-weight HighPriority=2.0    # set
  --fairness-key-weight HighPriority=default  # unset one
  --fairness-key-weight-clear-all            # unset all
  ```
  
  Similar to:
  ```
  temporal task-queue config set \
    --task-queue my-queue \
    --queue-rps-limit default
    ```

File: internal/temporalcli/commands.taskqueue_config.go [hunk 7]
```diff
@@ -56,12 +56,121 @@ func (c *TemporalTaskQueueConfigGetCommand) run(cctx *CommandContext, args []str
...
+	// maxFairnessKeyLength matches server-side limit in temporal/common/priorities/priority_util.go
+	maxFairnessKeyLength = 64
+
+	// minFairnessWeight matches server-side limit in temporal/service/matching/fairness_util.go
+	// Server clamps weights to this minimum when applying them
+	minFairnessWeight = 0.001
+
+	// maxFairnessWeight is the maximum weight value allowed
+	// Server clamps weights above this value down to this maximum
+	maxFairnessWeight = 1000.0
```
dnr: we should not enforce these at all. clients do not need to know the effective range that the server is currently configured with, only that the the weight must be positive.

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -56,12 +56,121 @@ func (c *TemporalTaskQueueConfigGetCommand) run(cctx *CommandContext, args []str
...
+	weights := make(map[string]float32)
+	seen := make(map[string]bool)
+
+	for _, input := range inputs {
+		parts := strings.SplitN(input, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid format: %q (expected key=weight)", input)
+		}
+
+		key := strings.TrimSpace(parts[0])
```
dnr: I don't really like "helpful" things like this.. there's nothing that says a valid fairness key can't start or end with whitespace, and the server doesn't enforce anything like that, so we shouldn't disallow that here.

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -79,43 +188,59 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+		if err != nil {
 			return err
 		}
-	} else if c.Command.Flags().Changed("fairness-key-rps-limit-default-reason") {
-		return fmt.Errorf("fairness-key-rps-limit-default-reason can only be set if fairness-key-rps-limit-default is updated")
+
+		// Warn about zero rate limit
+		if fairnessRateLimitIsZero {
+			cctx.Printer.Println("WARNING: Setting fairness key rate limit default to 0 will STOP ALL TRAFFIC for fairness keys.")
+			cctx.Printer.Println("         This will prevent tasks with fairness keys from being dispatched until the limit is changed.")
```
dnr:
```suggestion
			cctx.Printer.Println("WARNING: Setting fairness key rate limit default to 0 will STOP ALL TRAFFIC on this task queue.")
			cctx.Printer.Println("         This will prevent any tasks from being dispatched until the limit is changed.")
```
all tasks have fairness keys, the fairness key may be the empty string

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -56,12 +56,121 @@ func (c *TemporalTaskQueueConfigGetCommand) run(cctx *CommandContext, args []str
...
+		}
+
+		trimmed[i] = trimmedKey
+	}
+	return trimmed, nil
+}
+
+// findConflictingKeys returns keys that appear in both set and unset lists
+func findConflictingKeys(setWeights map[string]float32, unsetKeys []string) []string {
+	conflicts := []string{}
```
dnr:
```suggestion
	var conflicts []string
```
avoids allocation if nothing is appended

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -152,10 +277,81 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+		return fmt.Errorf("invalid fairness key in unset list: %w", err)
+	}
+
+	// Check for conflicts between set and unset
+	if setWeights != nil && trimmedUnsetKeys != nil {
+		conflicts := findConflictingKeys(setWeights, trimmedUnsetKeys)
+		if len(conflicts) > 0 {
+			return fmt.Errorf("fairness keys appear in both set and unset operations: %v", conflicts)
+		}
+	}
```
dnr:
```suggestion
	conflicts := findConflictingKeys(setWeights, trimmedUnsetKeys)
	if len(conflicts) > 0 {
		return fmt.Errorf("fairness keys appear in both set and unset operations: %v", conflicts)
	}
```
findConflictingKeys returns an empty slice if given nil args, right?

internal/temporalcli/commands.taskqueue_config.go
```diff
@@ -152,10 +277,81 @@ func (c *TemporalTaskQueueConfigSetCommand) run(cctx *CommandContext, args []str
...
+		})
+		if err != nil {
+			return fmt.Errorf("error fetching current config for unset-all: %w", err)
+		}
+		if descResp.Config != nil && descResp.Config.FairnessWeightOverrides != nil && len(descResp.Config.FairnessWeightOverrides) > 0 {
+			keys := make([]string, 0, len(descResp.Config.FairnessWeightOverrides))
+			for key := range descResp.Config.FairnessWeightOverrides {
+				keys = append(keys, key)
+			}
+			request.UnsetFairnessWeightOverrides = keys
```
dnr:
```suggestion
import expmaps "golang.org/x/exp/maps"

			request.UnsetFairnessWeightOverrides = expmaps.Keys(descResp.Config.FairnessWeightOverrides)
```

internal/temporalcli/commands.taskqueue_config.go
majiru: Sure, I've removed these enforcements. Do you feel the same way about the fairness key limit?

dnr (approved):
works for me!

PR #990, opened 2026-04-21
`temporal worker deployment create|create-version`

jaypipes:
Adds implementation of the `temporal worker deployment create` and `temporal worker deployment create-version` CLI commands using only direct gRPC API calls, not the sdk-go client code.

Adds some basic unit tests for both commands though due to the server-side validation of proper AWS IAM credentials, the happy-path `temporal worker deployment create-version` call with AWS Lambda compute config is skipped until such time as we can figure out adding real AWS test fixtures.


go.mod
```diff
@@ -216,3 +214,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+replace github.com/temporalio/cli/cliext => ./cliext
```
chaptersix: nit: Ideally there's no change in this file.

chaptersix (approved):

rossnelson:
We need to bump the ui-server dependency to [2.48.4](https://github.com/temporalio/ui-server/releases/tag/v2.48.4) 

This way the dev server will have feature parody with the CLI

jaypipes:
> We need to bump the ui-server dependency to [2.48.4](https://github.com/temporalio/ui-server/releases/tag/v2.48.4)
> 
> This way the dev server will have feature parody with the CLI

done!

PR #1001, opened 2026-04-27
worker deployment update-version-compute-config

jaypipes:
Adds implementation of the `UpdateWorkerDeploymentVersionComputeConfig` gRPC API call with a `temporal worker deployment
update-version-compute-config` CLI command. There is a `--remove` CLI flag that allows the user to explicitly delete the compute config from a Worker Deployment Version.


internal/temporalcli/commands.yaml
```diff
@@ -979,6 +979,7 @@ commands:
         - worker deployment set-ramping-version
         - worker deployment delete-version
         - worker deployment update-version-metadata
+        - worker deployment update-version-compute-config
```
chaptersix, line 982:
```suggestion
        - worker deployment replace-version-compute-config
```

jaypipes: done! :)

jaypipes:
@chaptersix I don't believe the test failure has anything to do with this particular PR.

chaptersix:
A couple things worth reconsidering here:

**Command naming** -- The CLI uses `update-version-metadata` for modifying properties on an existing version, and the gRPC API is `UpdateWorkerDeploymentVersionComputeConfig`. Using "replace" instead of "update" breaks the pattern -- we want CLI commands to be predictable so users can remember and guess at what a command or flag might be. Something like `update-version-compute-config` would be more consistent.

**Implicit delete behavior** -- Currently if `--aws-lambda-function-arn` is omitted, the command silently removes the compute config. That's surprising given no other command treats the absence of a flag as a destructive action. Something like a `--remove` flag (similar to how `update-version-metadata` uses `--remove-entries`) would make the behavior predictable and consistent with the rest of the CLI.

jaypipes:
> A couple things worth reconsidering here:
> 
> **Command naming** -- The CLI uses `update-version-metadata` for modifying properties on an existing version, and the gRPC API is `UpdateWorkerDeploymentVersionComputeConfig`. Using "replace" instead of "update" breaks the pattern -- we want CLI commands to be predictable so users can remember and guess at what a command or flag might be. Something like `update-version-compute-config` would be more consistent.
> 
> **Implicit delete behavior** -- Currently if `--aws-lambda-function-arn` is omitted, the command silently removes the compute config. That's surprising given no other command treats the absence of a flag as a destructive action. Something like a `--remove` flag (similar to how `update-version-metadata` uses `--remove-entries`) would make the behavior predictable and consistent with the rest of the CLI.

@chaptersix all great points. I will rework the patch!

jaypipes:
@chaptersix sorry for the delay and thank you for your patience. I've updated the code per your review recommendations.

internal/temporalcli/commands.worker.deployment.go
```diff
@@ -969,6 +970,97 @@ func (c *TemporalWorkerDeploymentCreateVersionCommand) run(cctx *CommandContext,
...
+		Namespace: ns,
+		DeploymentVersion: &deployment.WorkerDeploymentVersion{
+			DeploymentName: deploymentName,
+			BuildId:        buildID,
+		},
+		Identity:  identity,
+		RequestId: requestID,
+	}
+
+	if c.Remove {
```
fretz12, line 1004: Though AwsLambdaFunctionArn, AwsLambdaAssumeRoleArn, AwsLambdaAssumeRoleExternalId are harmlessly ignored when --remove is passed in, there's precedence in this repo (i.e. see TemporalOperatorNexusEndpointCreateCommand) to guard against non-supported permutations.  Maybe a guard like this in the top of the method: 
```
	if c.Remove && (c.AwsLambdaFunctionArn != "" || c.AwsLambdaAssumeRoleArn != "" || c.AwsLambdaAssumeRoleExternalId != "") {
		return fmt.Errorf("--remove cannot be combined with --aws-lambda-function-arn, --aws-lambda-assume-role-arn, or --aws-lambda-assume-role-external-id")
	}
```

internal/temporalcli/commands.worker.deployment.go
```diff
@@ -969,6 +970,97 @@ func (c *TemporalWorkerDeploymentCreateVersionCommand) run(cctx *CommandContext,
 	return nil
 }
 
+// awsLambdaProviderDetailsPayload returns the encoded Payload representing AWS
+// Lambda compute provider details.
+func (c *TemporalWorkerDeploymentUpdateVersionComputeConfigCommand) awsLambdaProviderDetailsPayload() (*commonpb.Payload, error) {
```
fretz12: I think this looks functionally identical to `TemporalWorkerDeploymentCreateVersionCommand.awsLambdaProviderDetailsPayload`? Maybe refactor into a helper func

internal/temporalcli/commands.worker.deployment.go
```diff
@@ -969,6 +970,97 @@ func (c *TemporalWorkerDeploymentCreateVersionCommand) run(cctx *CommandContext,
...
+	providerDetails := map[string]any{
+		"arn": c.AwsLambdaFunctionArn,
+	}
+	if c.AwsLambdaAssumeRoleArn != "" {
+		providerDetails["role"] = c.AwsLambdaAssumeRoleArn
+	}
+	if c.AwsLambdaAssumeRoleExternalId != "" {
+		providerDetails["role_external_id"] = c.AwsLambdaAssumeRoleExternalId
+	}
+	err := validateAWSLambdaProviderDetails(providerDetails)
```
fretz12: Seems this will reject if all three AWS Lambda flags are not supplied together. Maybe worth it to specify in help all 3 must be supplied together

jaypipes: The help strings for those three CLI options states that they are required when `--aws-lambda-function-arn` is specific

internal/temporalcli/commands.worker.deployment.go
jaypipes: done.

internal/temporalcli/commands.worker.deployment.go
jaypipes: done.

jaypipes:
@fretz12 thanks very much for your code review and sorry for the long delay in pushing another revision. I've addressed your comments. thanks!

fretz12 (approved):

PR #1005, opened 2026-04-29
Fix flakey test by disabling EC2 metadata lookup in test case

kevinawoo:
## Summary
- Disable EC2 IMDS lookups inside `TestCreateWorkerDeploymentVersion_Errors`.
- Keep the fake AWS Lambda validation failure deterministic and local.

## Root Cause
The test intentionally passes fake AWS Lambda and assume-role inputs. Server-side worker-deployment validation uses `go.temporal.io/auto-scaled-workers`, which imports the AWS SDK and validates the `aws-lambda` compute provider. That path builds default AWS config and calls STS/Lambda validation.

Before STS can assume the fake role, the AWS SDK needs base credentials. In a local or CI environment with no AWS env creds/profile/container creds, the AWS SDK falls back to EC2 Instance Metadata Service. Outside EC2, that means waiting on IMDS/network retries even though the test only wants an expected validation failure.

## Scope of the Fix
This is broader AWS SDK behavior: `AWS_EC2_METADATA_DISABLED=true` is useful for any test/process that invokes AWS SDK default credential resolution without real credentials and should not hit EC2 IMDS.

In this CLI test suite, the enabled surface is narrow:
- `TestCreateWorkerDeploymentVersion_Errors` is the only enabled test that passes a complete fake AWS Lambda compute config to the server.
- The later missing-field subcases in that same test fail in CLI-side validation before server-side AWS validation.
- `TestCreateWorkerDeploymentVersion_LambdaComputeConfig` would also hit the AWS path, but it is skipped because it needs real AWS fixtures.

So the fix is intentionally test-local rather than package-wide, to avoid changing behavior for future tests that might intentionally exercise AWS credential resolution.

## Validation
- `go test -mod=readonly ./internal/temporalcli -run 'TestSharedServerSuite/TestCreateWorkerDeploymentVersion_Errors' -count=1 -v`


chaptersix:
I think this PR is closer to the fix: https://github.com/temporalio/temporal/pull/10127

@jaypipes will take a closer look soon.

I appreciate the contributions!

jaypipes (approved):
This doesn't hurt anything and does indeed improve the total test time by reducing the waiting on network retries. ++

chaptersix (approved):

PR #1006, opened 2026-04-29
Fix help with value flags

### LINKED ISSUE #1003: --help flag prints 'pflag: help requested' when used with other flags
When `--help` is combined with another flag that takes a value, the CLI prints `pflag: help requested` instead of showing the help text.

**Reproduce:**

```
$ temporal workflow list --address 123 --help
Error: pflag: help requested
```

**Expected:**

```
$ temporal workflow list --help
List Workflow Executions. [...]
```

This happens with any flag that takes a value, not just `--address`:

```
$ temporal workflow list --namespace foo --help
Error: pflag: help requested
```

Order doesn't matter -- `--help --address 123` also fails.

Reported on v1.6.2.

kevinawoo:
## Summary
- avoid routing built-in help invocations through extension probing when known value flags are present
- initialize Cobra's default help flag before extension preflight parses command flags
- add regression coverage for `--help` combined with `--address` and `--namespace`

## Root Cause
Extension preflight checked raw remaining args with a string-prefix heuristic. Values for known flags, such as `123` in `--address 123`, looked like positional extension args, so the preflight parsed `--help` before Cobra had registered the default help flag and surfaced `pflag: help requested` as a real error.

Fixes #1003.

## Validation
- `go run ./cmd/temporal workflow list --address 123 --help`
- `go test ./internal/temporalcli -run 'TestHelp|TestExtension' -count=1`
- `go test ./internal/temporalcli -count=1`


internal/temporalcli/commands.extension.go
```diff
@@ -31,18 +31,15 @@ func tryExecuteExtension(cctx *CommandContext, tcmd *TemporalCommand) (error, bo
...
-
 	// Group args into these lists:
 	// - cliParseArgs: args to validate (subset of cliPassArgs)
 	// - cliPassArgs: known CLI args to pass to extension
 	// - extArgs: args to pass to extension and use for extension lookup
+	foundCmd.InitDefaultHelpFlag()
 	cliParseArgs, cliPassArgs, extArgs := groupArgs(foundCmd, remainingArgs)
+	if findErr == nil && !slices.ContainsFunc(extArgs, isPosArg) {
+		return nil, false
+	}
```
chaptersix: nit: fix comment alignment

add back the comments removed on 34 (in the correct location) 

add a comment on why InitDefaultHelpFlag() is being called

kevinawoo: done!

chaptersix (approved):

davidporter-id-au (approved):

PR #1008, opened 2026-04-29
chore: remove EXPERIMENTAL labels from config subcommands

chaptersix:
## Summary
- Remove `(EXPERIMENTAL)` from config subcommand summaries: `config`, `config set`, `config get`, `config delete`, `config delete-profile`, `config list`
- Follows up on #972 which removed experimental labels from the config global flags but missed the subcommands

## Test plan
- [x] `make gen` produces clean output
- [x] `make build` succeeds
- [ ] `temporal config --help` no longer shows EXPERIMENTAL


ks-temporal (approved):
Verified "./temporal config --help no longer shows EXPERIMENTAL" in local env for those subcommands.

chaptersix:
holding off on this. we might want to make some changes that would be considered breaking.

PR #1017, opened 2026-04-30
Unwrap System Nexus Operations in event history

spkane31:
## What was changed
- `workflow show` (table and detailed modes, not `-o json`) now displays the actual operation name instead of the generic `NexusOperationScheduled`/`NexusOperationCompleted` event type when the Nexus endpoint is `__temporal_system`. 
- `workflow describe` no longer includes pending Nexus operations on the `__temporal_system` endpoint in the "Pending Nexus Operations" list or count.   

## Why?
System Nexus operations (endpoint `__temporal_system`) are implementation details of high-level SDK operations like `SignalWithStartWorkflowExecution`. Surfacing them as `NexusOperationScheduled`/`NexusOperationCompleted` events is confusing. By unwrapping the operation name, the history and describe output reflect what the workflow actually did at a semantic level. Similarly, pending system operations in describe are noise the user has no actionable interest in.

## Checklist

1. Closes NA

2. How was this tested:
- Unit tests added in `commands.workflow_show_test.go` covering all six NexusOperation event types, the non-system-endpoint no-op path, and the missing-scheduled-event (reverse traversal / orphaned completion) path.
- Functional tests added to `commands.workflow_view_test.go` to show system operations are unwrapped but user defined are not
- Manual validation: ran `temporal workflow show -w <id>` against a live workflow that executes a `SignalWithStartWorkflowExecution` system nexus operation and confirmed events 5 and 6 now show as `SignalWithStartWorkflowExecutionScheduled` and `SignalWithStartWorkflowExecutionCompleted`.     
4. Any docs updates needed? No


chaptersix:
Could add new or amend existing functional tests. It'd be great to have some assertion both human readable output and json.

internal/temporalcli/commands.workflow_exec.go
```diff
@@ -704,6 +704,68 @@ func coloredEventType(e enums.EventType) string {
 	return fn(e.String())
 }
 
+const temporalSystemNexusEndpoint = "__temporal_system"
+
+// systemNexusOpDisplayName returns a display name for Nexus operation events on the
+// "__temporal_system" endpoint, replacing the generic NexusOperation* name with the
+// actual operation name + event-type suffix. Returns "" for all other events.
+// Populates s.systemNexusOps when it encounters a qualifying Scheduled event.
+func (s *structuredHistoryIter) systemNexusOpDisplayName(event *history.HistoryEvent) string {
```
chaptersix, line 716: we also have a `--reverse` flag. Could you add a test for this case?

chaptersix (approved):

PR #1023, opened 2026-05-12
Generate CLI environment variables reference

### LINKED ISSUE #697: Create an index of environment variables for the docs site
The code generation we do for the docs site should also generate an index of environment variables similar to: https://docs.temporal.io/cli#environment-variables

NgThachThanh:
## Summary
- generate an `environment-variables.mdx` page from options with `implied-env`
- include variable name, related flag, description, and default value
- add a small gen-docs test check for the new page

Fixes #697

## Test plan
- `go test ./cmd/gen-docs ./internal/commandsgen`
- `go test ./...`
- `go run ./cmd/gen-docs -input internal/temporalcli/commands.yaml -input cliext/option-sets.yaml -output /tmp/temporal-cli-docs`


chaptersix:
Thank you for your PR! we took a closer look and decided we need a more unified view than what we can do here so this PR is no longer applicable.

Issue #581, opened 2024-06-04
[Bug] Dev Server concurrent execution test is incorrect and flaky


mjameswh:
### Describe the bug

We recently added an integration test that starts a total of 80 dev server, with a maximum of 8 instances running concurrently. This is used to assert, among other things, the reliability of the "get ephemeral port number" logic.

However, that test is incorrect, because it calls `FailNow` and similar from goroutines, which is not allowed. The test is also [flaky](https://github.com/temporalio/cli/actions/runs/9308352928/job/25621584716?pr=580). Incorrect usage of test APIs is like not the cause of the flakiness, but may be masking the actual errors, making the issue harder to diagnose.

Given that the CLI's test harness internally use `FailNow`, fixing this will likely require developing a substitute for the existing test harness.

PR #1026, opened 2026-05-16
Run concurrent dev server starts as parallel subtests

highlyavailable:
`TestServer_StartDev_ConcurrentStarts` spawned its worker goroutines by hand and called the testify `require` helpers from inside them, both directly and through the test harness. `require` calls `t.FailNow`, which Go only allows from the goroutine running the test, so any failure it hit was reported unreliably.

Each instance now runs as a parallel subtest with its own `*testing.T`, so the assertions run on a goroutine where `FailNow` is valid. A buffered channel caps concurrency at 8, same as before.

While verifying with `-race`, the corrected test surfaced a real data race in `initCommand`: concurrent `Execute()` calls read and write the global `color.NoColor` from `fatih/color`. That looks like the actual flakiness this test was meant to catch. It's a separate production bug so I've left it out of this PR and filed it as #1027.

Refs #581


chaptersix:
My understanding is that `t.Parallel()` concurrency is bounded by the `-parallel` flag (default: `GOMAXPROCS`) per the [go test docs](https://pkg.go.dev/cmd/go#hdr-Testing_flags), meaning the semaphore may never actually block if that limit is below 8. Before this PR, concurrency was managed entirely by the goroutine pool with no dependency on the test runner's parallelism setting.

chaptersix:
>While verifying with -race, the corrected test surfaced a real data race in initCommand: concurrent Execute() calls read and write the global color.NoColor from fatih/color. That looks like the actual flakiness this test was meant to catch. It's a separate production bug so I've left it out of this PR and filed it as https://github.com/temporalio/cli/issues/1027.

awesome! thank you

chaptersix:
PR appears to be stale. feel free to re-open if that's not the case

PR #1029, opened 2026-05-18
:boom: Workflow delete now prompts for confirmation

jiechenz:
## What was changed

1. Added a warning to `temporal workflow delete` explaining that deleting Workflow Executions in a
global Namespace removes them from all replicas, and that requests sent to a passive cluster are forwarded to the active cluster by default unless `--grpc-meta xdc-redirection=false` is specified.

2. Delete workflow now prompts for confirmation

## Why?
The CLI should make that blast radius clear before users confirm deletion.

The passive-cluster note helps users who intentionally want to target a passive cluster avoid the default frontend forwarding behavior.

## Checklist

1. Closes 

2. How was this tested:

unit-test and local run.

single deletion
```
go run ./cmd/temporal  workflow delete -w pay-retention-92728
WARNING: Deleting Workflow Executions in a global Namespace removes them from all replicas. Requests sent to a passive cluster are forwarded to the active cluster by default; to target the passive cluster directly, specify `--grpc-meta xdc-redirection=false`.
Delete Workflow "pay-retention-92728"? y/N
Error: user denied confirmation
exit status 1
```

batch deletion
```
go run ./cmd/temporal workflow delete --query "WorkflowId = 'pay-retention-92728'"

WARNING: Deleting Workflow Executions in a global Namespace removes them from all replicas. Requests sent to a passive cluster are forwarded to the active cluster by default; to target the passive cluster directly, specify `--grpc-meta xdc-redirection=false`.
Start batch against approximately 1 workflow(s)? y/N y
Started batch for job ID: ddab94d3-719f-4a4a-b379-12cdd8229e36
```

3. Any docs updates needed?


yux0 (approved):

temporal-nick (approved):

internal/temporalcli/commands.workflow.go
```diff
@@ -69,13 +71,23 @@ func (c *TemporalWorkflowDeleteCommand) run(cctx *CommandContext, args []string)
...
+	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, singleOrBatchOverrides{
+		AllowYesWithWorkflowID: true,
+	})
 
 	// Run single or batch
 	if err != nil {
 		return err
 	} else if exec != nil {
-		_, err := cl.WorkflowService().DeleteWorkflowExecution(cctx, &workflowservice.DeleteWorkflowExecutionRequest{
+		yes, err := cctx.promptYes(workflowDeleteSingleConfirmationMessage(exec), c.Yes)
```
fretz12, line 90: Prompting would be a breaking change if used in a script. Can you follow this in the PR title/description?
```
Breaking changes must be treated with high review rigor, marked with 💥 (:boom:) in commit messages and the PR title/description, and called out with 💥 in the release notes with a clear explanation.
```
FYI we're in process of [standardizing this](https://www.notion.so/Temporal-CLI-Design-Principles-3578fc5677388062823cc1d0868dbbbd?d=3618fc5677388027be70001c500d621c&source=copy_link#3578fc56773881518c73d844a81bd465)

internal/temporalcli/commands.workflow.go
```diff
@@ -69,13 +71,23 @@ func (c *TemporalWorkflowDeleteCommand) run(cctx *CommandContext, args []string)
 	}
 	defer cl.Close()
 
-	exec, batchReq, err := c.workflowExecOrBatch(cctx, c.Parent.Namespace, cl, singleOrBatchOverrides{})
+	cctx.Printer.Println(workflowDeleteWarning)
```
fretz12: This will print for all namespace types including non-global ones. I'd suggest either
- Making the message more specific, i.e. "WARNING: If this Namespace is global, deleting Workflow Executions in the namespace removes them ..."
- Or better UX at the cost of extra RPC: Call DescribeNamespace, check GetIsGlobalNamespace and print warning conditionally

internal/temporalcli/commands.workflow.go
```diff
@@ -96,6 +108,18 @@ func (c *TemporalWorkflowDeleteCommand) run(cctx *CommandContext, args []string)
 	return nil
 }
 
+func workflowDeleteConfirmationMessage(action string) string {
```
fretz12: nit: we can just inline this

internal/temporalcli/commands.workflow.go
fretz12: Currently the warning goes to stdout, so `--output json` callers get a non-JSON line with their output. Send to stderr for warnings:
```suggestion
	fmt.Fprintln(cctx.Options.Stderr, workflowDeleteWarning)
```

harani-mukkala: +1 on this, would it be possible to show this warning only for replicated users in a easier way? Users running a single cluster might be confused with this warning.

internal/temporalcli/commands.workflow.go
jiechenz: Updated

internal/temporalcli/commands.workflow.go
jiechenz: Good point! Updated to check on global namespace.

fretz12 (approved):

PR #1046, opened 2026-05-21
Add support for Standalone Nexus Operations

Quinn-With-Two-Ns:
## NOTE: REQUIRES OSS SERVER VERSION v1.32.0

## What changed?

Add support for stand alone nexus operations to the CLI.

## Checklist

<!-- Your PR should satisfy all these requirements. However, feel free to remove items that don't apply to the PR. Consider giving this checklist to an AI agent before opening your PR. -->

**Stability**
- [ ] Breaking changes are marked with 💥 in the PR title and release notes
- [ ] Changes to JSON output (`-o json` / `-o jsonl`) are treated as breaking changes

**Design**
- [x] New commands follow `temporal <noun> <verb>` structure (e.g. `temporal workflow start`)
- [x] New flags are named after the API concept, not the implementation mechanism (good: `--search-attribute`, bad: `--index-field`)
- [x] New flags don't duplicate an existing flag that serves the same purpose
- [ ] New flags do not have short aliases without strong justification
- [x] Experimental features are marked with `(Experimental)` in `commands.yaml`

- [x] All flags shown in help text and examples are implemented and functional
- [x] Summaries use sentence case and have no trailing period
- [x] Long descriptions end with a period and include at least one example invocation
- [ ] Examples use long flags (`--namespace`, not `-n`), one flag per line
- [x] Placeholder values use `YourXxx` form (`YourWorkflowId`, `YourNamespace`)

**Behavior**

**Tests**
- [x] Added unit test(s) (`func TestXxx`) where applicable

## Manual tests

<!-- Edit the code samples below to provide setup and happy-path and error-path testing instructions. -->

**Setup**
```
  temporal operator nexus endpoint create                                                                                                  
      --name my-endpoint                                                                                                                 
      --target-namespace default
      --target-task-queue my-tq
```

  **Happy path**
```
  $ temporal nexus operation start                                                                                                         
      --endpoint my-endpoint      
      --service my-service                                                                                                                 
      --operation my-op                                                                                                                  
      --operation-id my-op-1
      --input '"hello"'     
  Started Nexus Operation:                                                                                                                 
    Endpoint    my-endpoint
    Service     my-service                                                                                                                 
    Operation   my-op                                                                                                                    
    OperationId my-op-1
    RunId       7e7c8b2a-...-...
    Namespace   default
```
                                                                                                                                           
  **Error case**
```
  $ temporal nexus operation start                                                                                                         
      --service my-service                                                                                                                 
      --operation my-op   
      --operation-id my-op-2                                                                                                               
  Error: required flag(s) "endpoint" not set                                                                                             
 ```

  **Composition** — start an operation, then fetch its result, then list/describe:        
```                                                 
  $ temporal nexus operation start
      --endpoint my-endpoint                                                                                                               
      --service my-service                                                                                                               
      --operation my-op                                                                                                                    
      --operation-id my-op-3
      --input '"world"'                                                                                                                    
                                                                                                                                         
  $ temporal nexus operation result --operation-id my-op-3
  Results:                                                                                                                                 
    Status  COMPLETED
    Result  "got: world"                                                                                                                   
                                                                                                                                         
  $ temporal nexus operation describe --operation-id my-op-3
    OperationId            my-op-3
    RunId                  7e7c8b2a-...-...                                                                                                
    Endpoint               my-endpoint
    Service                my-service                                                                                                      
    Operation              my-op                                                                                                         
    Status                 Succeeded                                                                                                       
    ...
```

### CODE REVIEW

File: internal/temporalcli/commands.nexus_operation.go [hunk 1]
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	switch v := resp.GetOutcome().(type) {
+	case *workflowservice.PollNexusOperationExecutionResponse_Result:
+		return printNexusOperationResult(cctx, operationID, runID, v.Result)
+	case *workflowservice.PollNexusOperationExecutionResponse_Failure:
+		if err := printNexusOperationFailure(cctx, operationID, runID, v.Failure); err != nil {
+			cctx.Logger.Error("Nexus operation failed, and printing the output also failed", "error", err)
+		}
+		return fmt.Errorf("nexus operation failed")
```
Evanthx, line 129: Should Nexus be capitalized? Same for line 112 above and 131, 492 below. Didn't do a complete check for places, but looking at the code for "fmt.Errorf.*nexus" I see it both ways!

internal/temporalcli/commands.nexus_operation.go
```diff
@@ -0,0 +1,503 @@
...
+	default:
+		return fmt.Errorf("unexpected nexus operation outcome type: %T", v)
+	}
+}
+
+// Matches the SDK's pollActivityTimeout in internal_activity_client.go.
+const pollNexusOperationTimeout = 60 * time.Second
+
+// pollNexusOperationOutcome polls for a nexus operation result using a
```
Evanthx, line 138: Feel free to ignore this comment - would this be better as a method on Handle then, both for reuse and so that if someone made a change to Handle.Get later this method would be visible there to get the equivalent change? ie, maybe Handle would then have a Get and a RawGet method, something like that?

internal/temporalcli/commands.nexus_operation.go
```diff
@@ -0,0 +1,503 @@
...
+}
+
+func printNexusOperationResult(cctx *CommandContext, operationID, runID string, result *common.Payload) error {
+	if cctx.JSONOutput {
+		var resultJSON json.RawMessage
+		var err error
+		if cctx.JSONShorthandPayloads {
+			var valuePtr any
+			if err = converter.GetDefaultDataConverter().FromPayload(result, &valuePtr); err != nil {
+				return fmt.Errorf("failed decoding result: %w", err)
```
Evanthx: Might add a little more context here - ie, Nexus operation completed, but failed decoding result, something like that?

Same for the other errors being returned below.

internal/temporalcli/commands.nexus_operation.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	cctx.Printer.Println(color.MagentaString("Results:"))
+	var valuePtr any
+	if err := converter.GetDefaultDataConverter().FromPayload(result, &valuePtr); err != nil {
+		return fmt.Errorf("failed decoding result: %w", err)
+	}
+	resultJSON, err := json.Marshal(valuePtr)
+	if err != nil {
+		return fmt.Errorf("failed marshaling result: %w", err)
```
Evanthx: Line 202 and 210 return the exact same string as 177 and 184. So if we get that error, we won't actually know which line failed, just that it's one of those two. Given that they are doing the same thing to the same object it might not matter much, but it's nice to just go to the offending log line without having to wonder which line it was.

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+        - cli reference
+        - cli-feature
+        - command-line-interface-cli
+        - temporal cli
+      tags:
+        - Nexus
+        - Temporal CLI
+
+  - name: temporal nexus operation
+    summary: Commands for managing Nexus Operations
```
VegetarianOrc, line 1662:
```suggestion
    summary: Commands for managing Standalone Nexus Operations
```

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+    description: |
+      These commands manage standalone Nexus Operation Executions.
```
VegetarianOrc:
```suggestion
      These commands manage Standalone Nexus Operation Executions.
```

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+      Nexus Operation commands follow this syntax:
+
+      ```
+      temporal nexus operation [command] [options]
+      ```
+
+  - name: temporal nexus operation start
+    summary: Start a new Standalone Nexus Operation (Experimental)
+    description: |
+      Start a new Standalone Nexus Operation Execution. Outputs the
```
VegetarianOrc:
```suggestion
      Start a new Standalone Nexus Operation. Outputs the
```

internal/temporalcli/commands.nexus_operation.go
```diff
@@ -0,0 +1,503 @@
...
+	}, printer.StructuredOptions{})
+}
+
+func printNexusOperationFailure(cctx *CommandContext, operationID, runID string, f *failure.Failure) error {
+	if cctx.JSONOutput {
+		failureJSON, err := cctx.MarshalProtoJSON(f)
+		if err != nil {
+			return fmt.Errorf("failed marshaling failure: %w", err)
+		}
+		_ = cctx.Printer.PrintStructured(struct {
```
Evanthx: printNexusOperationResult gets the error from PrintStructured and returns it (line 208), this ignores any error values from PrintStructured and returns nil. Should they both behave the same way?

internal/temporalcli/commands.yaml
```diff
@@ -5025,6 +5229,58 @@ option-sets:
...
+        required: true
+      - name: operation
+        type: string
+        description: Nexus Operation name.
+        required: true
+      - name: operation-id
+        type: string
+        description: |
+          Nexus Operation ID.
+          If not supplied, the Server generates a unique ID.
```
VegetarianOrc: Nit, this is generated client side.
```suggestion
          If not supplied, a unique ID is generated.
```

internal/temporalcli/commands.nexus_operation_test.go
```diff
@@ -0,0 +1,786 @@
...
+		)
+		return res.Err == nil
+	}, 30*time.Second, 500*time.Millisecond)
+
+	s.NoError(res.Err)
+	s.Contains(res.Stdout.String(), opID)
+}
+
+func (s *SharedServerSuite) TestNexusOperationCancel() {
+	endpointName := "test-ep-" + uuid.NewString()[:8]
```
Evanthx: TestNexusOperationCancel and TestNexusOperationTerminate both set up endpoints - can that be combined into a helper function using setupNexusEndpointAndWorker?

Evanthx (approved):
Found a few nit-picks! Your call on whether they are worth addressing, they are all quite minor.

internal/temporalcli/commands.nexus_operation.go
```diff
@@ -0,0 +1,503 @@
...
+		if err := printNexusOperationFailure(cctx, operationID, runID, v.Failure); err != nil {
+			cctx.Logger.Error("Nexus operation failed, and printing the output also failed", "error", err)
+		}
+		return fmt.Errorf("nexus operation failed")
+	default:
+		return fmt.Errorf("unexpected nexus operation outcome type: %T", v)
+	}
+}
+
+// Matches the SDK's pollActivityTimeout in internal_activity_client.go.
```
VegetarianOrc:
```suggestion
// Matches the SDK's pollNexusOperationTimeout in internal_nexus_client.go.
```

VegetarianOrc (approved):
A few minor nits, but looks good to me!

internal/temporalcli/commands.nexus_operation.go
Quinn-With-Two-Ns: in `errors` it should be lower case based on convention, same as workflow and logs it may be uppercase

internal/temporalcli/commands.nexus_operation.go
Quinn-With-Two-Ns: Done

internal/temporalcli/commands.nexus_operation.go
Quinn-With-Two-Ns: Tried to make the messages more distinct

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+
+      ```
+      temporal nexus operation [command] [options]
+      ```
+
+  - name: temporal nexus operation start
+    summary: Start a new Standalone Nexus Operation (Experimental)
+    description: |
+      Start a new Standalone Nexus Operation. Outputs the
+      Operation ID and Run ID.
```
chaptersix:
```suggestion
      Start a Nexus Operation. Outputs the
      Operation ID and Run ID.
```

SANO is an internal term right? SAA is just Activity.

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+
+      ```
+      temporal nexus operation start \
+          --endpoint YourEndpoint \
+          --service YourService \
+          --operation YourOperation \
+          --operation-id YourOperationId \
+          --input '{"some-key": "some-value"}'
```
chaptersix, line 1684: from the cli, how does a user discover the proper service endpoint and operation?

internal/temporalcli/commands_test.go
```diff
@@ -240,6 +240,8 @@ func (s *SharedServerSuite) SetupSuite() {
 				"history.enableSignalWithStartFromWorkflow": true,
 				"activity.enableStandalone":                 true,
 				"activity.longPollTimeout":                  2 * time.Second,
+				"nexusoperation.enableStandalone":           true,
+				"history.enableChasmCallbacks":              true,
```
chaptersix, line 244: I we should somehow track when we can disable these overrides. some of these could be removed.

internal/temporalcli/commands.yaml
```diff
@@ -1619,6 +1619,210 @@ commands:
...
+
+      ```
+      temporal nexus operation list
+      ```
+    option-sets:
+      - client
+    docs:
+      description-header: >-
+        Learn how to use Temporal Nexus commands for starting, listing,
+        and managing Standalone Nexus Operation Executions.
```
chaptersix: I don't think this description quite matches the commands. for you to learn about how to use endpoints. there already has to be executions.

chaptersix (approved):
I don't want to block you based on these but makes corrections if needed.

chaptersix:
Also, could you note the min server version required for these commands? (Just to make sure we don't release these changes before the next OSS release)

internal/temporalcli/commands.yaml
Quinn-With-Two-Ns: The user has to know that, same as to start an activity the user has to know the task queue and activity type

internal/temporalcli/commands.yaml
Quinn-With-Two-Ns: Yeah we should be consistent with SAA here

internal/temporalcli/commands.yaml
Quinn-With-Two-Ns: Talked offline

PR #1048, opened 2026-05-23
Fix/schedule search attributes text mode

### LINKED ISSUE #590: [Bug] Typed search attributes inside schedule actions are not printed correctly in text mode

As the description says.

This may apply in other situations, too—any situation in which we wrap typed search attributes inside another object in the CLI code would be impacted. Also, this is worse with server 1.24 because that server now returns schedule search attributes as typed search attributes. So search attributes that would have been returned from 1.23 and earlier are now missing from CLI output.

This can be worked around by using JSON output instead (`-o json`), which prints the attributes correctly.

### Describe the bug

Typed search attributes are printed correctly when they are [printed as part of a proto.Message](https://github.com/temporalio/cli/blob/1f6323192eb5b66c3c406f51369bb94e5015cbfe/temporalcli/internal/printer/printer.go#L253), but incorrectly when [printed as part of a normal JSON object](https://github.com/temporalio/cli/blob/1f6323192eb5b66c3c406f51369bb94e5015cbfe/temporalcli/internal/printer/printer.go#L260). This is likely because there is no [Marshaler](https://pkg.go.dev/encoding/json#Marshaler) implementation for [SearchAttributes](https://github.com/temporalio/sdk-go/blob/5bff6f8c47f3bf8491934964ad5044ce150d59fd/internal/internal_search_attributes.go#L40) in the Go SDK.

We don't necessarily want to add one, either, as this might be unnecessary for most people, so we should try to come up with a workaround for this in the CLI instead.

### Minimal Reproduction

Launch a 1.24.1 server (a CLI which uses 1.24.1 can be built from #584):

```
./temporal server start-dev
```

Then create a search attribute and schedule which uses the search attribute:

```
❯ ./temporal operator search-attribute create --name CustomKeywordField --type Keyword
Search attributes have been added

❯ ./temporal schedule create -s foo --task-queue foo --type foo --workflow-id asdf --interval 10d --search-attribute CustomKeywordField=\"schedule-string-val\"
```

When describing the schedule in text mode, notice the `Action`'s `TypedSearchAttributes` field is empty:

```
❯ ./temporal schedule describe -s foo
  ScheduleId        foo
  Action            {"ID":"asdf","Workflow":"foo","Args":[],"TaskQueue":"foo","WorkflowExecutionTimeout":0,"WorkflowRunTimeout":0,"WorkflowTaskTimeout":10000000000,"RetryPolicy":null,"Memo":{},"TypedSearchAttributes":{},"UntypedSearchAttributes":{}}
  Spec              [{"every":"10d 0h 0m 0s"}]
  Jitter            0s
  OverlapPolicy     Skip
  CatchupWindow     365d 0h 0m 0s
  PauseOnFailure    false
  Paused            false
  NextRunTime       1 week from now
  LastRunTime
  RunningWorkflows  []
  CreatedAt         1 minute ago
  LastUpdateAt      a long while ago
  ActionCounts      {"Total":0,"MissedCatchupWindow":0,"SkippedOverlap":0}
```

However, the schedule prints correctly in JSON mode:

```
❯ ./temporal schedule describe -s foo -o json
{
  "schedule": {
...
    "action": {
      "startWorkflow": {
        "workflowId": "asdf",
        "workflowType": {
          "name": "foo"
        },
        "taskQueue": {
          "name": "foo",
          "kind": "TASK_QUEUE_KIND_NORMAL"
        },
        "input": {},
        "workflowExecutionTimeout": "0s",
        "workflowRunTimeout": "0s",
        "workflowTaskTimeout": "10s",
        "searchAttributes": {
          "indexedFields": {
            "CustomKeywordField": {
              "metadata": {
                "encoding": "anNvbi9wbGFpbg==",
                "type": "S2V5d29yZA=="
              },
              "data": "InNjaGVkdWxlLXN0cmluZy12YWwi"
            }
          }
        },
        "header": {}
      }
    },
    "policies": {
      "overlapPolicy": "SCHEDULE_OVERLAP_POLICY_SKIP",
      "catchupWindow": "31536000s"
    },
    "state": {}
  },
...
}
```

### Environment/Versions

- OS and processor: any
- Temporal Version: CLI 0.12 and Server 1.24
- Are you using Docker or Kubernetes or building Temporal from source? source, as above

NasitSony:
## What was changed
Convert `SearchAttributes` field in `printableSchedule` from `*commonpb.SearchAttributes` to `map[string]interface{}`. Added `searchAttributesToMap` helper that uses `temporalproto.CustomJSONMarshalOptions` to correctly serialize typed search attributes.

## Why?
Typed search attributes were printing as empty `{}` in text mode because standard `json.Marshal` doesn't know how to serialize `*commonpb.SearchAttributes` typed fields correctly. JSON mode worked because it used the proto marshaler. This fix ensures text mode uses the same proto-aware marshaling.

## Checklist
1. Closes #590

2. How was this tested: 
All existing tests pass (go test ./...). The fix applies to both describeResultToPrintable and listEntryToPrintable to cover both describe and list commands.

3. Any docs updates needed? 
No — this is a bug fix with no user-facing API changes.

### DIFF
--- a/internal/temporalcli/commands.gen.go
+++ b/internal/temporalcli/commands.gen.go
@@ -2284,8 +2284,12 @@ func NewTemporalScheduleListMatchingTimesCommand(cctx *CommandContext, parent *T
 	s.Parent = parent
 	s.Command.DisableFlagsInUseLine = true
 	s.Command.Use = "list-matching-times [flags]"
-	s.Command.Short = "List matching times for a Schedule"
-	s.Command.Long = "List the times a Schedule would fire within a given time range.\nUse this command to preview when a Schedule will trigger Workflow\nExecutions without actually running them.\n\nFor example:\ntemporal schedule list-matching-times \\\n--schedule-id \"YourScheduleId\" \\\n--start-time \"2024-01-01T00:00:00Z\" \\\n--end-time \"2024-01-31T23:59:59Z\""
+	s.Command.Short = "List matching times for a Schedule (Experimental)"
+	if hasHighlighting {
+		s.Command.Long = "\nNote: This is an experimental feature and may change in the future.\n\nList the times a Schedule would fire within a given time range.\nUse this command to preview when a Schedule will trigger Workflow\nExecutions without actually running them.\n\nFor example:\n\n\x1b[1m  temporal schedule list-matching-times \\\n    --schedule-id \"YourScheduleId\" \\\n    --start-time \"2024-01-01T00:00:00Z\" \\\n    --end-time \"2024-01-31T23:59:59Z\"\x1b[0m"
+	} else {
+		s.Command.Long = "\nNote: This is an experimental feature and may change in the future.\n\nList the times a Schedule would fire within a given time range.\nUse this command to preview when a Schedule will trigger Workflow\nExecutions without actually running them.\n\nFor example:\n\n```\n  temporal schedule list-matching-times \\\n    --schedule-id \"YourScheduleId\" \\\n    --start-time \"2024-01-01T00:00:00Z\" \\\n    --end-time \"2024-01-31T23:59:59Z\"\n```"
+	}
 	s.Command.Args = cobra.NoArgs
 	s.Command.Flags().Var(&s.StartTime, "start-time", "Start of time range to list matching times. Required.")
 	_ = cobra.MarkFlagRequired(s.Command.Flags(), "start-time")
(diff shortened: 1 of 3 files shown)


chaptersix:
@NasitSony looks like this PR has some of your changes from another PR.

NasitSony:
@chaptersix Closing in favor of #1060 — cleaner branch without extra commits.

PR #1057, opened 2026-05-29
fix: use allow instead of ignore for dependency-type in dependabot config

chaptersix:
## Summary
- `dependency-type` is only valid inside `allow`, not `ignore` in dependabot.yml
- Replaces the invalid `ignore` entry with `allow: [{dependency-type: direct}]` to achieve the same goal of skipping indirect/transitive dependency updates
- Adds a CI workflow using `check-jsonschema` to validate `dependabot.yml` on PRs and pushes to main
- Fixes the Dependabot config validation failure introduced in #1044

## Test plan
- [ ] Verify the `.github/dependabot.yml` validation check passes on this PR
- [ ] Verify `check-jsonschema --builtin-schema vendor.dependabot .github/dependabot.yml` passes locally


ks-temporal (approved):
Looks fine. One question is whether you want to specify fixed versions for check-jsonschema and python, or leave it unspecified for latest.

PR #1070, opened 2026-06-01
Use proxy.SystemNexusOperations for nested payload decoding

chaptersix:
## Summary
- Replace CLI's local `systemNexusOps` registry with `proxy.SystemNexusOperations` from `go.temporal.io/api`
- Decode payloads nested inside system Nexus operation bytes in the human-readable display path only
- Preserve `-o json` output unchanged for SDK replay compatibility
- Expose `PayloadCodec` on `cliext.ClientOptionsBuilder` so the CLI can pass the codec to the display-path decoder

Stacked on #1017. Depends on temporalio/api-go#281.

## Test plan
- [x] Unit tests for unwrap/inject request and response
- [x] Unit tests for codec application to nested payloads
- [x] Unit tests for display name mapping
- [x] Existing functional tests pass


chaptersix:
closing. was just for demo purposes

PR #1071, opened 2026-06-01
fix: add guidance when search-attribute create hits per-type limit

### LINKED ISSUE #318: [Feature Request] Improve DX of adding Search Attributes

Can easily run into this error:

```
t operator search-attribute create --namespace default --name test4 --type Text
Error: unable to add search attributes: Unable to create search attributes: cannot have more than 3 search attribute of type Text.
```

- Higher default max
- Say in the error how to raise the limit (or link to an error page that says it)

amit-writes-code:
## Related issues

Closes #318 (partial)

This PR addresses **1 of the 3** approaches discussed for issue #318:

1. ⬜ Raise the defaults for the dev server — not done (server-side schema concern, see note below)
2. ⬜ Expose a CLI flag like `--custom-search-attributes-limit` — not done
3. ✅ **Improve the error message when hitting the limit** — done in this PR

## What changed?

`temporal operator search-attribute create` previously surfaced the server's
per-type limit failure as a bare `unable to add search attributes: ... cannot
have more than 3 search attribute of type Text.`, which gives the user no idea
why the limit exists or what to do next.

With SQL-based Visibility, custom search attributes are mapped onto a fixed pool
of pre-allocated fields per type, so this limit cannot be raised at runtime. When
this specific error is detected, the command now appends a short hint: remove an
unused attribute of that type, reuse an existing one, choose a different type, or
use Elasticsearch-based Visibility (which has no per-type limit), plus a docs
link. All other failures are forwarded unchanged.

**Scope note:** issue #318 also suggests raising the default max (approaches 1 and 2
above). The per-type pool size is fixed in the visibility schema, not a runtime
setting, so raising it is a server-side schema concern rather than something the
CLI can pass through. This PR therefore intentionally addresses only the
error-message portion (approach 3); the other two can be tackled separately.
## Checklist

**Stability**
- [x] Breaking changes are marked with 💥 in the PR title and release notes — n/a, not breaking
- [x] Changes to JSON output (`-o json` / `-o jsonl`) are treated as breaking changes — n/a, no JSON structure change

**Design**

_(Remaining Design items removed — no new commands or flags.)_

**Help text**

_(Removed — no changes to `commands.yaml` or help text; the change is in the command's `run` error path.)_

**Behavior**
- [x] Error messages are lowercase with no trailing punctuation — the CLI-added clause is lowercase and ends on the docs URL; the leading capital/period belong to the wrapped upstream server error (unchanged)

**Tests**
- [x] Added functional test(s) (`SharedServerSuite`) — `TestOperator_SearchAttribute_LimitGuidance`
- [ ] Added unit test(s) (`func TestXxx`) where applicable — covered by the functional test

## Manual tests

**Setup**
```
temporal server start-dev --headless
```

**Happy path**
```
$ temporal operator search-attribute create \
    --name MyText1 \
    --type Text
Search attributes have been added
```

**Error case**
```
$ for i in 1 2 3 4; do
    ./temporal operator search-attribute create --name "MyText$i" --type Text
  done
Search attributes have been added
Search attributes have been added
Search attributes have been added
Error: unable to add search attributes: Unable to create search attributes: cannot have more than 3 search attribute of type Text.; this per-type limit is fixed for sql-based visibility and cannot be raised at runtime, so remove an unused attribute of this type, reuse one, choose a different type, or use elasticsearch visibility which has no per-type limit: https://docs.temporal.io/visibility
$ echo $?
1
```

**Composition**
```
$ temporal operator search-attribute create --name MyText1 --type Text
$ temporal operator search-attribute list
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,9 +6,9 @@ updates:
       interval: weekly
     cooldown:
       default-days: 14
+    allow:
+      - dependency-type: direct
     ignore:
-      - dependency-name: "*"
-        dependency-type: indirect
       # Temporal dependencies are managed manually to ensure coordinated upgrades across all temporal packages
       - dependency-name: "go.temporal.io/*"
 
--- a/.github/workflows/validate-dependabot.yml
+++ b/.github/workflows/validate-dependabot.yml
@@ -1,0 +1,29 @@
+name: Validate Dependabot Config
+on:
+  pull_request:
+    paths:
+      - '.github/dependabot.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/dependabot.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  validate:
+    name: Validate Dependabot Config
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: '3.12'
+
+      - name: Validate dependabot.yml
+        run: |
+          pip install check-jsonschema==0.37.2
+          check-jsonschema --builtin-schema vendor.dependabot .github/dependabot.yml


amit-writes-code:
@chaptersix seems the build is taking forever to run, I squashed 2 commits into 1 could that be the reason?

PR #1076, opened 2026-06-02
ci/docs: escape MDX-incompatible patterns in gen-docs output

lennessyy:
## Summary

- Add `escapeMDXDescription()` to escape patterns in command descriptions that break Docusaurus MDX compilation:
  - **Heading IDs** (`{#custom-id}`) — converted to MDX-compatible `{/* #custom-id */}` comment syntax, preserving custom anchors
  - **Angle-bracket placeholders** (`<base64-encoded-cert>`, `<key>`, etc.) — escaped with backslashes so MDX doesn't interpret them as HTML tags (lowercase-only to avoid escaping real HTML tags)
  - **Curly braces in JSON examples** (`'YourKey={"your": "value"}'`) — escaped so MDX doesn't interpret them as JSX expressions
- Code fences and inline code (backtick spans) are left untouched
- Fix `encodeJSONExample()` regex to correctly match patterns like `'YourKey={"your": "value"}'` (the previous regex was non-greedy and failed on these)
- Align the JSON escaping regex between `escapeMDXDescription` and `encodeJSONExample` so both handle nested JSON like `'{"a": {"b": "c"}}'`
- Add cross-reference comments between the two escaping functions so maintainers understand the split: `encodeJSONExample` handles option description table cells, `escapeMDXDescription` handles command description body text
- No changes to `commands.yaml` — the escaping is applied automatically during generation

## Context

The documentation site uses Docusaurus with MDX, which has a stricter parser than standard Markdown. New CLI commands added since early 2026 (task queue fairness weights, cloud connectivity rules, etc.) introduced angle-bracket placeholders and JSON examples in descriptions that MDX interprets as JSX. The `{#id}` heading syntax also breaks under Docusaurus 3.10's MDX parser. These patterns worked in standard Markdown but break the docs site build.

## Test plan

- [x] Regenerated docs from latest CLI + cloud-cli and verified clean Docusaurus build
- [x] Verified code fences, inline backtick spans, and real HTML tags are not affected
- [x] Verified heading IDs are converted to working `{/* #id */}` anchors

🤖 Generated with [Claude Code](https://claude.com/claude-code)


harani-mukkala (approved):

jsundai:
Thanks, [harani-mukkala](https://github.com/harani-mukkala)! 🙌

chaptersix:
We should consider by PR first.

PR #1094, opened 2026-06-21
test: fix concurrent start test assertions

### LINKED ISSUE #581: [Bug] Dev Server concurrent execution test is incorrect and flaky

nanookclaw:
## Summary

Fixes #581 by removing `testing.T` assertions from the worker goroutines in `TestServer_StartDev_ConcurrentStarts`. Each concurrent start now returns an `error`, and the main test goroutine collects those errors and reports them with `require.NoError`.

The test still starts 40 dev server instances with a fixed worker pool of 6 concurrent executions. The replacement command helper runs `server start-dev` with a cancellable context and captured stdout/stderr, so failures from early command exit or cleanup are preserved in the final assertion message without calling `FailNow`-style methods outside the test goroutine.

## Verification

- `gofmt -w internal/temporalcli/commands.server_test.go`
- `git diff --check`
- `go test ./internal/temporalcli -run '^TestServer_StartDev_ConcurrentStarts$' -count=1`


internal/temporalcli/commands.server_test.go
```diff
@@ -139,58 +142,120 @@ func startDevServerAndRunSimpleTest(t *testing.T, args []string, dialAddress str
 }
 
 func TestServer_StartDev_ConcurrentStarts(t *testing.T) {
-	startOne := func() {
-		h := NewCommandHarness(t)
-		defer h.Close()
+	startOne := func() error {
+		ctx, cancel := context.WithCancel(context.Background())
```
chaptersix: @nanookclaw I like the approach. let's use `t.Context()` instead of `context.Background()`

internal/temporalcli/commands.server_test.go
```diff
@@ -139,58 +142,120 @@ func startDevServerAndRunSimpleTest(t *testing.T, args []string, dialAddress str
...
 		}()
 	}
 	wg.Wait()
+	close(errCh)
+	for err := range errCh {
+		require.NoError(t, err)
+	}
+}
+
+func executeConcurrentStartCommand(ctx context.Context, args ...string) *CommandResult {
```
chaptersix: let's add `func (h *CommandHarness) ExecuteWithContext(ctx context.Context, args ...string) *CommandResult {` 

so we can keep using the harness

internal/temporalcli/commands.server_test.go
```diff
@@ -139,58 +142,120 @@ func startDevServerAndRunSimpleTest(t *testing.T, args []string, dialAddress str
...
 		go func() {
-			resCh <- h.Execute("server", "start-dev", "-p", port, "--http-port", httpPort, "--headless", "--log-level", "never")
+			resCh <- executeConcurrentStartCommand(ctx, "server", "start-dev", "-p", port, "--http-port", httpPort, "--headless", "--log-level", "never")
 		}()
 
 		// Try to connect for a bit while checking for error
 		var cl client.Client
-		h.EventuallyWithT(func(t *assert.CollectT) {
+		var lastDialErr error
+		for start := time.Now(); time.Since(start) < 3*time.Second; {
```
chaptersix: how about something like this:

```
 ticker := time.NewTicker(200 * time.Millisecond)
  defer ticker.Stop()
  timeout := time.NewTimer(3 * time.Second)
  defer timeout.Stop()

  outer:
  for {
      select {
      case res := <-resCh:
          return concurrentStartCommandResultError("got early server result", res)
      case <-timeout.C:
          break outer
      case <-ticker.C:
          var err error
          cl, err = client.Dial(client.Options{HostPort: "127.0.0.1:" + port})
          if err == nil {
              break outer
          }
          lastDialErr = err
      }
  }
```

nanookclaw:
Updated in ce9c1a5 to address the review comments:

- added `CommandHarness.ExecuteWithContext(ctx, args...)` while preserving existing `Execute` behavior
- changed the concurrent start test to use `t.Context()` and the harness
- replaced the readiness sleep loop with a ticker/timer select
- removed the bespoke concurrent-start command helper

Local verification:

- `go test ./internal/temporalcli -run '^$'`
- `go test ./internal/temporalcli -run '^TestServer_StartDev_ConcurrentStarts$' -count=1`

chaptersix:
@nanookclaw not sure if you still want to work on this PR but it LGTM. just need to fix the build failures.

chaptersix:
> @nanookclaw not sure if you still want to work on this PR but it LGTM. just need to fix the build failures.

CI has been a pain recently

PR #1100, opened 2026-06-25
Add batch cancel, terminate and delete operators for standalone activities

ks-temporal:
## What changed?

Add new activity delete subcommand for deleting a standalone activity, and
add support for --query and related flags for batch operations in activity
cancel, terminate and delete for standalone activities. These are marked
experimental. Exactly one of --activity-id or --query is required for these
activity operators.

## Checklist

**Stability**
- [ ] (N/A) Breaking changes are marked with 💥 in the PR title and release notes
- [ ] (N/A) Changes to JSON output (`-o json` / `-o jsonl`) are treated as breaking changes

**Design**
- [ ] (N/A) This feature does not depend on Cloud-only APIs or behavior (it works against an OSS server)
- [ ] (N/A) New commands follow `temporal <noun> <verb>` structure (e.g. `temporal workflow start`)
- [x] New flags are named after the API concept, not the implementation mechanism (good: `--search-attribute`, bad: `--index-field`)
- [x] New flags don't duplicate an existing flag that serves the same purpose
- [x] New flags do not have short aliases without strong justification (Justification for -q and -y: consistency)
- [x] Experimental features are marked with `(Experimental)` in `commands.yaml`

- [x] All flags shown in help text and examples are implemented and functional
- [x] Summaries use sentence case and have no trailing period
- [x] Long descriptions end with a period and include at least one example invocation
- [x] Examples use long flags (`--namespace`, not `-n`), one flag per line
- [x] Placeholder values use `YourXxx` form (`YourWorkflowId`, `YourNamespace`)

**Behavior**

**Tests**
- [ ] Added unit test(s) (`func TestXxx`) where applicable

## Manual tests

**Setup**
Start N=5 activities replacing `{N}` with 1, 2, ... and so on.
```
temporal server start-dev --headless
temporal activity start \
    --start-to-close-timeout 60s \
    --type YourActivityType \
    --task-queue YourTaskQueue \
    --activity-id YourActivityId{N}
```
Start a worker with type YourActivityType to listen to task queue of YourTaskQueue,
that either sleeps for 2 minutes, or returns activity pending error.

Repeat the setup step for each of the tests below.

**Happy path**
First, test batch cancellation, as follows, and enter `y` when prompted.
```
$ temporal activity cancel \
    --query 'ActivityType = "YourActivityType"'
Start batch against approximately 5 standalone activities(s)? y/N y
Started batch for job ID: ...
```

Check the RunState of these activities using describe, replacing `{N}` with 1,2,... and so on.
```
$ temporal activity describe \
    --activity-id YourActivityId{N}
...
Status                 Running
RunState            CancelRequested
...
```

Also check the status in the list as follows.
```
$ temporal activity list \
    --query 'ActivityType = "YourActivityType"'
...
```

Next, test batch deletion, as follows, and enter `y` when prompted.
```
$ temporal activity delete \
    --query  'ActivityType = "YourActivityType"'
Start batch against approximately 5 standalone activities(s)? y/N y
Started batch for job ID: ...
```

Now, listing the activities should not show any.
```
$ temporal activity list \
    --query 'ActivityType = "YourActivityType"'
```

To test, batch termination, start the activities again using the setup step
mentioned above. Then test batch termination as follows. Here `--yes` is
used to avoid the prompt.
```
$ temporal activity terminate \
    --query  'ActivityType = "YourActivityType"' \
    --yes
Start batch against standalone activities matching query "ActivityType=\"YourActivityType\""? y/N yes
Started batch for job ID: ...
```
Now, listing the activities should show Status of Terminated for all.
```
$ temporal activity list \
    --query 'ActivityType = "YourActivityType"'
    Status ...
  Terminated ...
  ...
```

Delete the activities to cleanup.
```
$ temporal activity delete \
    --query  'ActivityType = "YourActivityType"' \
    --yes
...
```

**Error case**
The following can be repeated by replacing `cancel` with `terminate` or `delete`.
```
$ temporal activity cancel \
    --activity-id YourActivityId1 \
    --query 'ActivityType="YourActivityType"'
Error: cannot set query when activity ID is set
$ echo $?
1
```

The following can be repeated by replacing `cancel` with `terminate` or `delete`.
```
$ temporal activity cancel \
    --run-id <RunID from setup for YourActivityId1> \
    --query 'ActivityType="YourActivityType"'
Error: cannot set query when activity ID is set
$ echo $?
1
```

**Composition**


internal/temporalcli/commands.activity.go
```diff
@@ -525,21 +528,114 @@ func (c *TemporalActivityCountCommand) run(cctx *CommandContext, args []string)
...
-	if err := handle.Cancel(cctx, client.CancelActivityOptions{Reason: c.Reason}); err != nil {
-		return fmt.Errorf("failed to request activity cancellation: %w", err)
+	opts := ActivityReferenceOrBatchOptions{
+		ActivityId: c.ActivityId,
+		RunId:      c.RunId,
+		Query:      c.Query,
+		Rps:        c.Rps,
+	}
+
+	// TODO: should --yes be required if -o json or jsonl is used?
```
spkane31: I don't think `-o json(l)` should require a `--yes` flag

internal/temporalcli/commands.activity.go
```diff
@@ -550,24 +646,128 @@ func (c *TemporalActivityTerminateCommand) run(cctx *CommandContext, args []stri
...
+		Rps:        c.Rps,
+	}
+
+	activityOptions, batchReq, err := opts.activityExecOrBatch(cctx, c.Parent.Namespace, cl, c.Yes, singleOrBatchOverrides{})
+	if err != nil {
+		return err
+	}
+
+	// Reason for single terminate or batch request falls back to defaultReason
 	reason := c.Reason
```
spkane31: You can use `cmp.Or` here: `reason := cmp.Or(c.Reason, defaultReason())`. This can be used in a few places

internal/temporalcli/commands.activity.go
```diff
@@ -550,24 +646,128 @@ func (c *TemporalActivityTerminateCommand) run(cctx *CommandContext, args []stri
...
+}
+
+func (c *TemporalActivityDeleteCommand) run(cctx *CommandContext, args []string) error {
+	cl, err := dialClient(cctx, &c.Parent.ClientOptions)
+	if err != nil {
+		return err
+	}
+	defer cl.Close()
+
+	// TODO: do we need this warning, similar to workflow delete?
```
spkane31: Yes, we should emit the same warning

internal/temporalcli/commands.activity_test.go
```diff
@@ -1369,3 +1402,308 @@ func (s *SharedServerSuite) TestActivity_Terminate_DefaultReason_NoUnknownUser()
...
+			s.Eventually(func() bool {
+				res := s.Execute(
+					"activity", "list",
+					"--address", s.Address(),
+					"--query", fmt.Sprintf(`CustomKeywordField = "%s" AND ExecutionStatus = "Terminated"`, uniqueKW),
+				)
+				return res.Err == nil && strings.Count(res.Stdout.String(), operator+"-test-") >= 5
+			}, 5*time.Second, 200*time.Millisecond)
+
+			// //TODO: is it better to use list above or handle below?
```
spkane31: The list above is cleaner imo

internal/temporalcli/commands.activity_test.go
```diff
@@ -1369,3 +1402,308 @@ func (s *SharedServerSuite) TestActivity_Terminate_DefaultReason_NoUnknownUser()
...
+		return nil, activity.ErrResultPending
+	})
+
+	for _, operator := range []string{"cancel", "terminate", "delete"} {
+		uniqueKW := operator + "-test-fail-" + uuid.NewString()[:8]
+		activityId := operator + "-test-fail"
+		started := s.startActivity(activityId,
+			"--search-attribute", fmt.Sprintf(`CustomKeywordField="%s"`, uniqueKW),
+		)
+		runId := started["runId"].(string)
```
spkane31: I don't think you need to start an activity for these, these are client side failures.

internal/temporalcli/commands.activity.go
ks-temporal: Replaced.

internal/temporalcli/commands.activity_test.go
ks-temporal: Removed startActivity.

internal/temporalcli/commands.activity.go
```diff
@@ -525,21 +529,107 @@ func (c *TemporalActivityCountCommand) run(cctx *CommandContext, args []string)
...
 }
 
+func (s *ActivityReferenceOrBatchOptions) activityExecOrBatch(
+	cctx *CommandContext,
+	namespace string,
+	cl client.Client,
+	yesFlag bool,
+	overrides singleOrBatchOverrides,
+) (*client.GetActivityHandleOptions, *workflowservice.StartBatchOperationRequest, error) {
+	// If activity is set, we return activity handle options with activity ID and run ID
```
dandavison, line 576: I don't find it easy to understand the function name here: `activityExecOrBatch`

internal/temporalcli/commands.activity.go
```diff
@@ -550,24 +640,117 @@ func (c *TemporalActivityTerminateCommand) run(cctx *CommandContext, args []stri
...
+
+	activityOptions, batchReq, err := opts.activityExecOrBatch(cctx, c.Parent.Namespace, cl, c.Yes, singleOrBatchOverrides{
+		AllowYesWithActivityID: true,
 	})
-	// Terminate may fail if the activity doesn't exist or has already completed.
-	if err := handle.Terminate(cctx, client.TerminateActivityOptions{Reason: reason}); err != nil {
-		return fmt.Errorf("failed to terminate activity: %w", err)
+	if err != nil {
+		return err
+	}
```
dandavison, line 750: Would it make sense to do this validation before the `DecribeNamespace` RPC?

dandavison (approved):
Great, this all looks good to me.

internal/temporalcli/commands.activity.go
ks-temporal: I can change the name, if you have a better suggestion. I borrowed the name from existing `workflowExecOrBatch` that has similar function for batch workflow commands.

internal/temporalcli/commands.activity.go
ks-temporal: This is also similar to workflow delete command handler. I think one reason to keep it after describe namespace could be that if the describe namespace fails, the process still continues after a "global namespace warning". But if there is an error in execorbatch cmd parsing then it will exit. So if both errors are triggered on a cmd line, then moving the the validation before, will suppress the namespace warning.

internal/temporalcli/commands.activity.go
spkane31: I think we should keep it consistent of `workflowExecOrBatch` or `activityExecOrBatch`.

dandavison: OK yeah I guess I'd rename both of them. I don't generally approve of abbreviations in function names -- we don't refer to executions as "execs" normally. Both "exec" and "batch" are verbs, but only one is a noun. But I can do it whenever.

PR #1110, opened 2026-07-06
Add option to disable newsfeed in UI

tomwheeler:
## Related issues

Related to #[1104](https://github.com/temporalio/cli/pull/1104)

## What changed?

This adds a command-line option that configures the embedded UI server to disable the news feed feature.

## Manual tests

**Setup #1 (default case)**
```
temporal server start-dev
```

**Happy path #1 (default case)**
Open the Web UI and observe that the news fetch feature is enabled, as it is by default. This is visible by way of a megaphone icon in the top navigation. 

**Setup two (feature disabled)**
```
temporal server start-dev --ui-disable-news-fetch
```

**Happy path two (feature disabled)**
Open the Web UI and observe that the news fetch feature is disabled, since the environment variable is set to disable it, as described in the documentation. This is evident by the megaphone icon being omitted from the top navigation and the Web UI not making a request for the feed.

### DIFF
--- a/internal/devserver/server.go
+++ b/internal/devserver/server.go
@@ -76,6 +76,7 @@ type StartOptions struct {
 	UIPort                int    // Required if UIIP is non-empty
 	UIAssetPath           string
 	UICodecEndpoint       string
+	UIDisableNewsFetch    bool
 	PublicPath            string
 	DatabaseFile          string
 	MetricsPort           int
@@ -166,7 +167,6 @@ func (s *Server) SuppressWarnings() {
 }
 
 func (s *StartOptions) buildUIServer() *uiserver.Server {
-	disableNewsFetch, _ := strconv.ParseBool(os.Getenv("TEMPORAL_DISABLE_NEWS_FETCH"))
 	return uiserver.NewServer(uiserveroptions.WithConfigProvider(&uiconfig.Config{
 		Host:                MaybeEscapeIPv6(s.UIIP),
 		Port:                s.UIPort,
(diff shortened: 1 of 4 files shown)


internal/devserver/server.go
```diff
@@ -166,6 +166,7 @@ func (s *Server) SuppressWarnings() {
 }
 
 func (s *StartOptions) buildUIServer() *uiserver.Server {
+	disableNewsFetch, _ := strconv.ParseBool(os.Getenv("TEMPORAL_DISABLE_NEWS_FETCH"))
```
chaptersix: I don't think we should silently fail here. I think it makes sense to include in the start options (unless those are generated somehow)

chaptersix: just pushed up another way of doing this. It's more consistent with the existing configuration of the UI server in the CLI.

chaptersix:
Replaced the env-var read with a `--ui-disable-news-fetch` flag that flows through `StartOptions`, matching the other `--ui-*` options.

**Help output:**

```
$ temporal server start-dev --help
...
      --ui-disable-news-fetch            Disable the Web UI newsfeed.
                                         When set, the UI will not
                                         request the newsfeed and the
                                         button to open the newsfeed
                                         panel is hidden.
...
```

**Demo — start the dev server with it:**

```
$ temporal server start-dev --ui-disable-news-fetch
Temporal CLI 0.0.0-DEV (Server 1.32.0-158.0, UI 2.50.1)

Temporal Server:      localhost:7233
Temporal Persistence: in-memory
Temporal UI:          http://localhost:8233
Temporal Metrics:     http://localhost:60955/metrics
```

tomwheeler:
Apologies for the delay on my part. I made my previous commit just before boarding a very long flight. 

I built an executable from your updated code and verified that it works as expected. That is, when I run the `temporal server start-dev` command, the icon for the feed is visible in the top navigation and clicking it displays the news feed. If I disable it by running `temporal server start-dev --ui-disable-news-fetch` instead, the icon is hidden.

chaptersix (approved):
pending description and title update

PR #1112, opened 2026-07-07
ci/docs: escape MDX-incompatible patterns and add -subdir to gen-docs

chaptersix:
## Summary

This consolidates the excellent groundwork in #980, #1076, and #981 into a single change, and verifies the result against a real Docusaurus 3.10.1 build. Huge thanks to @lennessyy — the approach here is entirely built on those PRs; this just stitches them together and irons out a few interactions between them.

gen-docs now escapes the patterns that break Docusaurus MDX (JSX) compilation, on **every** path that writes a command description (including the new split paths):

- bare angle-bracket placeholders in prose (e.g. `<base64-encoded-cert>`, `<key>`) → `\<...\>`
- single-quoted JSON examples (e.g. `'{"a":"b"}'`, `'Key={"a":"b"}'`) → braces escaped in body text, backticked in option tables
- custom heading IDs (e.g. `## Heading {#id}`) → `{/* #id */}`, the form that compiles under Docusaurus 3.10 and still produces the custom anchor

Fenced code blocks and inline code spans are left untouched.

It also adds the `-subdir` flag (subcommands of the named command are written to a subdirectory, e.g. `-subdir cloud` → `cloud/*.mdx`, with deeper subcommands nested as headings), and a generic auto-generated notice.

## What changed relative to the existing PRs

All three PRs were on the right track. The differences here are about how they interact:

- **#1076 (MDX escaping):** kept as the core of this change, including the `{#id}` → `{/* #id */}` heading conversion — which I confirmed is exactly right for the 3.10 upgrade (see verification below). Escaping is now also applied to the `-subdir` split output, so cloud docs get the same treatment. Re-added unit tests for the escaping logic.
- **#980 (`-subdir`):** kept the split mechanism. Unified the `encodeJSONExample` regex so it matches both `'{...}'` and `'Key={...}'` (the standalone-vs-key-value cases the two PRs handled separately). Intentionally did **not** carry over #980's index-page generation: `command-reference/index.mdx` and `cloud/index.mdx` are hand-maintained on the docs site (custom ordering, the `ReleaseNoteHeader` component), so gen-docs deliberately does not emit them — generating them would overwrite that curated content. The companion docs PR restores those files after each regeneration.
- **#981 (generic notice):** folded in — the previous notice pointed at paths that no longer exist and don't hold for cloud-cli inputs.

## Companion PR
https://github.com/temporalio/documentation/pull/4836


internal/commandsgen/docs.go
```diff
@@ -92,45 +125,140 @@ func (w *docWriter) writeSubcommand(c *Command) {
...
+	buf.WriteString("---")
+	buf.WriteString("\n\n")
+	buf.WriteString(autoGeneratedNotice)
 
-		// alphabetize options
+	if w.isLeafCommand(c) {
+		buf.WriteString(fmt.Sprintf("This page provides a reference for the `temporal %s` command.\n\n", fullCmdName))
+		buf.WriteString(escapeMDXDescription(c.Description) + "\n\n")
+
+		var options []Option
```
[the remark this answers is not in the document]
chaptersix: Good catches — both fixed in 06cd622.

- Direct-child leaves (e.g. `cloud login`) now go through the same `writeLeafOptions` path as deeper leaves, so they render a Global Flags section. I also guarded that path so the `#global-flags` link is only emitted when inherited options actually exist — otherwise a leaf with none would link to an anchor that never gets generated (dangling anchor -> failed build under `onBrokenAnchors: throw`).
- You're also right about the description: a non-leaf split page like `cloud namespace` is analogous to a mid-tree group (`operator namespace` in the non-split output), which does render its description. It now writes the command description before the boilerplate. Verified with a Docusaurus 3.10.1 build.

internal/commandsgen/docs.go
```diff
@@ -92,45 +125,140 @@ func (w *docWriter) writeSubcommand(c *Command) {
...
+}
+
+func (w *docWriter) writeSplitSubcommand(c *Command, splitRoot string) {
+	fileName := splitFileName(c, splitRoot)
+	splitParts := strings.Split(splitRoot, " ")
+	cmdParts := strings.Split(c.FullName, " ")
+	relativeDepth := len(cmdParts) - len(splitParts)
+	prefix := strings.Repeat("#", relativeDepth)
+	leafName := strings.Join(cmdParts[len(splitParts)+1:], " ")
+
```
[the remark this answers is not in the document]
chaptersix: I looked at this closely and ended up keeping the relative path (`### ha update`, `### cert-ca create`) rather than the final segment. The reason: within one aggregated file the relative path gives each heading a unique, stable anchor, and other docs link to those anchors. Switching to the final segment breaks them — I confirmed it with a build: using `### update` made the site build fail with a broken anchor from `/cloud/high-availability/enable` -> `cloud/namespace#ha-update` (the heading had been `ha update`). Final-segment names would also collide within a file (multiple `### update` under `ha`, `cert-ca`, etc.), producing ambiguous auto-suffixed anchors. So this is an intentional deviation from the non-split output for the deep aggregated split files. Added a comment in the code explaining it.

internal/commandsgen/docs.go
```diff
@@ -45,13 +61,31 @@ type docWriter struct {
...
 	depth := c.depth()
+
+	// A split parent (e.g. "cloud") has no standalone file; its children are
+	// each written into a subdirectory instead.
+	if w.splitParents[c.FullName] {
+		return nil
+	}
+	// If any ancestor is a split parent, route this command into that subdirectory.
+	if depth >= 1 {
+		parts := strings.Split(c.FullName, " ")
```
ks-temporal, line 81: Suggestion: mention in comment that FullName never includes multiple continuous spaces.

internal/commandsgen/docs.go
```diff
@@ -92,45 +125,145 @@ func (w *docWriter) writeSubcommand(c *Command) {
...
+// splitFileName returns the file path for a command within a split parent.
+// For example, with splitRoot "cloud" and command "cloud namespace", it returns
+// "cloud/namespace".
+func splitFileName(c *Command, splitRoot string) string {
+	splitParts := strings.Split(splitRoot, " ")
+	cmdParts := strings.Split(c.FullName, " ")
+	if len(cmdParts) <= len(splitParts) {
+		return ""
+	}
+	return strings.Join(splitParts, "-") + "/" + cmdParts[len(splitParts)]
```
ks-temporal, line 204: Suggestion: mention in comment that subdirectory for say  splitRoot of "one two" is "one-two" instead of "one/two"

internal/commandsgen/docs.go
```diff
@@ -92,45 +125,145 @@ func (w *docWriter) writeSubcommand(c *Command) {
...
+}
+
+func (w *docWriter) writeSplitCommand(c *Command, splitRoot string) {
+	fileName := splitFileName(c, splitRoot)
+	splitParts := strings.Split(splitRoot, " ")
+	cmdParts := strings.Split(c.FullName, " ")
+	leafName := cmdParts[len(splitParts)]
+	fullCmdName := strings.Join(cmdParts, " ")
```
ks-temporal: Suggestion: c.FullName directly later in this func, instead of defining new fullCmdName.

internal/commandsgen/docs.go
```diff
@@ -254,11 +387,74 @@ func (w *docWriter) isLeafCommand(c *Command) bool {
...
+//     a JSX expression and fails on the '#'. These are converted to Docusaurus's
+//     MDX-compatible {/* #id */} comment form, which preserves the custom anchor.
+//
+// Content inside fenced code blocks (```) and inline code spans (backticks) is
+// left untouched, since MDX does not interpret those as JSX.
+func escapeMDXDescription(desc string) string {
+	lines := strings.Split(desc, "\n")
+	inCodeBlock := false
+	for i, line := range lines {
+		if strings.HasPrefix(strings.TrimSpace(line), "```") {
```
ks-temporal, line 440: Suggestion: I wonder if terminating triple-quote on last text line at the end of the same line should be handled too. Alternatively, mention in comment that triple quote must be by itself in a line to start or end a code block.

ks-temporal (approved):
Looks good. I left some suggestions, mostly to update code comment.

internal/commandsgen/docs.go
chaptersix: Done in 2f8862d — added a note that FullName segments are single-space separated.

internal/commandsgen/docs.go
chaptersix: Done — clarified that a multi-word split root is hyphenated into one directory ("one two" -> one-two/three).

internal/commandsgen/docs.go
chaptersix: Done — dropped fullCmdName and use c.FullName directly.

internal/commandsgen/docs.go
chaptersix: Went with documenting the assumption: added a note that fence markers are expected on their own line (the standard CommonMark form these descriptions use).

Issue #851, opened 2025-09-04
Temporal cli gives context deadline exceeded.  Tctl does not


esn89:
### What are you really trying to do?

Use `temporal` to connect to the frontend via gRPC.

### Describe the bug

Using `temporal` to talk to the frontend gives `context deadline exceeded`.  Using `tctl` is fine.

Here it is working with `tctl`:

![screenshot](https://i.imgur.com/lumYdqQ.png)

Here it is NOT working with `temporal`:

![screenshot](https://i.imgur.com/e0YWa93.png)

It is going to the SAME backend FQDN.  I am showing `env` to illustrate that I do not have any `TEMPORAL_` env vars set.

### Minimal Reproduction

Install temporal like how I did on Kubernetes of 1.33.3-gke.1136000

`helm install temporal temporalio/temporal --version 0.65.0`

Install istio for ingress.  DO NOT add the sidecar injection into the namespace.  We do not want the mesh feature.  Ingress only.

`./istioctl install --set profile=minimal -y`

Set up ingress gateway:

`helm upgrade --install istio-ingressgateway --namespace istio-system istio-official/gateway --version 1.27.0 --values ingress-gateway.yaml`

My `ingress-gateway.yaml` looks like this:

```yaml
name: istio-ingressgateway
service:
  loadBalancerIP: $MY_INTERNAL_IP
  annotations:
    cloud.google.com/load-balancer-type: Internal
    networking.gke.io/internal-load-balancer-subnet: internal1
  ports:
  - name: status-port
    port: 15021
    protocol: TCP
    targetPort: 15021
  - name: http2
    port: 80
    protocol: TCP
    targetPort: 80
  - name: https
    port: 443
    protocol: TCP
    targetPort: 443
  - name: temporal
    port: 7233
    protocol: TCP
    targetPort: 7233
autoscaling:
  maxReplicas: 10
  targetCPUUtilizationPercentage: 75
resources:
  requests:
    cpu: 250m
    memory: 512Mi
  limits:
    memory: 1Gi
    cpu: "1"
```

Once it is up (you will need to have your TLS certs) for Gateway and Virtualservice:

gateways:

```yaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  name: web-temporal-gateway
  namespace: istio-system
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - web-temporal.mycompany.com
    port:
      name: temporal-mutual
      number: 443
[...]

Issue #224, opened 2023-05-03
[Feature Request] Show actual error message when mTLS certs are missing


taonic:
### Is your feature request related to a problem? Please describe.
Right now, if mTLS certs (`--tls-cert-path` and `--tls-cert-path`) are left out by mistake, the cli returns an empty error.

```
➜  workflow reset --address='<namespace>.tmprl.cloud:7233' --namespace='<namespace>' --workflow-id "fbfcc301-e801-48c0-8bba-4d8b4afcfe7d" --reason test --event-id 3 --run-id "ac56c805-87fa-486d-af43-eb9fb54518a6"
Error: reset failed:
('export TEMPORAL_CLI_SHOW_STACKS=1' to see stack traces)
```

### Describe the solution you'd like
It would be more helpful if it tells you want went wrong, e.g. server requires mTLS so please provide `--tls-cert-path` and `--tls-cert-path`.

### Additional context
Similar issue was previously reported in `tctl` in https://github.com/temporalio/tctl/issues/228 and https://github.com/temporalio/tctl/issues/353. At least one Temporal customer has submitted a support ticket with this issue.

Quinn-With-Two-Ns:
On the CLI rewrite branch the connection failure is logged, but not returned as the output. We need to decide if logging in the CLI is appropriate 

```
time=2024-03-02T15:37:55.037 level=ERROR msg="failed reaching server: last connection error: connection error: desc = \"error reading server preface: read tcp 192.168.0.20:56053->54.191.166.44:7233: read: connection reset by peer\""
Exit code: 1                                    
```

cretz:
TLS doesn't always provide details on why handshake failed. We delegate to the Go TLS library, and are subject to its error messages. We may be able to guess that TLS expectation exists but not was provided by checking error message though.

Quinn-With-Two-Ns:
To be clear my question was not with what is logged, but the fact that we log at all and if we want that in a CLI

cretz:
I think we do want to log connection failure (which is the output, it just happens to be in log format and on stderr, so it can be seen regardless of stdout output format). Whether we want to more clearly clarify that it is due to TLS failure may be worth confirming though.

PR #1114, opened 2026-07-08
Add staged connection diagnosis for opaque dial failures

rossnelson:
## Related issues

Related to #224 and #851.

## What changed?

**This PR adds bounded, evidence-based diagnosis after a failed connection, turning opaque dial errors into supported next steps when the evidence allows.**

Before this change, failed connections could end as opaque dial errors. After the real failure, the CLI runs bounded DNS, TCP, and TLS checks in sequence. It gives advice only when the evidence supports it. Diagnosis stops after three seconds or cancellation. `TEMPORAL_CLI_DISABLE_CONNECT_DIAGNOSIS` disables it. A generic TLS handshake failure does not prove mTLS, so the CLI gives no certificate advice.

This PR adds one connection-specific check to the existing `CommandOptions.Fail` path. It leaves all other commands, generated command flow, Activity errors, extensions, stdout, status, and usage behavior unchanged.

connection-error-diagnosis

## Checklist

**Stability**
- [ ] Breaking changes are marked with 💥 in the PR title and release notes
- [ ] Changes to JSON output (`-o json` / `-o jsonl`) are treated as breaking changes

**Design**
- [ ] New commands follow `temporal <noun> <verb>` structure (e.g. `temporal workflow start`)
- [ ] New flags are named after the API concept, not the implementation mechanism (good: `--search-attribute`, bad: `--index-field`)
- [ ] New flags don't duplicate an existing flag that serves the same purpose
- [ ] New flags do not have short aliases without strong justification
- [ ] Experimental features are marked with `(Experimental)` in `commands.yaml`

- [ ] All flags shown in help text and examples are implemented and functional
- [ ] Summaries use sentence case and have no trailing period
- [ ] Long descriptions end with a period and include at least one example invocation
- [ ] Examples use long flags (`--namespace`, not `-n`), one flag per line
- [ ] Placeholder values use `YourXxx` form (`YourWorkflowId`, `YourNamespace`)

**Behavior**

**Tests**
- [x] Added unit test(s) (`func TestXxx`) where applicable

## Manual tests

**Setup**

No manual setup was used.

**Happy path**

No manual happy-path run was used. After merge, the full suite passed in `internal/temporalcli` in 166.201s:

```text
go test ./... -count=1 -timeout=10m
```

**Error case**

Connection-focused `go test -race` tests passed. The concurrent development-server race still reports the existing upstream global color race in unchanged `commands.go` lines. This PR does not claim that the full race suite passes.

**Composition**

No manual composition test was used.


chaptersix:
@codex review

chaptersix:
I think the codex review is worth taking a look at but everything else lgtm. IMO we are free to change the error output at will.

chaptersix (approved):

PR #1115, opened 2026-07-08
Bump go.temporal.io/server to v1.31.2

lina-temporal:
- Bumps server version for an OSS security release
- Dockerfile's Alpine is already up-to-date


chaptersix (approved):
assuming tests pass LGTM

PR #1122, opened 2026-07-11
Add opt-in periodic CLI update notices

### LINKED ISSUE #235: [Feature Request] Check and inform user when newer version is available
I don't always remember to check this repo to see if a newer version is available, and so just recently spent a bunch of time investigating an issue that turns out to have been solved since I last upgraded. In retrospect, "check before you test" may be somewhat reasonable while this project is <1.0, but as the time between versions inevitably increases, users will more frequently end up in the same "oh shoot, I should have checked for upgrades first" camp as I did.

A message or similar output on start or on `--version` that informs of new releases of the CLI. Added bonus: a built-in self-upgrade command. Alternate: like the pip example below, include the command to upgrade in the output for easy copy-pasta.

### Additional context
The message I'm envisioning is similar to what `pip` gives:

```text
[notice] A new release of pip is available: 23.0.1 -> 23.1.2
[notice] To update, run: pip install --upgrade pip
```

chaptersix:
Closes #235.

## What changed

- add a profile-independent `cli.update_check.enabled` config property for opting in
- query `temporal.download` for the latest stable CLI release on a jittered 60–84 hour cadence
- cache only the latest version and scheduling metadata in the existing Temporal TOML config
- emit newer-version notices to stderr with 1, 3, 7, then 14-day reminder backoff
- include the update check in all CLI invocations, including Cobra's `--version` path
- preserve CLI-owned top-level config when connection profiles are rewritten
- suppress update-check network and config failures so normal CLI behavior is unaffected

## User impact

Users can opt in with:

```sh
temporal config set --prop cli.update_check.enabled --value true
```

Once enabled, the CLI periodically checks for a newer release and prints a throttled notice to stderr without changing structured stdout.

## Validation

- `go test ./internal/temporalcli -count=1`
- `go test ./... -count=1`
- `git diff --check`


chaptersix:
eh

PR #1123, opened 2026-07-11
chore: exclude UI Server from Dependabot

chaptersix:
## What changed

- Exclude `github.com/temporalio/ui-server/v2` from Dependabot version updates.

## Why

UI Server is a Temporal-owned dependency whose upgrades can also move coordinated `go.temporal.io` dependencies. The existing ignore rule only matches modules under `go.temporal.io/*`, so it does not match UI Server's GitHub module path.

This keeps UI Server upgrades in the same manual, coordinated update process as the other Temporal dependencies.

## Validation

- `check-jsonschema --builtin-schema vendor.dependabot .github/dependabot.yml`
- `git diff --check`


harani-mukkala (approved):
LTGM

cc @rossnelson Could you verify this as well.

PR #1137, opened 2026-07-23
Delegate help and completion to extensions when applicable

jeri-temporal:
Shell completion always sets __complete as the first argument, so to
delegate to extensions, `temporal __complete cloud n` needs to be
rewritten as `temporal-cloud __complete n`. `help` can be invoked the
same way

## Related issues

CLDDX-150

## What changed?

* `tryExecuteExtension` delegates `help` and shell completion to extensions that match the remaining arguments
* extensions are registered during shell completion, so "temporal <TAB>" will show them

## Checklist

**Design**
* [x] This feature does not depend on Cloud-only APIs or behavior (it works against an OSS server)

**Tests**
* [x] Added unit test(s) (`func TestXxx`) where applicable

## Manual tests

**Setup**
Install at least one extension, like the [cloud cli](https://github.com/temporalio/cloud-cli#quick-install) (`brew install temporalio/prerelease/temporal-cloud` on Mac or Linux with homebrew installed)

**Happy path**

```
$ temporal <TAB>
activity    -- Operate on Activity Executions
batch       -- Manage running batch jobs
cloud       -- An extension command located at /opt/homebrew/bin/temporal-cloud
completion  -- Generate the autocompletion script for the specified shell
config      -- Manage config files (EXPERIMENTAL)
env         -- Manage environments
help        -- Help about any command
nexus       -- Start, list, and operate on Nexus Operations
operator    -- Manage Temporal deployments
schedule    -- Perform operations on Schedules
server      -- Run Temporal Server
task-queue  -- Manage Task Queues
worker      -- Read or update Worker state
workflow    -- Start, list, and operate on Workflows

$ temporal cl<TAB>
temporal cloud

$ temporal cloud <TAB>
account          -- Manage Temporal Cloud account
apikey           -- Manage Temporal Cloud API keys
async-operation  -- Manage async operations
connectivity     -- Manage Temporal Cloud connectivity rules
custom-role      -- [Experimental] Manage Temporal Cloud custom roles
help             -- Help about any command
login            -- Authenticate with Temporal Cloud
logout           -- Clear Temporal Cloud authentication credentials
namespace        -- Manage Temporal Cloud namespaces
nexus            -- Manage Temporal Cloud Nexus Operations
region           -- Manage Temporal Cloud regions
service-account  -- Manage Temporal Cloud service accounts
user             -- Manage Temporal Cloud users
user-group       -- Manage Temporal Cloud user groups
whoami           -- Display the current authenticated identity
```

```
$ temporal help --all
...
Available Commands:
  activity    Operate on Activity Executions
  batch       Manage running batch jobs
  cloud       An extension command located at /opt/homebrew/bin/temporal-cloud
  completion  Generate the autocompletion script for the specified shell
  config      Manage config files (EXPERIMENTAL)
  env         Manage environments
  help        Help about any command
  nexus       Start, list, and operate on Nexus Operations
  operator    Manage Temporal deployments
  schedule    Perform operations on Schedules
  server      Run Temporal Server
  task-queue  Manage Task Queues
  worker      Read or update Worker state
  workflow    Start, list, and operate on Workflows
...

temporal help cloud
The Temporal Cloud CLI provides commands for managing and operating Temporal Cloud resources,
including namespaces, users, and account settings.

Example:

temporal cloud namespace get --namespace my-namespace.my-account

Usage:
  temporal cloud [command]

Available Commands:
  account         Manage Temporal Cloud account
  apikey          Manage Temporal Cloud API keys
  async-operation Manage async operations
  connectivity    Manage Temporal Cloud connectivity rules
  custom-role     [Experimental] Manage Temporal Cloud custom roles
  help            Help about any command
  login           Authenticate with Temporal Cloud
  logout          Clear Temporal Cloud authentication credentials
  namespace       Manage Temporal Cloud namespaces
  nexus           Manage Temporal Cloud Nexus Operations
  region          Manage Temporal Cloud regions
  service-account Manage Temporal Cloud service accounts
  user            Manage Temporal Cloud users
  user-group      Manage Temporal Cloud user groups
  whoami          Display the current authenticated identity
```

**Error case**

```
temporal z<TAB>
```

(no completions shown)

Behavior remains unchanged when an unknown argument is passed to `temporal help`:

```
$ temporal help zzz
The Temporal CLI manages, monitors, and debugs Temporal apps. It lets you run
a local Temporal Service, start Workflow Executions, pass messages to running
Workflows, inspect state, and more.
...
```

### DIFF
--- a/internal/temporalcli/commands.extension.go
+++ b/internal/temporalcli/commands.extension.go
@@ -36,8 +36,20 @@ func (err ExtensionNonZeroExit) Unwrap() error {
 // tryExecuteExtension tries to execute an extension command if the command is not a built-in command.
 // It returns an error if the extension command fails, and a boolean indicating whether an extension was executed.
 func tryExecuteExtension(cctx *CommandContext, tcmd *TemporalCommand) (error, bool) {
+	// Special commands like "help" and "__complete" should be set aside and delegated to an extension command that matches
+	// the rest of the given arguments. "temporal help my-extension" should be rewritten to "temporal-my-extension help"
+	// Some of these commands, like "__complete" used for shell completion, don't actually get registered by Cobra until
+	// just before they're invoked, so we need to split out the delegatable commands before trying to Find() a matching
+	// subcommand.
+	delegatableCommands, nonDelegatedArgs := splitDelegatedCommands(cctx.Options.Args)
+
+	// If a completion command (used for generating the script that invokes "__complete") already exists or has been
+	// explicitly disabled, this will do nothing, but if neither of those cases are true, we want to make sure
+	// this command is registered so extensions can't shadow it.
+	tcmd.Command.InitDefaultCompletionCmd()
+
 	// Find the deepest matching built-in command and remaining args.
-	foundCmd, remainingArgs, findErr := tcmd.Command.Find(cctx.Options.Args)
+	foundCmd, remainingArgs, findErr := tcmd.Command.Find(nonDelegatedArgs)
 
 	// Cobra normally adds --help/-h before parsing, but extension dispatch
 	// pre-parses flags before Cobra's execution path runs. We Initialize it so that
@@ -58,6 +70,7 @@ func tryExecuteExtension(cctx *CommandContext, tcmd *TemporalCommand) (error, bo
 
 	// Search for an extension executable.
 	cmdPrefix := strings.Fields(foundCmd.CommandPath())
+
 	extPath, extArgs := lookupExtension(cmdPrefix, extArgs)
 
 	// Parse CLI args that need validation.
@@ -71,6 +84,16 @@ func tryExecuteExtension(cctx *CommandContext, tcmd *TemporalCommand) (error, bo
 		return nil, false
 	}
 
+	if len(delegatableCommands) > 0 && isCompletionCommand(delegatableCommands[0]) && len(extArgs) == 0 {
+		// __complete always expects at least one argument, the last of which is the current subcommand
+		// or argument to expand, with an empty string matching all possibilities.
+		// ["temporal", "__complete", "activity"] means this cli should return any subcommands and extentions that
+		// match "activity", whereas ["temporal", "__complete", "activity", ""] means we should show what's available
+		// on the activity subcommand. The same logic applies to extension commands, so even if we matched an extension,
+		// if there are no further args, it's still this cli's responsibility to respond to the completion request.
+		return nil, false
+	}
+
 	// Apply --command-timeout if set.
 	ctx := cctx.Context
 	if timeout := tcmd.CommandTimeout.Duration(); timeout > 0 {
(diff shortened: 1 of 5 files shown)


internal/temporalcli/commands.help.go
```diff
@@ -53,36 +56,64 @@ func customizeHelpCommand(rootCmd *cobra.Command) {
 }
 
 // registerExtensionCommands adds discovered extensions as placeholder commands
-// so they appear in the default help output. It filters extensions based on
-// the current command's path in the hierarchy.
+// so they appear in shell completion and the default help output. It filters extensions
+// based on the current command's path in the hierarchy.
 func registerExtensionCommands(cmd *cobra.Command) {
```
jeri-temporal, line 61: Now that this is used for shell completion as well, it feels like it probably belongs in `commands.extension.go`, but I didn't want to move it at the same time as making modifications to its behavior. If you agree, I'm happy to move it

internal/temporalcli/commands.extension.go
```diff
@@ -94,6 +119,36 @@ func tryExecuteExtension(cctx *CommandContext, tcmd *TemporalCommand) (error, bo
 	return nil, true
 }
 
+// splitDelegatedCommands separates out commands that should be delegated to an extension
+// from the rest of the args given. These commands are inherently position-dependent, so they're
+// only treated specially when they're at the start of the list of arguments.
```
ks-temporal, line 124: Suggestion: Is it worth mentioning in the comment that the returned slices should be treated as read-only, and some operations such as append must not be done on them?

internal/temporalcli/commands.extension.go
```diff
@@ -229,13 +283,12 @@ func discoverExtensions() [][]string {
 			}
 
 			path := extensionBinaryToCommandPath(baseName)
-			key := strings.Join(path, "/")
-			if seen[key] {
+			key := strings.Join(path, " ")
```
ks-temporal: Suggestion: since filename can contain " " but not "/", could this cause ambiguity if baseName itself has space in some path fragment? e.g., baseName of "temporal-my command" vs "temporal-my-command" both will map to the same key I think. I don't think we support cli command component with space like temporal "my command" anyway, but to avoid any confusion, I suggest to keep the old delimiter of "/" if possible. One option could be to return an array of items (containing cmd path with space, and exe path) instead of map (from cmd path with space to exe path), and use the previous "seen" logic. If you decide to change this, some code in the caller func (registerExtensionCommands) will need to be updated too. Thanks.

internal/temporalcli/commands.help.go
```diff
@@ -53,36 +56,64 @@ func customizeHelpCommand(rootCmd *cobra.Command) {
...
+				parent = found
+				continue
+			}
+
+			var short string
+
+			if i == len(extPath)-1 {
+				short = fmt.Sprintf("An extension command located at %s", executablePath)
+			} else {
+				short = fmt.Sprintf("Extension commands under %s", strings.Join(ext[:len(cmdPath)+i+1], " "))
```
ks-temporal, line 103: Suggestion: is it possible to include this code path in a test?

ks-temporal (approved):
Approving with a few minor comments, inline in code.

internal/temporalcli/commands.extension.go
jeri-temporal: Can do!

internal/temporalcli/commands.extension.go
jeri-temporal: If we ever choose to support executables with spaces in their name, I think they have the potential to cause conflicts no mater whether we treat `temporal-my command` identically to `temporal-my_command` or `temporal-my-command`, but that's where PATH ordering comes in, IMO.

Switching from `/` to ` ` was part of an earlier approach to how command descriptions were built though, so I went ahead and switched it back.

internal/temporalcli/commands.help.go
jeri-temporal: Done!

PR #1138, opened 2026-07-27
feat(commandsgen): generate CLI environment variables docs index

### LINKED ISSUE #697: Create an index of environment variables for the docs site

sornapudisuresh:
## Summary
- Extends `gen-docs` to emit `environment-variables.mdx` from `implied-env` options (alphabetized env var → flag → description).
- Adds a tip distinguishing shell env vars from `temporal env` presets.
- Documents special cases: `TEMPORAL_GRPC_META_*` and legacy TLS variables.
- Skips the page when `-subdir` is used so cloud-cli docs generation does not overwrite it.
- Closes #697

## Test plan
- [x] `go test ./internal/commandsgen/ ./cmd/gen-docs/`
- [x] `go test ./...`
- [x] `go run ./cmd/gen-docs -input internal/temporalcli/commands.yaml -input cliext/option-sets.yaml -output dist/docs` and inspect `dist/docs/environment-variables.mdx`


chaptersix:
Thank you for your PR.

see linked documentation repo PR.

PR #1143, opened 2026-07-28
Fix inverted `--yes` description on `operator namespace delete`

### LINKED ISSUE #1142: [Bug] `--yes` on `operator namespace delete` is documented as the inverse of what it does
`temporal operator namespace delete` documents `-y, --yes` as **"Request confirmation before deletion."** It does the opposite — it supplies the confirmation so the delete runs unattended.

### Repro (CLI 1.7.2, Server 1.31.1)

```bash
temporal server start-dev --port 17233 --headless &
export TEMPORAL_ADDRESS=localhost:17233
temporal operator namespace create --namespace deltest

temporal operator namespace delete --namespace deltest --yes < /dev/null
# Are you sure you want to delete namespace deltest? Type namespace name to confirm: deltest
# Namespace deltest has been deleted.        <- nothing was confirmed
```

Without `--yes` the same command exits 1 with `Error: user denied confirmation or mistyped the namespace name`, and the namespace survives.

### Source

`internal/temporalcli/commands.yaml`, under `temporal operator namespace delete`:

```yaml
      - name: yes
        short: y
        type: bool
        description: Request confirmation before deletion.
```

This is also what reaches the generated `--help` and https://docs.temporal.io/cli/operator.

### Suggested fix

`description: Don't prompt to confirm deletion.` — matching `search-attribute remove` ("Don't prompt to confirm removal.") and `workflow terminate` ("Don't prompt to confirm termination.") in the same file, both of which are correct. Looks like an isolated slip rather than an intentional difference.

### Why it seems worth fixing

An inverted description is worse than a missing one. The prompt on this command is a type-the-namespace-name check — the strongest confirmation in the CLI — and the description says the flag *adds* confirmation. Anyone reaching for `--yes` to make a namespace delete safer, in a runbook or in tooling that reads `--help`, gets an unattended delete of every Workflow Execution and Task Queue in that namespace.

Happy to send a PR if useful; I'd need a pointer to the codegen step to include the regenerated files.

starfleeth:
## Related issues

Closes #1142

## What changed?

`temporal operator namespace delete` documented `-y, --yes` as **"Request confirmation before deletion."** It does the opposite — it supplies the confirmation so the delete runs unattended. The behavior is correct and unchanged; only the description was wrong.

New wording matches the other 16 `--yes` flags in `commands.yaml`, all of which use "Don't prompt to confirm …". This was the only one that deviated. Repro and detail in #1142.

No test added: the description is display-only metadata with no execution path, which is also why this survived — it can't fail a test, a build, or a review of the Go.

⚠️ **`commands.gen.go` was hand-applied, not regenerated** (no Go toolchain on this machine). The string appears exactly once as a plain literal, so it should be byte-identical to generator output — the `Regen code, confirm unchanged` CI step will confirm.

### DIFF
--- a/internal/temporalcli/commands.gen.go
+++ b/internal/temporalcli/commands.gen.go
@@ -2010,7 +2010,7 @@ func NewTemporalOperatorNamespaceDeleteCommand(cctx *CommandContext, parent *Tem
 		s.Command.Long = "Removes a Namespace from the Service.\n\n```\ntemporal operator namespace delete [options]\n```\n\nFor example:\n\n```\ntemporal operator namespace delete \\\n    --namespace YourNamespaceName\n```"
 	}
 	s.Command.Args = cobra.MaximumNArgs(1)
-	s.Command.Flags().BoolVarP(&s.Yes, "yes", "y", false, "Request confirmation before deletion.")
+	s.Command.Flags().BoolVarP(&s.Yes, "yes", "y", false, "Don't prompt to confirm deletion.")
 	s.Command.Run = func(c *cobra.Command, args []string) {
 		if err := s.run(cctx, args); err != nil {
 			cctx.Options.Fail(err)
--- a/internal/temporalcli/commands.yaml
+++ b/internal/temporalcli/commands.yaml
@@ -2355,7 +2355,7 @@ commands:
       - name: yes
         short: y
         type: bool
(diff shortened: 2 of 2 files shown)


dandavison (approved):

dandavison:
Thanks!

PR #1150, opened 2026-08-03
Reject update-options --start-delay for workflow Activities

dandavison:
## What changed?
Reject update-options --start-delay for workflow Activities

**Tests**

### DIFF
--- a/internal/temporalcli/commands.activity_test.go
+++ b/internal/temporalcli/commands.activity_test.go
@@ -440,12 +440,10 @@ func (s *SharedServerSuite) TestActivityStandalone_UpdateOptions_StartDelay() {
 }
 
 func (s *SharedServerSuite) TestActivityUpdateOptions_StartDelayRejectedForWorkflowActivity() {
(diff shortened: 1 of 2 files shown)


internal/temporalcli/commands.activity_test.go
```diff
@@ -439,6 +439,30 @@ func (s *SharedServerSuite) TestActivityStandalone_UpdateOptions_StartDelay() {
 	s.Eventually(started.Load, 2*time.Second, 100*time.Millisecond)
 }
 
+func (s *SharedServerSuite) TestActivityUpdateOptions_StartDelayRejectedForWorkflowActivity() {
+	run := s.waitActivityStarted()
```
spkane31: nit: these are client side validations so you don't even need to start an activity.

spkane31 (approved):

internal/temporalcli/commands.activity_test.go
dandavison: Thanks, good point, done.

PR #1157, opened 2026-08-06
Improve the core Schedule CLI experience

rossnelson:
## Summary

- Add focused `temporal schedule patch` operations for notes, policy, state, cadence, timing, and workflow-action fields.
- Preserve omitted fields and add bounded conflict retries with truthful submission and output-error handling.
- Correct Schedule create priority handling and test Schedule update replacement behavior.

https://github.com/user-attachments/assets/3466abf8-b53b-40c1-a1c2-2bca4359ee81

## Test plan

- [x] `go test ./internal/temporalcli -count=1`
- [x] `go test ./... -count=1`
- [x] `make gen` produces byte-identical output
- [x] `make build`


internal/temporalcli/commands.schedule.go
```diff
@@ -253,6 +260,22 @@ func (c *ScheduleConfigurationOptions) toScheduleSpec(spec *client.ScheduleSpec)
...
+		return nil, fmt.Errorf("priority key must be between 0 and %d", math.MaxInt32)
+	}
+	if len(sw.FairnessKey) > 64 {
+		return nil, fmt.Errorf("fairness key must be at most 64 bytes")
+	}
+	if math.IsNaN(float64(sw.FairnessWeight)) ||
+		sw.FairnessWeight < 0 ||
+		(sw.FairnessWeight > 0 && sw.FairnessWeight < 0.001) ||
+		sw.FairnessWeight > 1000 {
+		return nil, fmt.Errorf("fairness weight must be between 0.001 and 1000")
```
dnr: please don't enforce stuff like this on the client side. these validations can depend on server config and should be validated on the server, so that if they change we don't have to go change all the clients

internal/temporalcli/commands.schedule.go
```diff
@@ -513,6 +548,498 @@ func (c *TemporalScheduleTriggerCommand) run(cctx *CommandContext, args []string
...
+		}
+		calendarCronString, err := toCronString(&calendarSpec)
+		if err != nil {
+			return nil, nil, err
+		}
+		calendarCron = append(calendarCron, calendarCronString)
+	}
+	for _, cronString := range cron {
+		trimmedCronString := strings.TrimSpace(cronString)
+		if strings.HasPrefix(trimmedCronString, "TZ=") || strings.HasPrefix(trimmedCronString, "CRON_TZ=") {
```
dnr: it is supported on the server, why not here?

internal/temporalcli/commands.yaml
```diff
@@ -3142,13 +3142,161 @@ commands:
...
+
+      A successful command confirms that the patch was submitted, not that it
+      has been applied on every backend.
+
+      Omitting cadence options preserves the existing cadence. Supplying any
+      `--calendar`, `--cron`, or `--interval` value replaces all existing
+      calendar, cron, and interval sources. For example, use
+      `--cron '0 12 * * *'` to replace cadence, or `--cadence-clear-all` to
+      clear it. Cadence clear requires the resulting Schedule to be paused;
+      pause explicitly with `--paused=true` when needed.
```
dnr: I kinda like this term "cadence", but... afaik it doesn't exist in any of the existing docs or cli or sdks. The term for "when to run the actions" is "spec", short for specification. Introducing a brand new term seems like it's going to add confusion. (I'm not claiming that's better, just that the consistency is important)

PR #1176, opened 2026-08-25
Fix cliext build, add it to CI workflow

jeri-temporal:
## Related issues

None directly, it came up while I was looking at adding structure to some of the auth failures in the client

I couldn't find anywhere that the cliext tests were getting executed, and attempting to build `cliext` by itself fails with 

```
./client.go:137:11: profile.Authority undefined (type "go.temporal.io/sdk/contrib/envconfig".ClientConfigProfile has no field or method Authority)
./config.oauth.go:144:20: assignment mismatch: 1 variable but envconfig.DefaultConfigFilePath returns 2 values
```

## What changed

* Added cliext tests to CI workflow (causes the above failures to surface in the CI workflow)
* Updated `cliext/go.mod` to use the same versions (as applicable) of packages as are used in the main `go.mod` (fixes the failures)
* Added `test` as a Makefile target (convenience when making changes across `cliext` and the main package)


chaptersix (approved):
Thanks!

PR #1196, opened 2026-09-08
Fix standalone activity stability labels and polish help docs

fretz12:
## What changed?

  - Removed the (Experimental) label from stable Standalone Activity commands.
  - Marked activity pause, unpause, reset, and update-options as experimental.
  - Updated Activity and Batch help text to accurately document Standalone Activity operations.
  - Add guard to disallow both activity ID and query as params
  - Fixed related documentation wording and regenerated commands.gen.go.


internal/temporalcli/commands.yaml
```diff
@@ -5476,9 +5476,9 @@ option-sets:
         type: string
         short: q
         description: |
-          Content for an SQL-like `QUERY` List Filter.
+          EXPERIMENTAL: Starts a batch Activity operation using an SQL-like `QUERY` List Filter.
+          This behavior may change in the future.
           You must set either --workflow-id or --query.
```
ks-temporal: Copilot already mentioned this earlier, but it will be good to update this to indicate that exactly one of workflow-id, activity-id or query must be set, since this is now used by WFA and SAA both.

fretz12: Added the doc update. Also added a guard where for activities you can't set activityID and query both, since it was actually allowing it.

internal/temporalcli/commands.gen.go
```diff
@@ -184,8 +184,8 @@ func (v *SingleActivityOrBatchOptions) BuildFlags(f *pflag.FlagSet) {
 	f.StringVarP(&v.Query, "query", "q", "", "Content for an SQL-like `QUERY` List Filter. You must set either --workflow-id or --query. Note: Using --query for batch activity operations is an experimental feature and may change in the future.")
 	f.StringVarP(&v.RunId, "run-id", "r", "", "Run ID. Only use with --workflow-id or --activity-id. Cannot use with --query.")
 	f.StringVar(&v.Reason, "reason", "", "Reason for batch operation. Only use with --query. Defaults to user name.")
-	f.BoolVarP(&v.Yes, "yes", "y", false, "Don't prompt to confirm signaling. Only allowed when --query is present.")
-	f.Float32Var(&v.Rps, "rps", 0, "Limit batch's requests per second. Only allowed if query is present.")
+	f.BoolVarP(&v.Yes, "yes", "y", false, "Don't prompt to confirm the batch operation. Only allowed when --query is present.")
```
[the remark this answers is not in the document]
fretz12: done

internal/temporalcli/commands.yaml
```diff
@@ -5455,16 +5469,17 @@ option-sets:
...
         type: string
         short: q
         description: |
-          Content for an SQL-like `QUERY` List Filter.
-          You must set either --workflow-id or --query.
-          Note: Using --query for batch activity operations is an experimental feature and may change in the future.
+          EXPERIMENTAL: Starts a batch Activity operation using an SQL-like `QUERY` List Filter.
+          This behavior may change in the future.
+          Targets a batch of Activities; cannot be combined with --workflow-id,
+          --run-id, or --activity-id.
```
ks-temporal: The previous text implied that exactly one of the options must be supplied (i.e., it is an error to not supply any). The new text implies that it may be okay to not set any (from the set of workflow-id, activity-id, query), and there is no other text which implies that exactly one of these must be set.  Basically we need to cover both information: (a) that query cannot be combined with another of workflow-id, run-id, activity-id, and (b) exactly one of query, workflow-id or activity-id must be set. Not sure how important (b) is. What do you think?

ks-temporal (approved):

internal/temporalcli/commands.yaml
fretz12: IMO this command does too much :)
But yes, I think we're already in agreement with A; good point on B, but note that workflow-id and run-id can only be set with activity-id. I've made that clarification

Issue #1193, opened 2026-09-04
[Feature Request] Decode workflow histories concurrently with remote codec servers


chriskuehl:
### Is your feature request related to a problem? Please describe.

When the CLI is using Temporal Cloud and configured with a remote codec server, exporting a large workflow history can be very slow:

```console
temporal workflow show \
  --workflow-id <workflow-id> \
  --output json > history.json
```

The codec protocol supports sending multiple payloads in one `/decode` request. However, the CLI's payload visitor makes a separate request for each payload or `Payloads` field it encounters and processes those visits serially.

Because workflows contain so many payloads, even a small amount of latency (~50ms) causes some exports to take 5+ minutes for us.

Increasing the RPC timeout prevents the operation from timing out, but does not address the latency.

### Describe the solution you'd like

Remote codec calls made while processing a workflow history should be concurrent, batched more broadly, or otherwise avoid paying one network round trip serially for every visited field.

`api-go` now supports opt-in concurrent payload visiting through `VisitPayloadsOptions.ConcurrencyLimit`, which may provide a path to doing this with bounded concurrency.

### Additional context

This was previously recognized in:

- https://github.com/temporalio/cli/issues/648

That issue explicitly mentioned concurrently calling the codec instead of using the serial visitor. It was closed by https://github.com/temporalio/cli/pull/676, which added a configurable RPC timeout but did not address the underlying latency.

There is also an open cross-SDK feature tracking concurrent payload visiting:

- https://github.com/temporalio/features/issues/772

The Temporal CLI item remains unchecked. The underlying Go support has since landed:

- https://github.com/temporalio/api-go/pull/244
- https://github.com/temporalio/sdk-go/issues/2223

The CLI currently constructs `NewPayloadCodecGRPCClientInterceptor` here:

https://github.com/temporalio/cli/blob/main/cliext/client.go

However, that interceptor is currently used without configuring the payload visitor's concurrency option.

PR #1206, opened 2026-09-13
Avoid duplicate codec calls for System Nexus history

chaptersix:
## Related issues

Related to https://github.com/temporalio/cli/issues/1193.

Context:

- https://github.com/temporalio/api-go/pull/297 teaches the API payload visitor to traverse marked System Nexus protobuf envelopes.
- https://github.com/temporalio/temporal/pull/10948 marks System Nexus results so API clients can identify those envelopes.
- https://github.com/temporalio/cli/pull/1017 added CLI-local nested decoding before that shared support existed.

## What changed?

CLI #1017 explicitly called the remote codec while rendering detailed System Nexus history. With api-go #297, the gRPC interceptor now finds and decodes the nested payloads itself. Keeping both paths can send the same payloads to the codec server twice.

This change:

- pins api-go main at `859a8e8d17c0a8c25fa2461214365841fee6d3a0`, which includes the marked System Nexus traversal;
- bumps the development server to `v1.32.0-163.3`, which includes the server marker from temporal#10948;
- removes the CLI's second remote decode path and its extra codec plumbing;
- resolves marked envelopes generically from their `messageType` metadata instead of maintaining a hard-coded operation/type registry; and
- extends the shared-server test to use an HTTP codec endpoint and assert that detailed `workflow show` both renders the decoded nested value and makes exactly one `/decode` request.

Raw JSON output remains unchanged. Because this feature has not been enabled, this change intentionally does not preserve compatibility with the earlier unmarked envelope implementation.

This does **not** solve the whole serialization issue in #1193. Payload groups elsewhere in history are still visited serially: SDK Go's `PayloadCodecGRPCClientInterceptorOptions` does not currently expose and pass api-go's `VisitPayloadsOptions.ConcurrencyLimit`. A follow-up SDK Go change is still required to make those codec-server requests concurrent.

## Checklist

- [x] Added functional tests
- [x] Added unit tests
- [x] This is a non-breaking change
- [ ] This feature does not depend on Temporal Cloud functionality

The test server uses the Cloud development tag because the marker is present in `v1.32.0-163.3` but not the stable OSS `v1.32.0` dependency previously used by the CLI test suite.

## Manual tests

Not run separately; covered by the shared-server end-to-end test.

Automated verification:

- `go test ./...`
- `(cd cliext && go test ./...)`


chaptersix:
not merging until there's a tagged api version

PR #39, opened 2022-11-30
Fix CI on linux and macOS

feedmeapples:
## What was changed

Updated GHA actions to their latest versions & tweaks

## Why?

Fix CI

## Checklist

1. Closes 

2. How was this tested:

CI

3. Any docs updates needed?

PR #1211, opened 2026-09-14
Block pre-release CLI versions from triggering docs updates

Duncanma:
## Summary

`trigger-docs.yml` fires on `release: types: [published]`, which GitHub sends for pre-releases too. `v1.9.0-rc.1` (run #38) and `v1.9.0` published with the pre-release flag set (run #39) both dispatched `update-cli-docs.yml` in `temporalio/documentation`; run #38 opened an unwanted docs PR ([documentation#5312](https://github.com/temporalio/documentation/pull/5312), now closed unmerged).

Two related defects in the same file, fixed alongside the gate:
- The tag was read from `github.ref_name`, which resolves to the selected *branch* on a manual run — run #37 dispatched `cli_release_tag=main`.
- `steps.get_user.outputs.GIT_NAME` and `secrets.GITHUB_TOKEN` were interpolated directly into `run:` blocks, unlike `github.actor` in the same file, which already goes through `env:`.

### Why not just add `prerelease == false` to the existing `published` trigger?

That fixes the reported symptom but opens a silent gap: promoting a release from pre-release to full release emits `released`, not `published`, so the GA docs update would never fire — `published` already fired while it was a pre-release and won't fire again. `types: [released]` fires on a direct full-release publish *and* on a pre-release → release promotion, and never for a pre-release, with no double-fire risk.

## Changes (3 commits, reviewable independently)

1. **Block pre-release CLI versions from triggering docs updates** — `published` → `released`, plus a job-level `if:` gate that also rejects drafts and side-channel tags (`v1.8.3-server-1.32.0-162.0`, `v1.7.4-standalone-nexus-operations`, `v1.6.3-serverless`, and similar) built from feature branches, as defense in depth. `workflow_dispatch` stays exempt so docs can still be backfilled on demand.
2. **Source the release tag from the release payload, not `github.ref_name`** — adds a required `release_tag` input for `workflow_dispatch` and sources `RELEASE_TAG` from the release payload's `tag_name` or the new input.
3. **Move interpolated values into `env:`** — hygiene, consistent with how `github.actor` is already handled.

## Verification

Checked every release in the repo's history against the new gate logic via the GitHub API: every pre-release and side-channel tag in history already carries `prerelease: true`, so the gate would have blocked all of them; no full release (`prerelease: false`) has ever carried a hyphenated tag suffix, so the tag-shape check hasn't excluded anything unexpected historically.

A gated-out release shows as a skipped (grey) run rather than a failure, so release managers aren't paged by a deliberate skip.

Companion PR in `temporalio/documentation`: [#5326](https://github.com/temporalio/documentation/pull/5326), a receiver-side guard on `update-cli-docs.yml` that validates the tag independent of this fix.

## Test plan

- [ ] Confirm a pre-release publish (e.g. a future `-rc.` tag) shows the `update` job as skipped, not failed
- [ ] Confirm a `workflow_dispatch` run still proceeds and reaches `temporalio/documentation`
- [ ] After merge, promote `v1.9.0` from pre-release to full release and confirm exactly one docs PR appears


.github/workflows/trigger-docs.yml
```diff
@@ -1,16 +1,45 @@
...
+    #     (v1.8.3-server-1.32.0-162.0, v1.7.4-standalone-nexus-operations,
+    #     v1.6.3-serverless), which are published from non-mainline code and must
+    #     not regenerate published docs.
+    # workflow_dispatch is exempt so docs can be backfilled for any tag on demand;
+    # the tag it passes is validated by the receiving workflow in
+    # temporalio/documentation.
+    if: |
+      github.repository == 'temporalio/cli'
+      && (github.event_name == 'workflow_dispatch'
+          || (github.event.release.prerelease == false
```
chaptersix, line 33: could you just check for dispatch or a release marked as latest? 

note, we mark a release as pre-release then change it to latest after the release assets have been attached. could you fork the repo and test this flow out?

chaptersix: forking allows you to create released without impacting real ones here.

.github/workflows/trigger-docs.yml
```diff
@@ -1,28 +1,77 @@
...
```
Duncanma, line 33: Tested this live in a fork (github.com/Duncanma/cli) rather than trusting the docs. Findings:

- A fresh pre-release publish fires `created` + `prereleased` + `published` — never `released`, and with `types: [released]` the workflow doesn't even start a run for it (not skipped — no run at all).
- Promoting an existing pre-release to a full release (`prerelease: true → false`, the flow you described) does fire `released` — confirmed via `github.event.action` logged directly from an Actions run, not inferred from the public events feed (which is a less reliable signal here — it shows `published` for these edits too, so it's not a good proxy).
- Doing that promotion together with `--latest` (your actual flow: mark pre-release, attach assets, then flip to latest) still fires `released`. It doesn't get suppressed by the make_latest change.
- The job-level `if:` correctly skipped every one of these on the fork (`github.repository == 'temporalio/cli'` is false there), so nothing could reach the real `temporalio/documentation` repo during testing.

One caveat worth recording: of 4 promotion tests, the very first one (the first release ever created on the fresh fork) produced no event at all even after several minutes — looks like one-time webhook registration lag on a brand-new repo, not a pattern, since the next 3 all fired within under a minute. Worth keeping in mind if `released` is ever silent for a real release — check a bit later before assuming the gate is broken.

Also added a commit for your "check for latest" idea: `make_latest` isn't actually in the release payload (confirmed against the live API — a release object exposes prerelease/draft/tag_name/etc., nothing about latest status), so I added a step that compares the tag against `GET /releases/latest` and skips if it doesn't match. This closes a gap prerelease/draft/tag-shape alone don't cover: a hotfix to an older minor line (e.g. patching v1.7.x after v1.8.0 is GA) is a real full release and would otherwise regenerate docs from an older version, overwriting the newer one's command reference.

.github/workflows/trigger-docs.yml
```diff
@@ -1,28 +1,77 @@
...
+        # generation. The release payload has no "is latest" field, so compare
+        # against /releases/latest instead. workflow_dispatch is exempt, same
+        # as the job-level gate: backfilling an older tag on purpose is fine.
+        if: github.event_name != 'workflow_dispatch'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          latest_tag="$(gh api "repos/${{ github.repository }}/releases/latest" --jq .tag_name)"
+          if [ "$latest_tag" != "$RELEASE_TAG" ]; then
```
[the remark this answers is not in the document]
Duncanma: Good catch — fixed. Replaced the `/releases/latest` comparison with a version-sort against all non-draft, non-prerelease tags (`sort -V`), so a hotfix mistakenly published with `--latest=true` no longer passes: it's compared against actual version order, not the mutable latest marker. Pushed in 28b801d.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/trigger-docs.yml
Duncanma: Tested live in a fork (github.com/Duncanma/cli): a fresh pre-release never fires `released`, but promoting an existing pre-release to a full release — including combined with `--latest`, matching your described flow — fires `released` reliably (confirmed via `github.event.action` logged from an actual Actions run, 3/4 tests; the one miss looked like one-time webhook lag on a brand-new fork, not a pattern). Full findings in the comment below on this line.

Also added the "check for latest" gate, then corrected it per Codex's follow-up review: it now compares against the highest published GA version via `sort -V` rather than trusting the `/releases/latest` marker, since that marker doesn't guarantee version order.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Duncanma: Agreed — forked to github.com/Duncanma/cli and ran the actual promote-to-latest flow there; nothing touched the real temporalio/documentation repo (the job's own repository check correctly skipped every run on the fork). Findings and cleanup details in the reply above and the comment below on this line.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/trigger-docs.yml
```diff
@@ -1,28 +1,85 @@
...
+        # would satisfy an equality check against /releases/latest while
+        # still being exactly the release this step exists to reject.
+        if: github.event_name != 'workflow_dispatch'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          highest="$(gh api "repos/${{ github.repository }}/releases" --paginate \
+            --jq '.[] | select(.draft == false and .prerelease == false) | .tag_name' \
+            | sort -V | tail -n1)"
```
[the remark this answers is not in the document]
Duncanma: Confirmed and fixed — tested locally: `sort -V` does rank `v2.0.0-serverless` above `v2.0.0`. Added the same strict `^vX.Y.Z$` tag-shape filter the job-level gate already uses to the release-listing query before sorting, so a side-channel tag can't win the highest-version comparison even if it were ever published as `prerelease: false`. Pushed in 1b3c619.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

